PR #153, opened 2015-04-21
A few minor patches for libgd

davidchappelle:


CMakeLists.txt
```diff
@@ -223,7 +223,6 @@ endif (USE_EXT_GD)
 
 add_subdirectory(tests)
 add_subdirectory(examples)
-add_subdirectory(docs)
```
vapier: this looks good

src/gd_interpolation.c
```diff
@@ -1634,7 +1639,7 @@ BGD_DECLARE(gdImagePtr) gdImageScale(const gdImagePtr src, const unsigned int ne
 {
 	gdImagePtr im_scaled = NULL;
 
-	if (src == NULL || src->interpolation_id < 0 || src->interpolation_id > GD_METHOD_COUNT) {
```
vapier: i don't think this is a good idea.  from what i can tell, the sign of an enum is not guaranteed by the C spec (in fact, it's explicitly left open ended), so deleting this range check opens us up to bugs.

newer C++ standards allow for explicitly controlling the type via scoped enums, but that's only in C++, not C.

we could add a helper func in the header like:
```
static inline bool gdInterpolationMethodIsValid(gdInterpolationMethod id)
{
  return (uintmax_t)id < GD_METHOD_COUNT;
}
```

but that seems like a lot of boiler plate to add a simple cast ...

also, it looks like this code incorrectly lets through GD_METHOD_COUNT when we should return directly.

src/gd_interpolation.c
```diff
@@ -2515,7 +2520,7 @@ BGD_DECLARE(int) gdTransformAffineBoundingBox(gdRectPtr src, const double affine
 
 BGD_DECLARE(int) gdImageSetInterpolationMethod(gdImagePtr im, gdInterpolationMethod id)
 {
-	if (im == NULL || id < 0 || id > GD_METHOD_COUNT) {
```
vapier: same issue here

src/gd_filter.c
```diff
@@ -15,7 +15,12 @@
 #include <time.h>
 #include <math.h>
 
-#define NDEBUG  /* Uncomment to enable assertions. */
+#ifndef NDEBUG
```
vapier: this change doesn't make sense.  the spec for assert.h specifically says that it's allowed to be included multiple times and NDEBUG can change value every time.  so you've effectively disabled the logic we had here to disable assertions.  it also means the double assert.h inclusion is now pointless.  see the documentation:
http://pubs.opengroup.org/onlinepubs/9699919799/basedefs/assert.h.html

if you're trying to avoid redefinitions of NDEBUG, then add an #undef line before the #define.  that'll force consistent behavior no matter how NDEBUG is set.

src/gd_interpolation.c
```diff
@@ -63,8 +63,13 @@
 #include <math.h>
 #include <assert.h>
 
-#define NDEBUG 1 /* TODO: disable/enable assertions in configure. */
-#include <assert.h>
+#ifndef NDEBUG
```
vapier: same issue here

pierrejoye:
What is the status on this PR? :)

davidchappelle:
Should be good to merge this pull request now.

vapier:
these changes are full of whitespace noise.  i'm all for cleaning that up, but they should not be in commits that make functional/unrelated changes.

davidchappelle:
Crap. It is my vim settings set to auto-remove trailing whitespace. Let me
clean that up.

PR #155, opened 2015-04-22
Missing nullcheck in gd_topal.c

jitendarKumar:
Issue : In API  [ gdImageTrueColorToPaletteBody()]  gdImageNeuQuant() may possibly return NULL. 

nim can hold NULL value.

Should be NULL checked.

### DIFF
--- a/src/gd_topal.c
+++ b/src/gd_topal.c
@@ -1519,11 +1519,11 @@ static int gdImageTrueColorToPaletteBody (gdImagePtr oim, int dither, int colors
 			gdImageDestroy(nim);
 		}
 		nim = gdImageNeuQuant(oim, colorsWanted, oim->paletteQuantizationSpeed ? oim->paletteQuantizationSpeed : 2);
-		if (!nim) {
-			return FALSE;
-		}
 		if (cimP) {
 			*cimP = nim;
+		} 
+		if (!nim) {
+			return FALSE;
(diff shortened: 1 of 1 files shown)


src/gd_topal.c
```diff
@@ -1519,6 +1519,9 @@ static int gdImageTrueColorToPaletteBody (gdImagePtr oim, int dither, int colors
 			gdImageDestroy(nim);
 		}
 		nim = gdImageNeuQuant(oim, colorsWanted, oim->paletteQuantizationSpeed ? oim->paletteQuantizationSpeed : 2);
+		if (!nim) {
+			return FALSE;
```
vapier: i don't think this is entirely correct ... you should set *cimP to nim before returning.  see top of this func as an example.

maybe something like:
```
         nim = gdImageNeuQuant(oim, colorsWanted, oim->paletteQuantizationSpeed ? oim->paletteQuantizationSpeed : 2);
         if (cimP) {
             *cimP = nim;
+        }
+        if (!nim) {
+            return FALSE;
+        }
-        } else {
+        if (nim) {
             gdImageCopy(oim, nim, 0, 0, 0, 0, oim->sx, oim->sy);
             gdImageDestroy(nim);
         }
         return TRUE;
```

pierrejoye:
Thanks, good catch! :)

PR #157, opened 2015-04-22
Remove Deadcode in gd.c

jitendarKumar:
In API gdImageLine()

if `dx == 0` Checked and return else if `dy==0` Checked and return. 

if ((dx == 0) && (dy == 0))  never TRUE , code has no impact , can be removed.


pierrejoye:
Thanks for the patch and sorry for the delay! :)

PR #163, opened 2015-05-07
add gdImageCreateFromWebpPtr impl 

wangkun611:


vapier:
the CMAKE_INSTALL_LIBDIR change wasn't specific to webp/vpx, but i didn't think worth making you cut it out into a sep patch.  thanks for the fixes!

PR #166, opened 2015-06-03
Dereference null return value in gdtest.c

jitendarKumar:
Return value of function which returns null is dereferenced without checking at multiple places.

### DIFF
--- a/src/gdtest.c
+++ b/src/gdtest.c
@@ -50,6 +50,14 @@ main (int argc, char **argv)
 	ref = gdImageCreateFromPng (in);
 
 	fclose (in);
+	if (!im) {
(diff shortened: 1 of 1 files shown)


src/gdtest.c
```diff
@@ -50,6 +50,14 @@ main (int argc, char **argv)
 	ref = gdImageCreateFromPng (in);
 
 	fclose (in);
+	if (!im) {
+               	fprintf(stderr, "gdImageCreateFromPng failed.\n");
```
vapier: indentation in this section is all broken ... please to fix :)

pierrejoye:
Thanks :)

PR #181, opened 2015-08-05
gdImageScaleTwoPass memory leak fix

uranium:
Fixing memory leak in gdImageScaleTwoPass, as reported by @cmb69 and confirmed by @vapier.
This bug actually bit me in production and I'm very thankful that it was reported with an easy fix.


vapier:
i cherry picked the commit since this pull request is kind of messy, and the commit messages are not that clear.  please try to keep each commit standalone with all the details in its commit message.

see the commit 4751b606fa38edc456d627140898a7ec679fcc24 that i pushed as an example.

uranium:
Thanks! Apologies for the mess, will try to improve on those points!

PR #182, opened 2015-08-27
Make BUILD_SHARED_LIBS configurable

leoyanggit:
As previously build both shared and static if BUILD_SHARED_LIBS is ON.
Otherwise build static only.


vapier:
thanks, merged now !

PR #188, opened 2015-10-30
Free of NULL, Not needed.

jitendarKumar:
In file gd_png.c : 
line 296 and 410  image_data is NULL and by passing as 1st parameter to function 'gdFree' is freed. 
line 297 and 411  row_pointers is NULL and by passing as 1st parameter to function 'gdFree' is freed. 

However, this free does not have any impact and free(NULL) is treated as a no-op. So gdfree not needed. 

Thanks


vapier:
how exactly did you come to this conclusion ?  if you look at the two places you changed the code, they're setjmp callbacks.  that means the code is not executed immediately, but only later on when the error handling function calls longjmp.  the point of these is to have a single error path that the libpng code calls when an error happens and that might be after image_data was allocated.

jitendarKumar:
You are right. The NULL checks are needed to have a single error path for libpng code calls.

PR #189, opened 2015-10-30
NULL check before Dereferencing null return value

jitendarKumar:
In circletexttest.c, return value of gdImageCreateFromJpeg() and gdImageCreateTrueColor() must be checked before Dereferncing.

### DIFF
--- a/src/circletexttest.c
+++ b/src/circletexttest.c
@@ -26,12 +26,24 @@ int main(void)
 	in = fopen("eleanor.jpg", "rb");
 	if(!in) {
 		im = gdImageCreateTrueColor(300, 300);
+		if(!im) {
+			fprintf(stderr, "gdImageCreateTrueColor failed \n");
+			return 1;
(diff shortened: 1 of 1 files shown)


src/circletexttest.c
```diff
@@ -26,12 +26,24 @@ int main(void)
...
+			return 1;
+		}
 	}
 #else
 	im = gdImageCreateTrueColor(300, 300);
+	if(!im) {
+		fprintf(stderr, "gdImageCreateTrueColor failed \n");
+		return 1;
+	}
 #endif /* HAVE_LIBJPEG */
```
vapier: please put a single null pointer check here rather than copying & pasting it 3 times

vapier:
please squash the two commits into just one

vapier:
i've squashed this and pushed by hand in commit 78dad445b3fc8a331cf30318daa1b604c5eb006a

PR #194, opened 2015-11-24
git,repo: Ignore build spew

jasonwilliams200OK:
Updated `.gitignore` and tested on Windows and FreeBSD that after running `cmake . && cmake --build .` checking `git status` does not show untracked files anymore (which wasn't the case before this change).


vapier:
erp, wrong button

PR #196, opened 2015-12-23
Add complex text layout support

asmaAL-Bahanta:
This pull request adds support for languages that require [complex text
layout](https://en.wikipedia.org/wiki/Complex_text_layout).

We are using [libraqm](https://github.com/HOST-Oman/libraqm), a small source
code-only library that wraps FriBidi (for bidirectional text support) and
HarfBuzz (for text shaping), and does proper BiDi and script itemization.

The CTL support is enabled by default but can be disabled at compiling time,
and we provide a fallback function that uses your original code without CTL
support.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -16,22 +16,21 @@ before_install:
     then
       sudo apt-get update -qq;
       sudo apt-get -y install nasm;
-      sudo apt-get -y install autotools-dev \
-                              libpng-dev \
-                              libz-dev \
-                              libjpeg-dev \
-                              libfreetype6-dev \
-                              libxpm-dev \
-                              libx11-dev \
-                              libxt-dev \
-                              libfontconfig-dev \
-                              libvpx-dev \
-                              libtiff-dev \
-                              libwebp-dev \
-                              autoreconf \
-                              automake \
-                              autoconf \
-                              libtool \
+      sudo apt-get -y install autotools-dev
+                              libpng-dev
+                              libz-dev
+                              libjpeg-dev
+                              libfreetype6-dev
+                              libxpm-dev
+                              libx11-dev
+                              libxt-dev
+                              libfontconfig-dev
+                              libvpx-dev
+                              libtiff-dev
+                              libwebp-dev
+                              automake
+                              autoconf
+                              libtool
                               html2text;
       if [ "$tbs_arch" = "x86" ];
       then
--- a/configure.ac
+++ b/configure.ac
@@ -208,6 +208,14 @@ if test "$gd_with_LIBFREETYPE" = "yes"; then
   AC_DEFINE([HAVE_FT2BUILD_H], [1], [Define if you have the ft2build.h header.])
 fi
 
+dnl Check for Raqm support.
+GD_LIB_PKG_CHECK([LIBRAQM], [RAQM], [raqm], [raqm], [
+  AC_CHECK_LIB([raqm], [raqm_create], [dnl
+    AS_VAR_APPEND([LIBRAQM_LIBS], [" -lraqm -lfreetype"])
+    gd_found_lib=yes
+  ])
+])
+
 dnl Check for fontconfig support.
 GD_LIB_PKG_CHECK([LIBFONTCONFIG], [FONTCONFIG], [fontconfig], [fontconfig], [
   AC_CHECK_LIB([fontconfig], [FcInit], [dnl
@@ -271,6 +279,7 @@ AC_MSG_RESULT([
    Support for Freetype 2.x library: $gd_with_LIBFREETYPE
    Support for Fontconfig library:   $gd_with_LIBFONTCONFIG
    Support for Xpm library:          $gd_with_LIBXPM
+   Support for complext text:        $gd_with_LIBRAQM
    Support for pthreads:             $ax_pthread_ok
 ])
 
--- a/src/gdft.c
+++ b/src/gdft.c
@@ -431,6 +431,107 @@ gdTcl_UtfToUniChar (char *str, Tcl_UniChar * chPtr)
 	return 1;
 }
 
+#ifdef HAVE_LIBRAQM
+#include <raqm.h>
+#endif
+
+typedef struct {
+	unsigned int index;
+	FT_Pos x_advance;
+	FT_Pos x_offset;
+	FT_Pos y_offset;
+	uint32_t cluster;
+} glyphInfo;
+
+static size_t
+textLayout(uint32_t *text, int len,
+		FT_Face face, gdFTStringExtraPtr strex,
+		glyphInfo **glyph_info)
+{
+	size_t count;
+	glyphInfo *info;
+
+#ifdef HAVE_LIBRAQM
+	size_t i;
+	raqm_glyph_t *glyphs;
+	raqm_t *rq = raqm_create ();
+
+	if (!rq || !raqm_set_text (rq, text, len) ||
+		!raqm_set_freetype_face (rq, face) ||
+		!raqm_set_par_direction (rq, RAQM_DIRECTION_DEFAULT) ||
+		!raqm_layout (rq)) {
+		raqm_destroy (rq);
+		return 0;
+	}
+
+	glyphs = raqm_get_glyphs (rq, &count);
+	if (!glyphs) {
+		raqm_destroy (rq);
+		return 0;
+	}
+
+	info = (glyphInfo*) gdMalloc (sizeof (glyphInfo) * count);
+	if (!info) {
+		raqm_destroy (rq);
+		return 0;
+	}
+
+	for (i = 0; i < count; i++) {
+		info[i].index = glyphs[i].index;
+		info[i].x_offset = glyphs[i].x_offset;
+		info[i].y_offset = glyphs[i].y_offset;
+		info[i].x_advance = glyphs[i].x_advance;
+		info[i].cluster = glyphs[i].cluster;
+	}
+
+	raqm_destroy (rq);
+#else
+	FT_UInt glyph_index = 0, previous = 0;
+	FT_Vector delta;
+	FT_Error err;
+	info = (glyphInfo*) gdMalloc (sizeof (glyphInfo) * len);
+	if (!info) {
+		return 0;
+	}
+	for (count = 0; count < len; count++) {
+		/* Convert character code to glyph index */
+		glyph_index = FT_Get_Char_Index (face, text[count]);
+
+		/* retrieve kerning distance */
+		if (! (strex && (strex->flags & gdFTEX_DISABLE_KERNING))
+			&& ! FT_IS_FIXED_WIDTH(face)
+			&& FT_HAS_KERNING(face)
+			&& previous
+			&& glyph_index)
+			FT_Get_Kerning (face, previous, glyph_index, ft_kerning_default, &delta);
+		else
+			delta.x = delta.y = 0;
+
+		err = FT_Load_Glyph (face, glyph_index, FT_LOAD_DEFAULT);
+		if (err) {
+			gdFree (info);
+			return 0;
+		}
+		info[count].index = glyph_index;
+		info[count].x_offset = 0;
+		info[count].y_offset = 0;
+		if (delta.x != 0)
+			info[count - 1].x_advance += delta.x;
+		info[count].x_advance = face->glyph->metrics.horiAdvance;
+		info[count].cluster = count;
+
+		/* carriage returns or newlines */
+		if (text[count] == '\r' || text[count] == '\n')
+			previous = 0;	/* clear kerning flag */
+		else
+			previous = glyph_index;
+	}
+#endif
+
+	*glyph_info = info;
+	return count;
+}
+
 /********************************************************************/
 /* font cache functions                                             */
 
@@ -950,20 +1051,23 @@ BGD_DECLARE(char *) gdImageStringFTEx (gdImage * im, int *brect, int fg, char *f
                                        gdFTStringExtraPtr strex)
 {
 	FT_Matrix matrix;
-	FT_Vector penf, oldpenf, delta, total_min = {0,0}, total_max = {0,0}, glyph_min, glyph_max;
+	FT_Vector penf, oldpenf, total_min = {0,0}, total_max = {0,0}, glyph_min, glyph_max;
 	FT_Face face;
 	FT_CharMap charmap = NULL;
 	FT_Glyph image;
 	FT_GlyphSlot slot;
 	FT_Error err;
-	FT_UInt glyph_index, previous;
+	FT_UInt glyph_index;
 	double sin_a = sin (angle);
 	double cos_a = cos (angle);
-	int len, i, ch;
+	int  i, ch;
 	font_t *font;
 	fontkey_t fontkey;
 	char *next;
 	char *tmpstr = 0;
+	uint32_t *text;
+	glyphInfo *info = NULL;
+	size_t count;
 	int render = (im && (im->trueColor || (fg <= 255 && fg >= -255)));
 	FT_BitmapGlyph bm;
 	/* 2.0.13: Bob Ostermann: don't force autohint, that's just for testing
@@ -1176,31 +1280,11 @@ BGD_DECLARE(char *) gdImageStringFTEx (gdImage * im, int *brect, int fg, char *f
 
 	oldpenf.x = oldpenf.y = 0; /* for postscript xshow operator */
 	penf.x = penf.y = 0;	/* running position of non-rotated glyphs */
-	previous = 0;		/* index of previous glyph for kerning calculations */
-	for (i=0; *next; i++) {
-		FT_Activate_Size (platform_independent);
-
+	text = (uint32_t*) gdCalloc (sizeof (uint32_t), strlen(next));
+	i = 0;
+	while (*next) {
+		int len;
 		ch = *next;
-
-		/* carriage returns */
-		if (ch == '\r') {
-			penf.x = 0;
-			previous = 0;		/* clear kerning flag */
-			next++;
-			continue;
-		}
-		/* newlines */
-		if (ch == '\n') {
-			/* 2.0.13: reset penf.x. Christopher J. Grayce */
-			penf.x = 0;
-			penf.y += linespace * ptsize * 64 * METRIC_RES / 72;
-			penf.y &= ~63;	/* round down to 1/METRIC_RES */
-			previous = 0;		/* clear kerning flag */
-			next++;
-			continue;
-		}
-
-
 		switch (encoding) {
 		case gdFTEX_Unicode: {
 			/* use UTF-8 mapping from ASCII */
@@ -1276,22 +1360,43 @@ BGD_DECLARE(char *) gdImageStringFTEx (gdImage * im, int *brect, int fg, char *f
 			next++;
 			break;
 		}
+		text[i] = ch;
+		i++;
+	}
 
-		/* Convert character code to glyph index */
-		glyph_index = FT_Get_Char_Index (face, ch);
+	FT_Activate_Size (platform_independent);
 
-		/* retrieve kerning distance */
-		if ( ! (strex && (strex->flags & gdFTEX_DISABLE_KERNING))
-		        && ! FT_IS_FIXED_WIDTH(face)
-		        && FT_HAS_KERNING(face)
-		        && previous
-		        && glyph_index)
-			FT_Get_Kerning (face, previous, glyph_index, ft_kerning_default, &delta);
-		else
-			delta.x = delta.y = 0;
+	count = textLayout (text , i, face, strex, &info);
+	gdFree (text);
 
-		penf.x += delta.x;
+	if (!count) {
+		gdFree (tmpstr);
+		gdCacheDelete (tc_cache);
+		gdMutexUnlock (gdFontCacheMutex);
+		return "Problem doing text layout";
+	}
 
+	for (i = 0; i < count; i++) {
+		FT_Activate_Size (platform_independent);
+
+		ch = text[info[i].cluster];
+
+		/* carriage returns */
+		if (ch == '\r') {
+			penf.x = 0;
+			continue;
+		}
+
+		/* newlines */
+		if (ch == '\n') {
+			/* 2.0.13: reset penf.x. Christopher J. Grayce */
+			penf.x = 0;
+			penf.y += linespace * ptsize * 64 * METRIC_RES / 72;
+			penf.y &= ~63;	/* round down to 1/METRIC_RES */
+			continue;
+		}
+
+		glyph_index = info[i].index;
 		/* When we know the position of the second or subsequent character,
 		save the (kerned) advance from the preceeding character in the
 		xshow vector */
@@ -1335,7 +1440,7 @@ BGD_DECLARE(char *) gdImageStringFTEx (gdImage * im, int *brect, int fg, char *f
 			return "Problem loading glyph";
 		}
 
-		horiAdvance = slot->metrics.horiAdvance;
+		horiAdvance = info[i].x_advance;
 
 		if (brect) {
 			/* only if need brect */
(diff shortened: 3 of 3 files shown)


configure.ac
```diff
@@ -208,6 +208,31 @@ if test "$gd_with_LIBFREETYPE" = "yes"; then
   AC_DEFINE([HAVE_FT2BUILD_H], [1], [Define if you have the ft2build.h header.])
 fi
 
+dnl CTL configure tests
```
vapier: delete all of this and use the new GD_LIB_PKG_CHECK helper

vapier:
this is against an old version of gd, so please rebase onto the latest master, and you must make sure Travis is passing first

configure.ac
asmaAL-Bahanta: we are using `AC_ARG_ENABLE` because ctl is a feature not a single package, so it should be `--enable-ctl` as `--with-ctl` does not seem appropriate.

asmaAL-Bahanta:
We fixed the failing test and travis build on linux.
Mac build uses cmake but we couldn't get it working under Ubuntu 14.04 so we didn't modify it to check for ctl libraries.

configure.ac
```diff
@@ -208,6 +208,31 @@ if test "$gd_with_LIBFREETYPE" = "yes"; then
...
+AC_ARG_ENABLE(ctl,
+    [AS_HELP_STRING([--disable-ctl],
+                    [disable complex text layout support])],
+    [enable_ctl=$enableval],
```
vapier: use the variable name gd_enable_CTL to match the existing gd_with_XXX style

configure.ac
```diff
@@ -208,6 +208,31 @@ if test "$gd_with_LIBFREETYPE" = "yes"; then
...
+    [enable_ctl='yes'])
```
vapier: this should be "auto", not "yes"

configure.ac
```diff
@@ -208,6 +208,31 @@ if test "$gd_with_LIBFREETYPE" = "yes"; then
...
+
+if test "x$gd_with_LIBFREETYPE" != "xyes" -a "x$enable_ctl" != 'xno'; then
+    AC_MSG_ERROR([Complex text support requested but FreeType is missing])
+fi
+
+have_ctl=""
+if test "x$enable_ctl" != 'xno'; then
+    PKG_CHECK_MODULES(CTL, [fribidi harfbuzz], have_ctl="yes", [])
```
vapier: split this up into two lib checks and use standard vars:
``PKG_CHECK_MODULES([FRIBIDI], [fribidi], [gd_with_FRIBIDI=yes], [gd_with_FRIBIDI=no])``

then you can do:

    if test "x$gd_with_FRIBIDI$gd_with_HARFBUZZ" != "xyesyes"; then
      if test "x$gd_enable_CTL" = "xyes"; then
        AC_MSG_ERROR(...)
      fi
      gd_enable_CTL="no"
    fi

configure.ac
```diff
@@ -208,6 +208,31 @@ if test "$gd_with_LIBFREETYPE" = "yes"; then
...
+fi
+
+if test "x$have_ctl" == "xyes"; then
```
vapier: this code is invalid -- you can't use ``==`` w/test.  you must use ``=``.

configure.ac
```diff
@@ -208,6 +208,31 @@ if test "$gd_with_LIBFREETYPE" = "yes"; then
...
+    AC_DEFINE(HAVE_CTL, 1, [Define if you have the complex text support.])
+    CPPFLAGS="$CPPFLAGS $CTL_CFLAGS"
+    LIBS="$LIBS $CTL_LIBS"
+    FEATURES="GD_CTL $FEATURES"
+    AC_CHECK_FUNC(fribidi_reorder_runs, AC_DEFINE(HAVE_FRIBIDI_REORDER_RUNS,  1, [Define if FriBiDi has fribidi_reorder_runs]), [])
+fi
+AM_CONDITIONAL(HAVE_CTL, test "x$have_ctl" == "xyes")
```
vapier: same error here w/``==``

src/gdft.c
```diff
@@ -431,6 +431,66 @@ gdTcl_UtfToUniChar (char *str, Tcl_UniChar * chPtr)
...
+#else
+typedef struct {
+	int index;
+	int x_offset;
+	int x_advance;
+	int y_offset;
+	uint32_t cluster;
+} raqm_glyph_info_t;
+
+unsigned fallback_shape(unsigned int *text,int len,
```
vapier: put a space after the first ``,``

vapier: this should be marked static

src/raqm.c
```diff
@@ -0,0 +1,695 @@
...
+    hb_tag_to_string (hb_script_to_iso15924_tag (script), buff); \
+    buff[4] = '\0';
+#endif
+
+/* for older fribidi versions */
+#ifndef HAVE_FRIBIDI_REORDER_RUNS
+typedef struct _FriBidiRunStruct FriBidiRun;
+struct _FriBidiRunStruct
+{
+  FriBidiRun *prev;
```
vapier: fix the style in this file -- indent with tabs

src/raqm.c
```diff
@@ -0,0 +1,695 @@
...
+};
+
+/* Reverse an array of runs */
+static void reverse_run (FriBidiRun *arr, const FriBidiStrIndex len)
+{
+  FriBidiStrIndex i;
+
+  assert (arr);
+
+  for (i = 0; i < len/2; i++)
```
vapier: use spaces around the ``/`` operator

src/raqm.c
```diff
@@ -0,0 +1,695 @@
...
+  FriBidiStrIndex i;
+  FriBidiLevel level;
+  FriBidiLevel last_level = -1;
+  FriBidiLevel max_level = 0;
+  FriBidiStrIndex run_count = 0;
+  FriBidiStrIndex run_start = 0;
+  FriBidiStrIndex run_index = 0;
+
+ if
+    (len == 0)
```
vapier: code style here is broken

src/raqm.c
```diff
@@ -0,0 +1,695 @@
...
+    {
+      goto out;
+    }
+
+  assert (bidi_types);
+  assert (embedding_levels);
+
+  {
```
vapier: use of braces here is pointless

src/raqm.h
```diff
@@ -0,0 +1,78 @@
...
+ * SOFTWARE.
+ *
+ */
+
+#ifndef _RAQM_H_
+#define _RAQM_H_
+
+/* For enabling debug mode */
+/*#define DEBUG 1*/
+#ifdef DEBUG
```
vapier: i really don't want to see these debug/test macros being duplicated further.  add some to gdhelpers.h and update the existing GD2_DBG/IO_DBG/GD_SS_DBG defines in the code base to use those.

asmaAL-Bahanta: These  macros are used for  [libraqm] (https://github.com/HOST-Oman/libraqm) tests.

src/raqm.c
asmaAL-Bahanta: This code is part of [libraqm](https://github.com/HOST-Oman/libraqm), and it is indented by spaces.
This particular code was submitted to FriBiDi and uses it's code style to be able to  sync it easily.

asmaAL-Bahanta:
I addressed all other comments.

Fahad-Alsaidi:
any update?

vapier:
i don't think we want to bundle libraqm in gd itself.  it should be treated as any other external 3rd party library.

asmaAL-Bahanta:
I have updated the pull request with raqm as standalone library.

src/gdft.c
```diff
@@ -1276,22 +1321,89 @@ BGD_DECLARE(char *) gdImageStringFTEx (gdImage * im, int *brect, int fg, char *f
...
-		glyph_index = FT_Get_Char_Index (face, ch);
+	FT_Activate_Size (platform_independent);
+#ifdef HAVE_LIBRAQM
+	raqm_t *rq;
+	raqm_direction_t dir;
+	dir = RAQM_DIRECTION_DEFAULT;
+	rq = raqm_create ();
+	if(rq == NULL){
+		raqm_destroy(rq);
+		if (tmpstr)
```
vapier: no need to do NULL checks

src/gdft.c
```diff
@@ -1276,22 +1321,89 @@ BGD_DECLARE(char *) gdImageStringFTEx (gdImage * im, int *brect, int fg, char *f
...
+			gdFree (tmpstr);
+		gdCacheDelete (tc_cache);
+		gdMutexUnlock (gdFontCacheMutex);
+		return "Problem creating Raqm Object";
+	}
+	if (!raqm_set_text (rq, text, i)){
+		raqm_destroy(rq);
+		if (tmpstr)
```
vapier: no need to do NULL checks

src/gdft.c
```diff
@@ -1276,22 +1321,89 @@ BGD_DECLARE(char *) gdImageStringFTEx (gdImage * im, int *brect, int fg, char *f
...
+			gdFree (tmpstr);
+		gdCacheDelete (tc_cache);
+		gdMutexUnlock (gdFontCacheMutex);
+		return "Problem setting Raqm Text";
+	}
+	if (!raqm_set_freetype_face (rq, face)){
+		if (tmpstr)
```
vapier: no need to do NULL checks

src/gdft.c
```diff
@@ -1276,22 +1321,89 @@ BGD_DECLARE(char *) gdImageStringFTEx (gdImage * im, int *brect, int fg, char *f
...
 
-		/* Convert character code to glyph index */
-		glyph_index = FT_Get_Char_Index (face, ch);
+	FT_Activate_Size (platform_independent);
+#ifdef HAVE_LIBRAQM
+	raqm_t *rq;
+	raqm_direction_t dir;
+	dir = RAQM_DIRECTION_DEFAULT;
+	rq = raqm_create ();
+	if(rq == NULL){
```
vapier: space after the if

src/gdft.c
```diff
@@ -431,6 +431,68 @@ gdTcl_UtfToUniChar (char *str, Tcl_UniChar * chPtr)
...
+		int y_advance;
+		int x_offset;
+		int y_offset;
+		uint32_t cluster;
+} raqm_glyph_t;
+
+static unsigned
+fallback_shape(unsigned int *text, int len,
+		FT_Face face, gdFTStringExtraPtr strex,
+		raqm_glyph_t** glyph_info)
```
vapier: spaces go before the ** not after

src/gdft.c
```diff
@@ -431,6 +431,68 @@ gdTcl_UtfToUniChar (char *str, Tcl_UniChar * chPtr)
 	return 1;
 }
 
+#ifdef HAVE_LIBRAQM
+#include <raqm.h>
+#else
+typedef struct {
+	unsigned int index;
+		int x_advance;
```
vapier: indentation is broken here

also, you use int, but you store values that are FT_Pos (signed long), so perhaps these should all be FT_Pos ?

src/gdft.c
```diff
@@ -1276,22 +1321,89 @@ BGD_DECLARE(char *) gdImageStringFTEx (gdImage * im, int *brect, int fg, char *f
...
-		if ( ! (strex && (strex->flags & gdFTEX_DISABLE_KERNING))
-		        && ! FT_IS_FIXED_WIDTH(face)
-		        && FT_HAS_KERNING(face)
-		        && previous
-		        && glyph_index)
-			FT_Get_Kerning (face, previous, glyph_index, ft_kerning_default, &delta);
-		else
-			delta.x = delta.y = 0;
+	if (!raqm_set_par_direction (rq, dir)){
+		if (tmpstr)
```
vapier: no need to do NULL checks

src/gdft.c
```diff
@@ -1276,22 +1321,89 @@ BGD_DECLARE(char *) gdImageStringFTEx (gdImage * im, int *brect, int fg, char *f
...
+	if (!raqm_layout (rq)){
+		if (tmpstr)
+			gdFree (tmpstr);
+		gdCacheDelete (tc_cache);
+		gdMutexUnlock (gdFontCacheMutex);
+		return "Problem setting Raqm layout";
+	}
+	info = raqm_get_glyphs (rq, &count);
+	if (info == NULL){
+		if (tmpstr)
```
vapier: no need to do NULL checks

.travis.yml
```diff
@@ -11,27 +11,33 @@ os:
   - linux
   - osx
 
+sudo: required # needed for trusty
+dist: trusty # needed for Raqm
+
 before_install:
   - if [ "$TRAVIS_OS_NAME" = "linux" ];
     then
+      sudo apt-add-repository -y ppa:as-bahanta/raqm;
```
vapier: i would leave this out which means raqm would not be tested by default.  you should submit the library for inclusion into places first.

khaledhosny: May be it is better to use a matrix and build once with and once without raqm, that would make sure the Raqm path does not bitrot? (note that the Mac build doesn’t use Raqm, so currently both code paths are exercised, though on different platforms).

khaledhosny: Also even if Raqm was added to Debian and Ubuntu today, it would be years before it reaches Travis, so the PPA will be needed for sometime anyway.

vapier: i'm aware of the impact.  it's unfortunate.  but i don't think we want to make gd's test system dependent upon a random ppa.  please don't take this personally as i would have the same response for anyone who wanted to add a ppa to the list.

khaledhosny: So what are the alternatives? I fear that the code will easily bitrot if it is not actively built and tested.

vapier: your concern is certainly valid.  but i think that's the route we should take as long as the package is unavailable in the distro.

looking at the changes, i don't think it's that big of a deal as code is not that intrusive.

khaledhosny: OK, I think this should work.

asmaAL-Bahanta:
I have Modified the pull request so that Raqm will not be tested by default.

.travis.yml
```diff
@@ -11,27 +11,27 @@ os:
   - linux
   - osx
 
+
```
vapier: please avoid unrelated whitespace changes

src/gdft.c
```diff
@@ -431,6 +431,68 @@ gdTcl_UtfToUniChar (char *str, Tcl_UniChar * chPtr)
...
+		FT_Face face, gdFTStringExtraPtr strex,
+		raqm_glyph_t **glyph_info)
+{
+	FT_UInt glyph_index, previous;
+	FT_Vector delta;
+	FT_Error err;
+	int count;
+	glyph_index = 0;
+	previous = 0;
+	raqm_glyph_t *g_info = (raqm_glyph_t*) gdMalloc(sizeof(raqm_glyph_t) * len + 1);
```
vapier: you need to check the return value to make sure it's not NULL

src/gdft.c
```diff
@@ -950,20 +1012,23 @@ BGD_DECLARE(char *) gdImageStringFTEx (gdImage * im, int *brect, int fg, char *f
...
 	double sin_a = sin (angle);
 	double cos_a = cos (angle);
-	int len, i, ch;
+	int  i, ch;
 	font_t *font;
 	fontkey_t fontkey;
 	char *next;
 	char *tmpstr = 0;
+	unsigned int *text;
+	raqm_glyph_t *info = NULL;
```
vapier: this is used outside of #ifdef HAVE_LIBRAQM, so won't this fail ?

src/gdft.c
```diff
@@ -1276,22 +1321,83 @@ BGD_DECLARE(char *) gdImageStringFTEx (gdImage * im, int *brect, int fg, char *f
...
+		gdMutexUnlock (gdFontCacheMutex);
+		return "Problem getting Raqm glyphs info";
+	}
+#else
+	count = fallback_shape (text , i, face, strex, &info);
+#endif
+
+	gdFree (text);
+
+	for(i = 0; i< count; i++) {
```
vapier: space before the <

src/gdft.c
asmaAL-Bahanta: No it won't fail, In case raqm is not defined there is a struct called `raqm_glyph_t` will be defined instead.

asmaAL-Bahanta:
Updated the pull request.

src/gdft.c
asmaAL-Bahanta: I have added Null check.

src/gdft.c
```diff
@@ -431,6 +431,71 @@ gdTcl_UtfToUniChar (char *str, Tcl_UniChar * chPtr)
...
+static unsigned
+fallback_shape(unsigned int *text, int len,
+		FT_Face face, gdFTStringExtraPtr strex,
+		raqm_glyph_t **glyph_info)
+{
+	FT_UInt glyph_index, previous;
+	FT_Vector delta;
+	FT_Error err;
+	int count;
+	glyph_index = 0;
```
vapier: just inline these two assignments above where the vars are declared

src/gdft.c
```diff
@@ -431,6 +431,71 @@ gdTcl_UtfToUniChar (char *str, Tcl_UniChar * chPtr)
...
+	previous = 0;
+	raqm_glyph_t *g_info = (raqm_glyph_t*) gdMalloc(sizeof(raqm_glyph_t) * len + 1);
+	if (!g_info) {
+		return NULL;
```
vapier: this func has a return type of unsigned, so using a pointer here makes no sense

src/gdft.c
```diff
@@ -431,6 +431,71 @@ gdTcl_UtfToUniChar (char *str, Tcl_UniChar * chPtr)
...
+			&& previous
+			&& glyph_index)
+			FT_Get_Kerning (face, previous, glyph_index, ft_kerning_default, &delta);
+		else
+			delta.x = delta.y = 0;
+
+		err = FT_Load_Glyph (face, glyph_index, FT_LOAD_DEFAULT);
+		if (err) {
+			gdMutexUnlock (gdFontCacheMutex);
+			return 0;
```
vapier: doesn't this leak the g_info memory ?

src/gdft.c
```diff
@@ -1276,22 +1324,83 @@ BGD_DECLARE(char *) gdImageStringFTEx (gdImage * im, int *brect, int fg, char *f
...
+	}
+	if (!raqm_set_text (rq, text, i)) {
+		raqm_destroy(rq);
+		gdFree (tmpstr);
+		gdCacheDelete (tc_cache);
+		gdMutexUnlock (gdFontCacheMutex);
+		return "Problem setting Raqm Text";
+	}
+	if (!raqm_set_freetype_face (rq, face)) {
+		gdFree (tmpstr);
```
vapier: shouldn't this raqm_destry too ?

src/gdft.c
```diff
@@ -1276,22 +1324,83 @@ BGD_DECLARE(char *) gdImageStringFTEx (gdImage * im, int *brect, int fg, char *f
...
-		/* Convert character code to glyph index */
-		glyph_index = FT_Get_Char_Index (face, ch);
+	FT_Activate_Size (platform_independent);
+#ifdef HAVE_LIBRAQM
+	raqm_t *rq;
+	raqm_direction_t dir;
+	dir = RAQM_DIRECTION_DEFAULT;
+	rq = raqm_create ();
+	if (rq == NULL) {
+		raqm_destroy(rq);
```
vapier: style here says space before the (

src/gdft.c
```diff
@@ -1276,22 +1324,83 @@ BGD_DECLARE(char *) gdImageStringFTEx (gdImage * im, int *brect, int fg, char *f
...
+		gdFree (tmpstr);
+		gdCacheDelete (tc_cache);
+		gdMutexUnlock (gdFontCacheMutex);
+		return "Problem creating Raqm Object";
+	}
+	if (!raqm_set_text (rq, text, i)) {
+		raqm_destroy(rq);
```
vapier: style here says space before the (

src/gdft.c
```diff
@@ -1276,22 +1324,83 @@ BGD_DECLARE(char *) gdImageStringFTEx (gdImage * im, int *brect, int fg, char *f
...
+		gdMutexUnlock (gdFontCacheMutex);
+		return "Problem getting Raqm glyphs info";
+	}
+#else
+	count = fallback_shape (text , i, face, strex, &info);
+#endif
+
+	gdFree (text);
+
+	for(i = 0; i < count; i++) {
```
vapier: space before the (

asmaAL-Bahanta:
I have addressed all the comments.

src/gdft.c
```diff
@@ -431,6 +431,70 @@ gdTcl_UtfToUniChar (char *str, Tcl_UniChar * chPtr)
...
+		FT_Face face, gdFTStringExtraPtr strex,
+		raqm_glyph_t **glyph_info)
+{
+	FT_UInt glyph_index = 0, previous = 0;
+	FT_Vector delta;
+	FT_Error err;
+	int count;
+	raqm_glyph_t *g_info = (raqm_glyph_t*) gdMalloc(sizeof(raqm_glyph_t) * len + 1);
+	if (!g_info) {
+		return NULL;
```
vapier: return value is int now, but you're still returning NULL -- a pointer

src/gdft.c
```diff
@@ -431,6 +431,70 @@ gdTcl_UtfToUniChar (char *str, Tcl_UniChar * chPtr)
...
+			&& FT_HAS_KERNING(face)
+			&& previous
+			&& glyph_index)
+			FT_Get_Kerning (face, previous, glyph_index, ft_kerning_default, &delta);
+		else
+			delta.x = delta.y = 0;
+
+		err = FT_Load_Glyph (face, glyph_index, FT_LOAD_DEFAULT);
+		if (err) {
+			gdMutexUnlock (gdFontCacheMutex);
```
vapier: is this supposed to mess with the mutex ?  seems wrong to me.  this should really be in the caller's code paths and not here.

asmaAL-Bahanta:
Updated.

asmaAL-Bahanta:
Any updates ?!

khaledhosny:
Any thing remaining to make the pull request ready?

khaledhosny:
Is there any chance of this code being merged?

vapier:
please rebase onto master and discard the travis commit ... that's been fixed since.  i'll try and take a look after that.

pierrejoye:
Have you checked gd-pango? Or what do you miss there?

I would also be sure that libraqm is portable (incl Windows), do you have any info if there is work in progress for that?

vapier:
merged via #246, thanks!

pierrejoye:
Hm.

Why is it merged?

vapier:
you mean this PR, or libraqm support in general ?  i thought we were moving towards supporting this, but i just saw/remembered your gd-pango question.

@khalidalqassabi: can you answer the gd-pango part ?

pierrejoye:
I am not confident to support it inside libgd as of now. We can have an extra module like what we did with pango until this library is proven to be stable and portable. The amount of changes adds complexity to something I would rather keep simple.

I also asked what is the core advantage of this library compared to pango f.e. that's something I like yo clarify before we decide to maintain support for this library.

Not really the biggest problem but these changes are creating issues with a fix we have to merge.

pierrejoye:
The raqm support in general :)

I am not saying I do not like it but it is a bit early to consider it as
core feature.

vapier:
yeah, sorry about that.  if they don't follow up, i'll revert.  fortunately we managed to delete a lot of code before it was merged ;).

khaledhosny:
The main benefit is less dependencies, Pango depends on GLib which can be pain to build on some platforms and almost depends on Cairo (it has alternative backends but Cairo is the most developed one). Also Raqm is much smaller than Pango (though it does a bit less than Pango as well).

khaledhosny:
BTW, the original intent was to not have a separate libraqm at all (it is just a single C file that could have been part of the repository).

Fahad-Alsaidi:
@pierrejoye  does anybody work in gd-pango?? last commit was in 2008 !! 

However, Raqm & pango do the same thing but Raqm is lighter and only depends on Harfbuzz & FriBiDi. I really don't care what will you prefer in GD but I only wish you could add CTL support in GD before 2050 !!

PR #198, opened 2016-01-09
Bug fix to _gdImageFillTiled

sherif-elmetainy:
_gdImageFillTiled was reading/writing outside the buffer, which was causing segmentation faults (Linux) and/or access violations (Windows). Due to a bug in calculation of array index, the function would read/write incorrect memory when checking a point whose x is higher than the image height,
and behave incorrectly when an image has higher greater than width.
 
The issue was raised in #118 and #104, it was fixed in the php implementation but the fix over there has more memory allocations. The issue can be reproduced using the following code:

```
gdImagePtr image = gdImageCreateTrueColor(400, 300);
if (image != NULL) {
	int white = gdImageColorClosestAlpha(image, 0xff, 0xff, 0xff, 0);
	int red = gdImageColorClosestAlpha(image, 0xff, 0x0, 0x0, 0);
	int blue = gdImageColorClosestAlpha(image, 0x0, 0x0, 0xff, 0);

	gdImageFilledRectangle(image, 0, 0, image->sx - 1, image->sy - 1, white);
	gdImagePtr tileImage = gdImageCreateTrueColor(2, 2);
	if (tileImage != NULL) {
		gdImageSetPixel(tileImage, 0, 0, red);
		gdImageSetPixel(tileImage, 1, 1, red);
		gdImageSetPixel(tileImage, 1, 0, blue);
		gdImageSetPixel(tileImage, 0, 1, blue);

		gdImageSetTile(image, tileImage);
		gdImageFill(image, 10, 10, gdTiled);

		FILE* outputFile = fopen("output.png", "wb+");
		if (outputFile != NULL) {
			int size;
			void* png = gdImagePngPtr(image, &size);
			if (png != NULL) {
				fwrite(png, 1, size, outputFile);
				gdFree(png);
			}
			fclose(outputFile);
		}
		gdImageDestroy(tileImage);
	}
	gdImageDestroy(image);
}
```


vapier:
having two commits here, especially when one deletes will the 2nd basically reverts all of that, makes no sense.  please squash them into one.

rather than declaring the testcase in the commit message, please put it into the testsuite directly so it gets built+tested when you run `make check`.

sherif-elmetainy:
I added a test case and squashed the commits as asked. However, I didn't know the naming convention used for naming test cases, I named the file bug00104_1.c where #104 is the issue where the bug was first reported. I can correct the file name if needed, just need to know the convention.

tests/gdimagefill/bug00104_1.c
```diff
@@ -0,0 +1,40 @@
+/* $Id$ */
```
vapier: drop this $Id$ ... it isn't useful anymore.  just put a description of the bug you're testing for here.

tests/gdimagefill/bug00104_1.c
```diff
@@ -0,0 +1,40 @@
...
+#include "gd.h"
+#include <stdio.h>
+#include "gdtest.h"
+
+int main()
```
vapier: this testcase doesn't look like the one you posted in the commit message.  is this what you meant to include ?

tests/gdimagefill/bug00104_1.c
```diff
@@ -0,0 +1,40 @@
...
+{
+	gdImagePtr im, tile;
+	int im_white, tile_red, tile_blue;
+	int x,y, error = 0;
```
vapier: space after the first comma

tests/gdimagefill/bug00104_1.c
```diff
@@ -0,0 +1,40 @@
...
+	char path[1024];
+	
+/*	fputs("flag 0\n", stdout); */
```
vapier: delete please

tests/gdimagefill/bug00104_1.c
```diff
@@ -0,0 +1,40 @@
...
+	im = gdImageCreate(200, 150);
+
+	tile = gdImageCreateTrueColor(2, 2);
+	
+
```
vapier: just one blank line

tests/gdimagefill/bug00104_1.c
```diff
@@ -0,0 +1,40 @@
...
+	tile_red = gdImageColorAllocate(tile,255,0,0);
+	tile_blue = gdImageColorAllocate(tile,0,0,255);
+	
+	gdImageSetPixel(tile, 0, 0, tile_red);
+	gdImageSetPixel(tile, 1, 1, tile_red);
+	gdImageSetPixel(tile, 1, 0, tile_blue);
+	gdImageSetPixel(tile, 0, 1, tile_blue);
+	
+
```
vapier: just one blank line

tests/gdimagefill/bug00104_1.c
```diff
@@ -0,0 +1,40 @@
...
+	gdImageSetPixel(tile, 0, 0, tile_red);
+	gdImageSetPixel(tile, 1, 1, tile_red);
+	gdImageSetPixel(tile, 1, 0, tile_blue);
+	gdImageSetPixel(tile, 0, 1, tile_blue);
+	
+	
+	gdImageSetTile(im,tile);
+	gdImageFill(im, 11,12, gdTiled);
+
+
```
vapier: just one blank line

tests/gdimagefill/bug00104_1.c
```diff
@@ -0,0 +1,40 @@
...
+	int x,y, error = 0;
+	char path[1024];
+	
+/*	fputs("flag 0\n", stdout); */
+	im = gdImageCreate(200, 150);
+
+	tile = gdImageCreateTrueColor(2, 2);
+	
+
+	tile_red = gdImageColorAllocate(tile,255,0,0);
```
vapier: spaces after the commas

tests/gdimagefill/bug00104_1.c
```diff
@@ -0,0 +1,40 @@
...
+	tile_blue = gdImageColorAllocate(tile,0,0,255);
```
vapier: same here

tests/gdimagefill/bug00104_1.c
```diff
@@ -0,0 +1,40 @@
...
+	
+	gdImageSetPixel(tile, 0, 0, tile_red);
+	gdImageSetPixel(tile, 1, 1, tile_red);
+	gdImageSetPixel(tile, 1, 0, tile_blue);
+	gdImageSetPixel(tile, 0, 1, tile_blue);
+	
+	
+	gdImageSetTile(im,tile);
+	gdImageFill(im, 11,12, gdTiled);
```
vapier: space after the 2nd comma

tests/gdimagefill/bug00104_1.c
```diff
@@ -0,0 +1,40 @@
...
+	tile_red = gdImageColorAllocate(tile,255,0,0);
+	tile_blue = gdImageColorAllocate(tile,0,0,255);
+	
+	gdImageSetPixel(tile, 0, 0, tile_red);
+	gdImageSetPixel(tile, 1, 1, tile_red);
+	gdImageSetPixel(tile, 1, 0, tile_blue);
+	gdImageSetPixel(tile, 0, 1, tile_blue);
+	
+	
+	gdImageSetTile(im,tile);
```
vapier: space after the comma

tests/gdimagefill/bug00104_1.c
```diff
@@ -0,0 +1,40 @@
...
+		
+	sprintf(path, "%s/gdimagefill/bug00104_1_exp.png", GDTEST_TOP_DIR);
+	if (!gdAssertImageEqualsToFile(path, im)) {
+		error = 1;
+	}
+	
+	gdImageDestroy(im);
+	return error;
+}
+
```
vapier: no trailing newlines please

tests/gdimagefill/CMakeLists.txt
```diff
@@ -4,6 +4,7 @@ SET(TESTS_FILES
 	bug00002_2
```
vapier: please add this new test & the png file to the tests/Makefile.am file

src/gd.c
```diff
@@ -2201,9 +2201,9 @@ static void _gdImageFillTiled(gdImagePtr im, int x, int y, int nc)
 	FILL_PUSH(y+1, x, x, -1);
 	while (sp>stack) {
 		FILL_POP(y, x1, x2, dy);
-		for (x=x1; x>=0 && (!pts[y + x*wx2] && gdImageGetPixel(im,x,y)==oc); x--) {
```
vapier: pts is a flattened array of pixels.  it would probably be easier to understand if it were used as a 2d array.  but it's not (currently).
```pts = (char *) gdCalloc(im->sy * im->sx, sizeof(char));```

so when the code here is trying to do `pts[(x,y)]`, it does so by `pts[x * im->sx + y]`.  changing it to `pts[x * wy2 + y]` no longer properly indexes the array.  it might not violate the bounds, but if you changed it to `pts[0]` everywhere, that too would not violate the bounds :).

tests/gdimagefill/bug00104_1.c
sherif-elmetainy: It is a similar one, but instead of writing the image to an output file, I am using gdAssertImageEqualsToFile like the rest of the tests. I also used different size image, but it's the same bug, as the bug occurs in any image with width > height. I changed the variable names to adhere to the same names used in the other tests (I didn't know if there is a convention, to I used the one in the other files).

So basically it's the same text.

src/gd.c
sherif-elmetainy: I disagree here, `x` should not be multiplied by `im->sx`. It should be multiplied by `im->sy`, this is the proper way to index the array. Since `x` is the column number, a new column starts every "height of column" pixels which is `sy` pixels. It can either be `x + y * im->sx` or `y + x * im->sy` but multiplying `x * im->sx` or `y * im->sy` would be both incorrect way to index the array and would result in out of bounds reads/writes.
Example `sx = 1000`, `sy = 1`. `pts[(100, 0)]` for `x = 100` and `y = 0`, should be `pts[100 * 1 + 0]` not `pts[100 * 100 + 0]`.

I agree it would be easier to understand if it was a 2d array, however, it would require `sy + 1` allocations which is fine but the allocation, cleanup and error recovery code would be harder to understand since a failure can occur in the middle of the allocations loop. By the way this is how the folks at PHP fixed it, but they didn't handle the case where allocation would fail during the allocations loop which could lead to memory leaks. 

So it's either more complex allocation, recovery and cleanup code or 1d array. Or may be we can have the best of both scenarios by using macros which was my initial solution but I decided against it.

sherif-elmetainy:
I made the changes request as per your review.

src/gd.c
vapier: you are correct that we want [y + x*im->sy]

i wasn't suggesting trying to allocate a 2d array since it's too expensive.  a local macro might not be so bad if it was localized to the func and you only pass in (x,y).  then again, doing a sx*sy*1 byte array when we just care about bools is a bit expensive.  we could add macros to handle setting bits instead which would cut the memory requirement to 1/8th.  if you're interested in that, it can be a follow up PR :).

PR #202, opened 2016-02-21
bugfix

konqui:
fixed typos which cause homebrew version of gd with option --with-libvpx to fail building

can you create a new branch out off the tag gd-2.1.1 and add my pull request to that branch.

so we have an new valid and compileable gd for homebrew


vapier:
this has already been fixed in commit 824a3aaa37b422779166ccc0ffc17e777f8cf291

PR #207, opened 2016-04-13
added a script to build the docs and put them in the gh-pages branch …

ChrisBarker-NOAA:
…for publishing on gitHub.io

See issue #85

The script creates (or uses) a parallel clone of the repo, switches that repo to the gh-pages branch, then generate the docs and copies them to the other repo, and commits and published them there.

If the script is run on a *nix machine by someone with commit privileges to the main gitHub repo that is setup up to build the docs (i.e. has NaturalDocs), it should publish the latest.

One could set this up to be run on a CI system on every commit to master if you wanted, but I don't know how to deal with the authorization issues, so I haven't tried that.

NOTE: the script is hard-coded to my fork of the repo, so I could test it out. See the result here:

http://chrisbarker-noaa.github.io/libgd

Also -- if the script is run from a branch for a different version, that version will get published -- so probably good to use a release version.

NOTE2: this is publishing to the repos gh-pages -- as the Web site is published elsewhere. I think it should stay that way, but they could be merged if you wanted -- I'm sure there is a way to put raw html in a Jekyll site somehow.

You could do this all by hand too, but I've found it helpful to have a single script to run to update.

### DIFF
--- a/README.md
+++ b/README.md
@@ -21,6 +21,8 @@ umbrella of PHP.net.
 
 Please visit our [homepage](http://www.libgd.org/) for more details.
 
+API documentation can be found here: http://libgd.github.io/libgd
+
 ## Supported Image Formats
 
 GD has builtin support for:
--- a/docs/naturaldocs/build_gh_pages.sh
+++ b/docs/naturaldocs/build_gh_pages.sh
@@ -1,0 +1,46 @@
+#!/bin/sh
+
+# a very hacky script to build and push to gh-pages
+# designed to be run from master (or any branch no the gh-pages branch)
+#
+# This will publish the html docs to the github.io url:
+# http://chrisbarker-noaa.github.io/libgd
+# this needs to be updated for the "proper" repo
+
+PARENTDIR=../../../
+REPONAME=libgd.gh_pages/
+GHPAGESDIR=$PARENTDIR$REPONAME
+
+# make sure the gh pages repo is there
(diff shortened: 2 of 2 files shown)


vapier:
our current infra looks like:
- make commits to https://github.com/libgd/website
- Travis runs, rebuilds the site, and pushes it to https://github.com/libgd/libgd.github.io

so in that regard, if we want to automate publishing of the manual, i think we'll want to update the website repo instead.  that can take care of checking out the libgd/libgd repo, building the man pages for any missing versions, and then committing+pushing.

make sense ?

ChrisBarker-NOAA:
Sure, If I understand correctly.

The idea is that:

The html docs would be built from the source in libgd/libgd repo.
Those html docs would be copied, committed and pushed to thelibgd/libgd.github.io repo.
And that would auto publish them at:

http://libgd.github.io/

That would be awesome.

-CHB

vapier:
yep.  the main script lives at https://github.com/libgd/website/blob/master/.travis.yml, but basically it runs `make publish`.  so if you updated that Makefile to include a new target (like `make manual`), and that target (either in the Makefile or by running some helper script) did the clone/build/install of the manual itself (for all the versions we care about), then it would make it out into the website automatically.

you can test this locally by running `make manual html` and then `make serve`, and then pointing a web browser to http://localhost:8000/.  once that looks good, you can post a PR to the website git repo.

PR #229, opened 2016-06-13
Fix #170: XBM images require width to be multiple of 8

### LINKED ISSUE #170: XBM images require width to be multiple of 8
When writing XBM images it is required that they have a width which is a multiple of 8. IOW, no automatic padding is done in gdImageXbmCtx(). However, it seems that this is easily accomplished, and the [respective line](https://github.com/libgd/libgd/blob/gd-2.1.1/src/gd_xbm.c#L234) is doubtful anyway, as x is always less than sx and y less than sy. Changing this line to the following should do the trick:

    if ((b == 128) || (x == sx - 1)) {

See also [the respective ticket](https://bugs.php.net/bug.php?id=53640) filed in the PHP bug tracker.

cmb69:
We remove this limitation by fixing the underlying implementation bug,
and add a respective regression test to the suite.

The test is not quite like we would want it, though. Firstly, just checking the XBM to have the expected size in bytes is neither clear nor resilient[1], but checking for the exact contents appears to be overly laborious with the current testing framework. Secondly, actually writing the XBM to disk is inefficient and not really necessary, but I haven't been able to verify the result after writing the XBM to an in-memory gdIOCtx.

[1] There might be an issue on Windows due to the line endings, but I haven't been able yet to set up the necessary build environment on a Windows machine so that I could verify. If there are issues, the test could be adapted to also allow the respective file size (presumably, 256 bytes), or simply skipped on Windows, as XBM is supposed to be a *nix thing only.

### DIFF
--- a/tests/xbm/CMakeLists.txt
+++ b/tests/xbm/CMakeLists.txt
@@ -1,0 +1,5 @@
+LIST(APPEND TESTS_FILES
+	github_bug_170
+)
+
+ADD_GD_TESTS()
--- a/tests/xbm/Makemodule.am
+++ b/tests/xbm/Makemodule.am
@@ -1,0 +1,5 @@
+libgd_test_programs += \
+	xbm/github_bug_170
+
+EXTRA_DIST += \
+	xbm/CMakeLists.txt
--- a/tests/xbm/github_bug_170.c
+++ b/tests/xbm/github_bug_170.c
@@ -1,0 +1,42 @@
+/*
+	Test writing of XBM images with a width that is not a multiple of 8
+
+	We create an image with a width of 11 pixels, and draw a circle on it.
+	To test that the padding is correctly applied, we write the image to disk
+	and assert that the number of bytes written matches our expectation.
+
+	See also <https://github.com/libgd/libgd/issues/170>.
+*/
+
+#include <inttypes.h>
+#include "gd.h"
+#include "gdtest.h"
+
+int main()
+{
+	gdImagePtr im;
+	int black;
+	FILE *outFile;
+	gdIOCtx *out;
+	off_t length;
+
+	/* create the test image */
+	im = gdImageCreate(11, 11);
+	gdImageColorAllocate(im, 255, 255, 255);
+	black = gdImageColorAllocate(im, 0, 0, 0);
+	gdImageArc(im, 5, 5, 10, 10, 0, 360, black);
+
+	/* write the file to disk, note the file length and delete the file */
+	outFile = gdTestTempFp();
+	out = gdNewFileCtx(outFile);
+	gdTestAssert(out != NULL);
+	gdImageXbmCtx(im, "github_bug_170.xbm", 1, out);
+	out->gd_free(out);
+	length = ftello(outFile);
+	fclose(outFile);
+
+	gdImageDestroy(im);
+
+	gdTestAssertMsg(length == 250, "expected to write 250 bytes; %jd bytes written", (intmax_t) length);
+	return gdNumFailures();
+}
--- a/src/gd_xbm.c
+++ b/src/gd_xbm.c
@@ -251,7 +251,7 @@ BGD_DECLARE(void) gdImageXbmCtx(gdImagePtr image, char* file_name, int fg, gdIOC
 			if (gdImageGetPixel(image, x, y) == fg) {
 				c |= b;
 			}
-			if ((b == 128) || (x == sx && y == sy)) {
+			if ((b == 128) || (x == sx - 1)) {
 				b = 1;
 				if (p) {
 					gdCtxPuts(out, ", ");
--- a/tests/CMakeLists.txt
+++ b/tests/CMakeLists.txt
@@ -61,6 +61,7 @@ if (BUILD_TEST)
 		tiff
 		wbmp
 		webp
+		xbm
 		xpm
 	)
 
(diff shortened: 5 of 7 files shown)


cmb69:
Forgot to mention that writing the XBM to disk has the advantage that the file might be manually examined if the test fails. That would require to uncomment the remove() call, though. Perhaps it would be better to only delete the file if the test succeeds, or to never delete the file at all, if we stick with writing the XBM to disk.

cmb69:
Oh, another issue: apparently, the new tests/xbm/ directory would have to be added to tests/CMakeLists.txt, but I didn't do so, because tests/xpm/ is also not contained in tests/CMakeLists.txt. Maybe this is for a good reason; otherwise it should be fixed.

vapier:
the cmake build is known to be missing some tests.  it's not by design.

that said, cmake does list xpm in `tests/CMakeLists.txt`.

tests/xbm/github_bug_170.c
```diff
@@ -0,0 +1,43 @@
...
+	long int length;
+
+	/* create the test image */
+	im = gdImageCreate(11, 11);
+	gdImageColorAllocate(im, 255, 255, 255);
+	black = gdImageColorAllocate(im, 0, 0, 0);
+	gdImageArc(im, 5, 5, 10, 10, 0, 360, black);
+
+	/* write the file to disk, note the file length and delete the file */
+	outFile = fopen("xbm/github_bug_170.xbm", "w");
```
vapier: use `gdTestTempFp` instead.  we do not want tests writing files in random spots anymore.

tests/xbm/github_bug_170.c
```diff
@@ -0,0 +1,43 @@
...
+	if (outFile == NULL) return 1;
```
vapier: delete this once you use `gdTestTempFp`

tests/xbm/github_bug_170.c
```diff
@@ -0,0 +1,43 @@
...
+	out = gdNewFileCtx(outFile);
+	if (out == NULL) return 1;
```
vapier: please don't one-line statements.  that said, use the gd assert helpers here instead of checking/returning yourself.

tests/xbm/github_bug_170.c
```diff
@@ -0,0 +1,43 @@
...
+#include "gd.h"
+#include "gdtest.h"
+
+int main()
+{
+	gdImagePtr im;
+	int black;
+	FILE *outFile;
+	gdIOCtx *out;
+	long int length;
```
vapier: this should be off_t, and you should use `ftello` below

cmb69:
> that said, cmake does list xpm in tests/CMakeLists.txt.

Ah, thanks, I had overlooked the conditional inclusion. Will fix the PR accordingly, and also address the other suggestions you have made as soon as possible.

cmb69:
Please review. If it's okay, I can merge into master and cherry-pick from there into 2.2.

tests/xbm/github_bug_170.c
```diff
@@ -0,0 +1,41 @@
...
+	out = gdNewFileCtx(outFile);
+	gdTestAssert(out != NULL);
+	gdImageXbmCtx(im, "github_bug_170.xbm", 1, out);
+	out->gd_free(out);
+	length = ftello(outFile);
+	fclose(outFile);
+
+	gdImageDestroy(im);
+
+	gdTestAssertMsg(length == 250, "expected to write 250 bytes; %d bytes written", length);
```
vapier: `length` is an `off_t`, so you'll prob want `%jd` instead of `%d` (with maybe a cast to `intmax_t`)

tests/xbm/CMakeLists.txt
```diff
@@ -0,0 +1,7 @@
+IF(XBM_FOUND)
```
vapier: there is no `XBM_FOUND` symbol in the cmake code.  looks like we always build the xbm code, so you'll want to drop this if statement.

tests/xbm/Makemodule.am
```diff
@@ -0,0 +1,7 @@
+if HAVE_LIBXPM
```
vapier: guessing you mean ``HAVE_LIBX``**B**``M`` here instead of `HAVE_LIBX``**P**`M``.  either way, there is no such symbol in the autoconf code, and we always enable xbm support, so you should drop this if statement.

tests/xbm/github_bug_170.c
cmb69: According to a [StackOverflow answer](http://stackoverflow.com/questions/586928/how-should-i-print-types-like-off-t-and-size-t#answer-22537979) `%jd` is not supported by VS 2013. Wouldn't that be an issue?

Also I'm not sure about `intmax_t`. This appears to be defined in `inttypes.h`, but I'm not sure if that's portable (especially considering MSVC).

Still have not been able to build libgd on Windows.

vapier: `intmax_t` and `inttypes.h` are in [POSIX](http://pubs.opengroup.org/onlinepubs/9699919799/basedefs/inttypes.h.html).

the alternative is cast it to `int64_t` and use `%"PRIi64"` (also from `inttypes.h`).

trying to use `%d` with `off_t` will trigger warnings.

pierrejoye:
Xpm has been forgot so it can be or should  expect fixed yes.

This PR looks good to me, good job, thanks  :)

vapier:
needs a `.gitignore` file but otherwise looks good to go

cmb69:
> needs a .gitignore file

There's already [tests/xbm/.gitignore](https://github.com/cmb69/libgd/blob/80ce084567118d00eb159cb70eedbc12178aeeef/tests/xbm/.gitignore). Isn't that sufficient?

vapier:
sorry, must have missed it when i was double checking

PR #237, opened 2016-06-19
Fix some issues reported by Coverity

cmb69:
* Unused value in gdft.c (ID 102122)
  Apparently, that's a leftover from the "mode_mono fix from Giuliano
  Pochini".
* Structurally dead code in gdft.c (ID 95850)
  We can slightly refactor the code to not use an `else` clause at all,
  what also makes the code cleaner.
* Copy-paste error in gd.c (ID 95837)
  Comparing that with [PHP's bundled libgd](https://github.com/php/php-src/blob/php-7.0.7/ext/gd/libgd/gd_arc.c#L94-L99)
  it appears that the duplication of the condition and the loop is useless.
  I'm not sure about that, though.

### DIFF
--- a/src/gd.c
+++ b/src/gd.c
@@ -1912,10 +1912,6 @@ BGD_DECLARE(void) gdImageFilledEllipse (gdImagePtr im, int mx, int my, int w, in
 		if(old_y2!=my2) {
 			for(i=mx1; i<=mx2; i++) {
 				gdImageSetPixel(im,i,my1,c);
-			}
-		}
-		if(old_y2!=my2) {
-			for(i=mx1; i<=mx2; i++) {
 				gdImageSetPixel(im,i,my2,c);
 			}
 		}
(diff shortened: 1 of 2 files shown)


vapier:
looks fine to me

pierrejoye:
I did not do it but we may refer the coverity  id to the commit message for tracking.

pierrejoye:
For the ellipse one, the best test case (and I suppose what I meant to do here) is to use it with alphablending. The key point being all points within (incl. edges) have to be drawn/set only once.

btw, would it be possible (for the next ones) to do one commit per CID?

pierrejoye:
If you can add the CID in the name of the PR then go ahead for the merge, thanks :)

2.2 and master please

cmb69:
> btw, would it be possible (for the next ones) to do one commit per CID?

The changes seemed so minimal for each fix, that I've bundled them in a single commit, but you're right that it's better to keep them separate. Will do so when I commit.

cmb69:
I have committed the fixes for Coverity #95850 and #102122 (master; cherry-picked to 2.2). I have not addressed Coverity #95837 (Copy-paste error in gdImageFilledEllipse()) yet, because I'd need to dig into alphablending (or maybe mocking gdImageSetPixel?) first to be able to assess the code (of course, the refactoring is trivial, but there might be a bug hidden, which would get obscured by the refactoring). Nonetheless I'm closing this PR – I've opened #238 to not forget about the outstanding issue.

Issue #248, opened 2016-07-05
Out-Of-Bounds Read in function read_image_tga of gd_tga.c


trylab:
Hi, I'd like to report a OOB read issue to you. This issue can be triggered when parsing a specially crafted tga image. You can compile ```libgd/tests/tga/bug00084.c``` with AddressSanitizer to reproduce this issue.
```
==2755==ERROR: AddressSanitizer: heap-buffer-overflow on address 0xb5503f9c at pc 0x081689da bp 0xbf86d738 sp 0xbf86d730
READ of size 1 at 0xb5503f9c thread T0
    #0 0x81689d9 in read_image_tga /tmp/repo/libgd/src/gd_tga.c:293:37
    #1 0x8165b32 in gdImageCreateFromTgaCtx /tmp/repo/libgd/src/gd_tga.c:74:6
    #2 0x8165986 in gdImageCreateFromTga /tmp/repo/libgd/src/gd_tga.c:25:10
    #3 0x81322a1 in main /tmp/repo/libgd/tests/tga/bug00084.c:10:10
    #4 0xb745daf2 in __libc_start_main /build/eglibc-X4bnBz/eglibc-2.19/csu/libc-start.c:287
    #5 0x8078b1b in _start (/tmp/repo/libgd/tests/tga/bug00084+0x8078b1b)

0xb5503f9c is located 0 bytes to the right of 796-byte region [0xb5503c80,0xb5503f9c)
allocated by thread T0 here:
    #0 0x81107a9 in malloc (/tmp/repo/libgd/tests/tga/bug00084+0x81107a9)
    #1 0x8167b77 in read_image_tga /tmp/repo/libgd/src/gd_tga.c:254:37

SUMMARY: AddressSanitizer: heap-buffer-overflow /tmp/repo/libgd/src/gd_tga.c:293 read_image_tga
Shadow bytes around the buggy address:
  0x36aa07a0: 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00
  0x36aa07b0: 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00
  0x36aa07c0: 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00
  0x36aa07d0: 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00
  0x36aa07e0: 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00
[...]

PR #250, opened 2016-07-12
Fix 247

oerdnj:
Fixes #247 

Before:
```
$ ./tga/bug00247 
=================================================================
==128235==ERROR: AddressSanitizer: heap-buffer-overflow on address 0x62500000e91c at pc 0x7fcfa43dd2d6 bp 0x7ffd36a97a90 sp 0x7ffd36a97a88
READ of size 4 at 0x62500000e91c thread T0
    #0 0x7fcfa43dd2d5 in gdImageCreateFromTgaCtx /home/ondrej/Projects/pkg-libgd/gd-libgd/src/gd_tga.c:103:5
    #1 0x7fcfa43dbcaf in gdImageCreateFromTga /home/ondrej/Projects/pkg-libgd/gd-libgd/src/gd_tga.c:25:10
    #2 0x4bd889 in main /home/ondrej/Projects/pkg-libgd/gd-libgd/tests/tga/bug00247.c:10:10
    #3 0x7fcfa1d96b44 in __libc_start_main /build/glibc-uPj9cH/glibc-2.19/csu/libc-start.c:287
    #4 0x4bd57c in _start (/home/ondrej/Projects/pkg-libgd/gd-libgd/tests/tga/.libs/lt-bug00247+0x4bd57c)

0x62500000e91c is located 0 bytes to the right of 8220-byte region [0x62500000c900,0x62500000e91c)
allocated by thread T0 here:
    #0 0x49fefb in malloc (/home/ondrej/Projects/pkg-libgd/gd-libgd/tests/tga/.libs/lt-bug00247+0x49fefb)
    #1 0x7fcfa44573c8 in gdMalloc /home/ondrej/Projects/pkg-libgd/gd-libgd/src/gdhelpers.c:75:9
    #2 0x7fcfa43e0aca in read_image_tga /home/ondrej/Projects/pkg-libgd/gd-libgd/src/gd_tga.c:226:24
    #3 0x7fcfa43dc42c in gdImageCreateFromTgaCtx /home/ondrej/Projects/pkg-libgd/gd-libgd/src/gd_tga.c:74:6
    #4 0x7fcfa43dbcaf in gdImageCreateFromTga /home/ondrej/Projects/pkg-libgd/gd-libgd/src/gd_tga.c:25:10
    #5 0x4bd889 in main /home/ondrej/Projects/pkg-libgd/gd-libgd/tests/tga/bug00247.c:10:10
[...]

PR #251, opened 2016-07-12
Fix 248

oerdnj:
fixes output buffer overflow in #248


oerdnj:
@cmb69 Please review, if you have a tad bit time.

src/gd_tga.c
```diff
@@ -273,26 +273,44 @@ int read_image_tga( gdIOCtx *ctx, oTga *tga )
...
-
+			
 			if ((decompression_buffer[buffer_caret] & TGA_RLE_FLAG) == TGA_RLE_FLAG) {
 				encoded_pixels = ( ( decompression_buffer[ buffer_caret ] & 127 ) + 1 );
 				buffer_caret++;
 
-				for (i = 0; i < encoded_pixels; i++) {
-					for (j = 0; j < pixel_block_size; j++, bitmap_caret++) {
-						tga->bitmap[ bitmap_caret ] = decompression_buffer[ buffer_caret + j ];
+				if (encoded_pixels > 0) {
```
cmb69: `encoded_pixels` is declared as `uint8_t`, so this would always be true. Either the if statement is useless, or it should be `encoded_pixels < 128`. Same issue below.

tests/tga/Makemodule.am
```diff
@@ -1,7 +1,9 @@
 libgd_test_programs += \
 	tga/bug00084 \
+	tga/bug00248 \
```
cmb69: Please also add this test to test/tga/CMakeLists.txt.

src/gd_tga.c
oerdnj: Nope, it won't be always true, but it could be changed to `if (encoded_pixels != 0)` for clarity. (It might also be an useless optimization as `encoded_pixels` would be always `!= 0` in normal case).

tests/tga/Makemodule.am
oerdnj: Done, thanks. (I did the same in #250)

src/gd_tga.c
cmb69: > Nope, it won't be always true, […]

Of course! Sorry for the noise.

oerdnj: I changed it to `!= 0` anyway to make more clear (although I suspect any decent compiler would do that for me anyway).

cmb69:
While I assumed that bug00248.tga is corrupt, both Gimp and Imagick's display open it without warning (contrary to bug00247.tga). Not sure, if Gimp and Imagick do advanced error correction here.

PR #261, opened 2016-07-17
Fix #252: gd_error() garbles variable arguments

### LINKED ISSUE #252: gd_error() garbles variable arguments
Apparently, gd_error() garbles its variable arguments (at least sometimes). Reproduce script:

    #include <stdio.h>
    #include "gd.h"
    #include "gd_errors.h"
    
    int main()
    {
        gd_error_ex(GD_WARNING, "The answer is %i\n", 42);
        gd_error("The answer is %i\n", 42);
        return 0;
    }

While one would expect to see the same message twice, gd_error() reports some random number instead of 42. I'm experiencing this with gcc version 4.9.2 (Debian 4.9.2-10) in debug mode.

cmb69:
Currently gd_error() forwards to gd_error_ex(). However, both functions
accept a variable number of arguments, and simply forwarding the va_list
isn't portable, see <http://c-faq.com/varargs/handoff.html>. This article
also describes the usual workaround, namely to let the second function
accept a va_list instead of variable number of arguments.

We do so by introducing a static helper, what does not affect API/ABI
compatibility.


cmb69:
The build failure is unrelated to this PR. Apparently, [`AM_ICONV` is unknown](https://travis-ci.org/libgd/libgd/jobs/145351736#L136) on the Travis OSX setup.

vapier:
adding a test in one commit and then deleting it in a follow up commit because it fails is bad git workflow.  you should have squashed the two into one commit.

git history should be "clean" in the sense that you don't introduce regressions/failures.  if you try to git bisect, you'll get stuck on the bad commit.

cmb69:
> adding a test in one commit and then deleting it in a follow up commit because it fails is bad git workflow. you should have squashed the two into one commit.

I agree. Sorry, won't happen again. (Even though squashing commits of a published PR seems to require a force push, what is also bad.)

vapier:
doing a force push on a PR isn't a big deal.  the history isn't lost, and you don't have to worry about the "downstream" users getting rewritten history because it's rare for anyone to take a PR and do work on top of it.

cmb69:
Well, I was afraid of the latter issue, but you're right, that's not supposed to happen often, if at all.

Issue #258, opened 2016-07-17
Shapes of gdImagePolygon and gdImageFilledPolygon can differ


cmb69:
I discovered that gdImagePolygon() and gdImageFilledPolygon() do not necessarily draw the same shape. Namely, horizontal edges are sometimes suppressed by the latter (what is a general case of PR #246, which I recently submitted and merged).

The following test script demonstrates the behavior:

```
<?php

$im = imagecreatetruecolor(200,200);

$full = imagecolorallocate($im, 255, 255, 255);
$transparent = imagecolorallocatealpha($im, 255, 255, 255, 80);

$coords = [10,50, 50,10, 50,50, 90,50, 50,50, 50,90, 10,50];

imagepolygon($im, translate($coords, 0, 0), count($coords)/2, $full);
imagefilledpolygon($im, translate($coords, 0, 100), count($coords)/2, $full);

imagepolygon($im, translate($coords, 100, 0), count($coords)/2, $transparent);
imagefilledpolygon($im, translate($coords, 100, 100), count($coords)/2, $transparent);

imagepng($im, 'actual.png');

function translate($points, $dx, $dy)
{
    $result = [];
    for ($i = 0; $i < count($points); $i++) {
        $result[] = $points[$i] + ($i % 2 ? $dy : $dx);
    }
    return $result;
}
```

Expected result:
![expected](https://cloud.githubusercontent.com/assets/2306138/16900131/e56f0bdc-4c1b-11e6-8265-d22eda702e4e.png)

Actual result:
![actual](https://cloud.githubusercontent.com/assets/2306138/16900134/e9725734-4c1b-11e6-8455-508b63b23d62.png)

When alpha is involved, the expected result is not really correct, but at least it is consistent with gdImagePolygon(). The fix which I used to create expected.png would be simple: instead of only [continue'ing](https://github.com/libgd/libgd/blob/gd-2.2.2/src/gd.c#L3185), we'd also have to draw a line:

```
gdImageLine(im, p[ind1].x, y1, p[ind2].x, y2, fill_color);
```

I'm not absolutely convinced that this fix should be applied, but a [comment in the sources](https://github.com/libgd/libgd/blob/gd-2.2.2/src/gd.c#L3188-L3190) suggests doing so would be correct. If we don't fix the behavior we most likely should also revert PR #246, which might not be necessary anyway, if we apply this fix.

pierrejoye:
[...]

PR #262, opened 2016-07-17
No pixel of a filled arc must ever be drawn multiple times

cmb69:
Otherwise we get artifacts regarding transparency. That happens with the
current implementation of gdImageFilledArc() unless gdChord or gdNoFill
are set. When gdPie is set, however, the filled arc is drawn in wedges,
which are polygons of three points, and so some overlap is natural.

To resolve the issue, we stick with the current algorithm of calculating the
wedges, but instead of drawing each polygon separately, we put the *relevant*
points in a large array, and draw a single polygon. That also is supposed to
improve the performance considerably.

Note that this modification will change the results when gdImageSetStyle()
or gdImageSetBrush() are used, but we believe that this modification is
also an improvement in this regard, even though it still might not make much
sense to use these functions with gdImageFilledArc().


src/gd.c
```diff
@@ -1740,8 +1738,8 @@ BGD_DECLARE(void) gdImageArc (gdImagePtr im, int cx, int cy, int w, int h, int s
 BGD_DECLARE(void) gdImageFilledArc (gdImagePtr im, int cx, int cy, int w, int h, int s, int e,
 									int color, int style)
 {
-	gdPoint pts[3];
-	int i;
+	gdPoint pts[363];
```
cmb69: This might be too demanding (~3KB on typical 32/64bit).

cmb69:
I've run the PHP GD test suite with this patch, and at least 6 tests using imagefilledarc() produce exactly the same filled arc as before. These are all rather small arcs (at most the size of the one in the test case for this issue), but some quick tests with relatively large arcs (rx = ry = 1199) show reasonable results. Very large arcs are most likely rather cornered anyway, due to insufficient precision (only 360 points); we might consider to use floats/doubles in the future. Unfortunately, the libgd test suite doesn't have any tests regarding filled arcs yet.

Note that this patch is supposed to be unaffected by issue #258, as horizontal oscillation is filtered out by the modified gdImageFilledArc() implementation.

cmb69:
For reference: the [PHP test case](https://github.com/php/php-src/blob/php-7.0.8/ext/gd/tests/imagecolorallocatealpha_basic.phpt) affected by this bug/fix produced before:
imagecolorallocatealpha_basic old

After applying the fix:
imagecolorallocatealpha_basic exp

pierrejoye:
Can you make it for 2.2.3?

cmb69:
> Can you make it for 2.2.3?

Yes. I'll also apply the fix to PHP's bundled libgd.

PR #270, opened 2016-07-21
Remove the huge test file, generate it on the fly instead

cmb69:


tests/gd2/php_bug_72339.c
```diff
@@ -8,14 +8,26 @@ int main()
...
-	fp = gdTestFileOpen2("gd2", "php_bug_72339_exp.gd2");
-	im = gdImageCreateFromGd2(fp);
-	if (im == NULL) {
-		return 0;
-	} else {
-		gdTestErrorMsg("Image should have failed to be loaded\n");
-		return 1;
+	long i;
+
+	/* we're creating the test image dynamically, due to it's size */
```
vapier: "its"

tests/gd2/php_bug_72339.c
```diff
@@ -8,14 +8,26 @@ int main()
...
+	fp = gdTestTempFp();
+	fputc(0x00, fp); fputc(0x02, fp);
+	fputc(0x00, fp); fputc(0x01, fp);
+	fputc(0x00, fp); fputc(0x01, fp);
+	fputc(0x00, fp); fputc(0x40, fp);
+	fputc(0x00, fp); fputc(0x02, fp);
+	fputc(0x54, fp); fputc(0xA0, fp);
+	fputc(0x5B, fp); fputc(0x00, fp);
+	for (i = 0; i < 0x4000000L; i++) {
```
vapier: the "L" suffix is unnecessary

tests/gd2/php_bug_72339.c
```diff
@@ -8,14 +8,26 @@ int main()
...
+		fputc(0x41, fp);
 	}
+	rewind(fp);
+
+	im = gdImageCreateFromGd2(fp);
+	fclose(fp);
+	
+	gdTestAssertMsg(im == NULL, "Image should have failed to be loaded%c", '\n');
```
vapier: what's with the %c weirdness ?

tests/gd2/php_bug_72339.c
```diff
@@ -8,14 +8,26 @@ int main()
...
-	if (im == NULL) {
-		return 0;
-	} else {
-		gdTestErrorMsg("Image should have failed to be loaded\n");
-		return 1;
+	long i;
+
+	/* we're creating the test image dynamically, due to it's size */
+	fp = gdTestTempFp();
+	fputc(0x00, fp); fputc(0x02, fp);
```
vapier: can't you declare a char[] array for all these bytes and then fwrite them in one go rather than calling fputc many times ?

tests/gd2/php_bug_72339.c
cmb69: That's a workaround, because gdTestAssertMsg() fails to compile with less than 3 arguments.

tests/gd2/php_bug_72339.c
cmb69: I thought as i is declared as long, it would be cleaner to add the respective suffix.

tests/gd2/php_bug_72339.c
cmb69: Thanks.

tests/gd2/php_bug_72339.c
cmb69: Yes, that should be possible. I can make an update later (have a meeting shortly).

tests/gd2/php_bug_72339.c
vapier: so change `i` to an `int` ?  i don't see a need for it to be `long` since the range is less than 30 bits.

tests/gd2/php_bug_72339.c
vapier: that's just silly.  i've fixed this in commit 4cd4a9b4c457259596669e8a3a279e02799baff5.

cmb69: Thanks! :-)

tests/gd2/php_bug_72339.c
cmb69: Okay, probably no need to support 16bit systems. :-)

vapier: yeah, i don't think we support systems where sizeof(int) < 4.  we might want to enforce it in configure if it comes up.

vapier:
lgtm

Issue #164, opened 2015-05-30
parity with CMake infrastructure vs autotools


brad0:
Trying to convert the OpenBSD port / package over from the autoconf build infrastructure to CMake I have found some bits that are not installed which should be and some other nits..

bin/gdlib-config
include/entities.h
include/gdpp.h
lib/pkgconfig/gdlib.pc

And these files are installed but right into share/docs instead of a subdirectory plus the more common name is share/doc... so could they be installed into share/doc/gd?

share/docs/
share/docs/INSTALL
share/docs/README.CMAKE
share/docs/README.JPN
share/docs/README.TESTING
share/docs/README.TXT
share/docs/arc.c
share/docs/copyrotated.c
share/docs/crop.c
share/docs/flip.c
share/docs/gif.c
share/docs/nnquant.c
share/docs/noIcon.pic
share/docs/noIcon.sgi
share/docs/noIcon.tga
share/docs/noIconAlpha.tga
share/docs/test_crop_threshold.png
share/docs/tgaread.c
share/docs/tiffread.c
share/docs/windows.c

pierrejoye:
Is it still an issue?

vapier:
the SONAME part can be solved by doing something like:
```
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
 if (BUILD_SHARED_LIBS)
   add_library(${GD_LIB} ${LIBGD_SRC_FILES})
+  set_target_properties(${GD_LIB} PROPERTIES
+       SOVERSION 3.0.8
+       VERSION 3
+  )
 endif()
```

we'll want to change the hardcoded versions though to use the variables set up in the top level CMakeLists.txt file, but that has versions hardcoded too.  so we need to fix that to use the same scripts that the autotools build does for pulling the version out of the header files.  i'll keep poking.

amadio:
Hi, I would like to add that installing the CMake module to find gd would be nice to have. In Gentoo, for example, the ebuild (or a normal build, for that matter) does not install `FindGD.cmake`, which could be used by dependent projects if available.

vapier:
please file a sep issue about that.  we don't install `FindGD.cmake` anywhere.

amadio:
If the module is not supposed to be installed, that's fine. It's still possible to use [FindPkgConfig](https://cmake.org/cmake/help/latest/module/FindPkgConfig.html) module from CMake to find libgd in another project. No need to open another issue for this. Thanks.

vapier:
[...]

PR #274, opened 2016-07-23
cmake: unify package & library versioning info #164

vapier:
This brings cmake up to par with the autotools builds and handling of version information.


pierrejoye:
What do you think about using:

cat src/gd.h | grep ^'#define GD_MAJOR_VERSION' |awk '{print $3}'

so we do not have to install perl anymore?

vapier:
i don't have a prob rewriting `getver.pl` as `getver.sh`

on the autotools side, this isn't an issue because the `configure.ac`->`configure` process itself requires perl (via autotools), and `getver.pl` is only executed at `autoconf` time.  it's new on the cmake side because there is no bootstrap process.

we could update `bootstrap.sh` to do the insertion into the cmake files too, but since cmake doesn't have a bootstrap process otherwise, it doesn't feel quite right.

pierrejoye:
Ah right. I will create another script for windows and maybe for cmake.
Could ve handy.

vapier:
you mean a replacement of getver.pl ?  we could have bootstrap.sh generate/update files that get committed so the windows files stay in sync.

vapier:
did you want me to change something here ?  or shall we move forward with this version and we can land follow up improvements as need be ?

pierrejoye:
We are fine I think.

Thanks :)

PR #283, opened 2016-08-04
Resolve #282: COPYING vs. docs/naturaldocs/license.txt

### LINKED ISSUE #282: COPYING vs. docs/naturaldocs/license.txt
I've noticed that [COPYING](https://github.com/libgd/libgd/blob/gd-2.2.3/COPYING) and [license.txt](https://github.com/libgd/libgd/blob/gd-2.2.3/docs/naturaldocs/license.txt) have been out of sync. I've fixed that with 8b428fd.

However, I think it would be best to have only COPYING, and to create license.txt dynamically, if that's 
even necessary. IMO the [preamble](https://github.com/libgd/libgd/blob/gd-2.2.3/docs/naturaldocs/license.txt#L1-L9) in license.txt could be added to COPYING, and marking the license up as code feels unnecessary, to say the least.

cmb69:
Basically, docs/naturaldocs/license.txt is a duplicate of COPYING, which
is going to be used only by Natural Docs. We remove the file and generate
it on the fly instead; also we adjust COPYING to make it easier to obtain
a pleasing result without sacrificing COPYING's readability.

The diff is best viewed with whitespace changes ignored.


vapier:
lgtm

PR #284, opened 2016-08-04
Hack to work around NaturalDocs encoding issues

cmb69:
NaturalDocs appears to assume that its input is ANSI encoded, so feeding
in UTF-8 encoded files will result in faulty character conversions even
though the system locale is set to UTF-8. This can be seen on
<https://libgd.github.io/manuals/2.2.3/files/license-txt.html>, where
Jaakko Hyvätti's name is garbled.

We work around this issue by explicitly iconv'erting the license file.
Other files don't seem to be affected, as these use ASCII characters only.

### DIFF
--- a/docs/naturaldocs/run_docs.sh
+++ b/docs/naturaldocs/run_docs.sh
@@ -64,6 +64,9 @@ echo "Title: License" | cat - ../../COPYING > tmp/license.txt
 sed -e "s/@VERSION@/$VERSION/g" preamble.txt > tmp/preamble.txt
 # ^^^ hack to get the version number in the docs.
 
+# Hack to work around NaturalDocs encoding issues
+iconv -f UTF-8 -t ISO-8859-1 -o tmp/license.txt tmp/license.txt
+
 # Run naturaldocs to create the manual.
 $(nd) --rebuild --rebuild-output --documented-only \
     -i tmp/ \


vapier:
"ANSI" isn't an encoding ... i guess you meant ASCII ?

that just switches to using ISO-8859-1 encoding which will also be garbled if you try to view it as UTF-8.  if it works in your browser, it's probably due to compatibility hacks many browsers have where they try to autodetect the encoding.  that license html though explicitly declares the output as UTF-8.

long story short, everything should be UTF-8 (or ASCII, which is the same as UTF-8 for the first 128 encodings).

cmb69:
> "ANSI" isn't an encoding ... i guess you meant ASCII ?

Actually, I meant ISO-8859-*. I'm not really sure whether it's actually ISO-8859-1 or ISO-8859-15 or may another one (might by even Windows CP-1252).

The problem is that NaturalDocs seems to assume the input is ISO-8859-*, even though it is UTF-8, and converts it to UTF-8. Of course, that can't work; therefore the hack to manually converting the file to ISO-8859-1 first.

vapier:
wrt ä, [ISO-8859-1](https://en.wikipedia.org/wiki/ISO/IEC_8859-1) and [ISO-8859-15](https://en.wikipedia.org/wiki/ISO/IEC_8859-15) are the same -- it's a byte with the value 0xE4

i'm running the bootstrap on my side and i'm not seeing corrupted output:
```
$ ./run_docs.sh
...
$ grep -o 'Jaakko [^ ]*' ./html/files/license-txt.html
Jaakko Hyvätti
$ grep -o 'Jaakko [^ ]*' ./html/files/license-txt.html | hexdump -C
00000000  4a 61 61 6b 6b 6f 20 48  79 76 c3 a4 74 74 69 0a  |Jaakko Hyv..tti.|
```

i have NaturalDocs-1.52 installed and my active locale is `en_US.UTF8`.

cmb69:
> i have NaturalDocs-1.52

I was still on 1.51; updating to 1.52 solved the issue! @pierrejoye Please update also on the machine where the docs are build; obviously this is still running 1.51.

PR #296, opened 2016-09-06
Fix integer overflow in gdImageWebpCtx

trylab:
Integer overflow can be happened in expression ``gdImageSX(im) * 4 * gdImageSY(im)``. It could lead to heap buffer overflow in the following code. 

This issue has been reported to the PHP Bug Tracking System. The proof-of-concept file will be supplied some days later. 

This issue was discovered by Ke Liu of Tencent's Xuanwu LAB.


cmb69:
Thanks!

PR #305, opened 2016-09-19
gdlib-config: punt!

vapier:
We've been shipping the pkg-config file for a while now and
want people to use that only, so drop this old script.

This also drop the FEATURES variable as it was only available
via `gdlib-config --features` (and no one has complained).


vapier:
i'll merge this soonish if no one has a reason to keep it.  this is for GD-2.3+.

PR #312, opened 2016-09-25
Fix #295: Catch libpng warnings

### LINKED ISSUE #295: Catch libpng warnings
When reading and writing PNG images with [`png_create_read_struct()`](https://github.com/libgd/libgd/blob/gd-2.2.3/src/gd_png.c#L230-L234) and [`png_create_write_struct()`](https://github.com/libgd/libgd/blob/gd-2.2.3/src/gd_png.c#L763-L769), respectively, we're catching errors with `gdPngErrorHandler`, but ignore warnings (last parameter == NULL). From the [libpng manual](http://www.libpng.org/pub/png/libpng-manual.txt):

> On non-fatal errors, png_warning() is called
to print a warning message, and then control returns to the calling code.
By default png_error() and png_warning() print a message on stderr via
fprintf() unless the library is compiled with PNG_NO_CONSOLE_IO defined
(because you don't want the messages) or PNG_NO_STDIO defined (because
fprintf() isn't available).

This appears to be undesirable, so we should set up a `gdPngWarningHandler`.

See also the respective <https://bugs.php.net/bug.php?id=71316>, which includes a test case that triggers a warning with libpng 1.6.x, but not with older versions (at least 1.2.x is silent).

cmb69:
Uncaught libpng warnings will either be printed to stderr or suppressed,
neither of which is desirable. Therefore we add a libpng warning handler,
similar to the already existing error handler, but the warning handler must
not `longjmp`.


cmb69:
Note that I consider this issue to be more a bug than an enhancement, so IMHO the patch should applied to GD-2.2 as well.

tests/png/bug00295.c
```diff
@@ -0,0 +1,45 @@
...
+ */
+
+
+#include "gd.h"
+#include "gd_errors.h"
+#include "gdtest.h"
+#include <string.h>
+
+
+#define EXPECTED_MSG "gd-png: libpng warning: iCCP: known incorrect sRGB profile"
```
vapier: the exact message depends upon the exact string libpng emits, and the exact code paths that libpng goes through.  do we really want to tie ourselves to that ?

tests/png/bug00295.c
```diff
@@ -0,0 +1,45 @@
...
+
+
+static void error_handler(int priority, const char *format, va_list args);
```
vapier: just move the func up here and you don't need the forward decl

tests/png/bug00295.c
cmb69: Thanks, good catch. Thinking about it, *this* test shouldn't be concerned about the message at all, and not even about GD_WARNING, but rather that there's no GD_ERROR.

tests/png/bug00295.c
cmb69: I'd welcome a general style guide for libgd! :-)

cmb69:
I've made the requested improvements. 

I'd prefer to squash the commits immediately before merging the PR (if at all).

tests/png/bug00295.c
```diff
@@ -0,0 +1,39 @@
...
+
+#include "gd.h"
+#include "gd_errors.h"
+#include "gdtest.h"
+#include <string.h>
+
+
+static void error_handler(int priority, const char *format, va_list args)
+{
+    char msg[256];
```
vapier: do we have to worry about 256 being too low ?

tests/png/bug00295.c
```diff
@@ -0,0 +1,39 @@
...
+
+    gdSetErrorMethod(error_handler);
+    fp = gdTestFileOpen2("png", "bug00295.png");
+
+    im = gdImageCreateFromPng(fp);
+
+    fclose(fp);
+    gdImageDestroy(im);
+
+    return gdNumFailures();
```
vapier: what guarantees that `error_handler` was called ?  should we have a global boolean like `called` that starts off `false` and `error_handler` sets to `true` ?

cmb69: There is no guarantee that `error_handler` will be called, and actually libpng < 1.6 wouldn't throw an error (and the behavior might change in a future version again; might be configurable). Therefore I think the test is okay: we're just checking that no `GD_ERROR` will ever be raised.

vapier: can we do a version check ?  if it's less than 1.6, don't require it to be set, but if it's >=1.6, require it ?

can we feed it a more damaged image that'll more strongly guarantee libpng calls us ?

tests/png/bug00295.c
cmb69: AFAIK libpng makes no guarantee regarding the max length of the error messages, so, yes, we have to worry. Is `vsnprintf` portable enough for our purposes?

vapier: imo, if it's in POSIX, it's fine.  if we find a system that lacks the function, we can figure out how to make the function available to it as a fallback.  but better to just wait for someone to complain than always try to make things work for a hypothetical setup that never materializes.

vsnprintf is in POSIX: http://pubs.opengroup.org/onlinepubs/9699919799/functions/vsnprintf.html

it shouldn't be hard to find replacements out there we can drop in (like [gnulib](https://www.gnu.org/software/gnulib/)).

tests/png/bug00295.c
vapier: to be clear, i'm not saying you must do this ... it'd just be a nice to have since the current test doesn't actually hard verify that we set the error handler.  if we broke things in the future, this test case wouldn't help catch regressions.  i.e. i'm pretty sure it'll "pass" if you build it against a version of gd that doesn't have your fixes here.

tests/png/bug00295.c
cmb69: > imo, if it's in POSIX, it's fine. […]

Sounds good to me, thanks. I'll switch to `vsnprintf` ASAP.

cmb69:
@vapier I've finally managed to rewrite the regression test so that it actually tests that the error handler is called exactly once from `gdPngWarningHandler()` (ignoring additional calls from `gdPngErrorHandler()` which are follow up errors).

I've rebased onto master and squashed the commits. Okay to merge?

woho:
Could it be that the merge of this fix was forgotten? I cannot find the commit on the current master branch.

Issue #216, opened 2016-06-05
Adding pdiff to tests suite


pierrejoye:
Adding a port of pdiff (bundled) to the tests suite to better fit our needs in the testing area. pdiff is as of today the best available tool for this purpose and can be used when work with compression or other operations where approximation may result in slight differences while visually the same.

It is GPL but it is fine as long as it is only part of the testing suite.

cmb69:
I'm not sure if [pdiff](http://pdiff.sourceforge.net/) would actually suit *our* needs. I'm afraid it would report too many false positives due to its advanced algorithm. Also [pHash](http://www.phash.org/) would probably be unsuitable.

Actually, we're mostly interested in *pixel-exact* operations. The only exceptions would be testing of the lossy output formats, and some precision/rounding issues (particularly, the freetype rending appears to be affected). The latter should ideally be solved, though.

Anyhow, I wonder if we won't come a long way with a simple [euclidean distance calculation](http://www.math.ttu.edu/outreach/sumac/pics/lhs.pdf). Besides that it's easy to implement (and maintain), it would primarily check for pixel exactness, but also allow for some differences (according to some given threshold).

cmb69:
I had a closer look at pdiff and an alternative Euclidean distance comparison, see <https://cmb69.github.io/pdiff/index.html>.

TL;DR: I don't think that we should add pdiff to our test framework.

pierrejoye:
Why not?

Euclidean may not fit for operations like filter or geometrical
transformations.

Also Cairo, which is all about rastering, do use pdiff. I fail to
understand why it would not math for us. Do you have cases why it does not
fit?

cmb69:
> Do you have cases why it does not fit?

See the examples and my conclusion on <https://cmb69.github.io/pdiff/index.html>.

pierrejoye:
@cmb69 is right. We can revisit this topic if needed when usable alternatives exist.

PR #326, opened 2016-10-08
Add basic test for writing and reading WebP

cmb69:
We do not compare the images, though, but rather offer a switch so the
images can be saved for manual inspection.


cmb69:
Interestingly, comparing the two images with perceptualdiff outputs:
````
Field of view is 45.000000 degrees
Threshold pixels is 100 pixels
The Gamma is 2.200000
The Display's luminance is 100.000000 candela per meter squared
Converting RGB to XYZ
Constructing Laplacian Pyramids
Performing test
PASS: Images are perceptually indistinguishable
0 pixels are different
````
So at least for this case #216 would make sense.

PR #332, opened 2016-10-11
Patch for bug https://bugs.php.net/bug.php?id=73280

libnex:
See https://bugs.php.net/bug.php?id=73280 for more info


vapier:
that bug is not public.  but even if it was, please put all relevant info into the commit message itself.

cmb69:
Thanks for the patch! Applied with commit 53110871.

vapier:
thanks, that commit message is much better :)

PR #346, opened 2016-11-16
Fix #338: Fatal and normal libjpeg/ibpng errors not distinguishable

### LINKED ISSUE #338: Fatal and normal libjpeg/ibpng errors not distinguishable
This issue has been reported in <https://bugs.php.net/bug.php?id=73479>. Basically it boils down to [fatal_jpeg_error()](https://github.com/libgd/libgd/blob/gd-2.2.3/src/gd_jpeg.c#L100-L119) raising GD_ERROR regardless whether carrying on is possible or not. It seems to me that in the former case a GD_WARNING would be more appropriate, so that clients are able to distinguish both cases. An alternative might be to introduce GD_CRIT to indicate the "extremely fatal error".

[gdPngErrorHandler()](https://github.com/libgd/libgd/blob/gd-2.2.3/src/gd_png.c#L52-L74) has the same issue.

cmb69:
libgd clients need to be able to distinguish between fatal and
"extremely fatal" libjpeg and libpng errors, because in the former case
execution can proceed, but in the latter case libgd calls exit().
Therefore we report fatal errors as GD_WARNING.


cmb69:
Are there any objections to merge that into GD-2.2?

pschlan:
In my opinion this needs to be merged urgently as it may cause critical problems in PHP-based applications, see comment in issue #338

PR #353, opened 2016-12-05
[CVE-2016-6911]: Fix invalid read in gdImageCreateFromTiffPtr()

omron93:
Posting fix for CVE.

https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=840806
https://bugzilla.redhat.com/show_bug.cgi?id=1388787
http://people.canonical.com/~ubuntu-security/cve/2016/CVE-2016-6911.html

Author is @cmb69, so why it is not in master?

Original patch:
```
From: "Christoph M. Becker" <cmbecker69@gmx.de>
Date: Tue, 2 Aug 2016 12:10:33 +0200
Subject: [CVE-2016-6911]: Fix invalid read in gdImageCreateFromTiffPtr()

tiff_invalid_read.tiff is corrupt, and causes an invalid read in
gdImageCreateFromTiffPtr(), but not in gdImageCreateFromTiff(). The culprit
is dynamicGetbuf(), which doesn't check for out-of-bound reads. In this case,
dynamicGetbuf() is called with a negative dp->pos, but also positive buffer
overflows have to be handled, in which case 0 has to be returned (cf. commit
75e29a9).

Fixing dynamicGetbuf() exhibits that the corrupt TIFF would still create
the image, because the return value of TIFFReadRGBAImage() is not checked.
We do that, and let createFromTiffRgba() fail if TIFFReadRGBAImage() fails.

This issue had been reported by Ibrahim El-Sayed to security@libgd.org.
...
```
(https://bugzilla.suse.com/attachment.cgi?id=697849)


cmb69:
> Author is @cmb69, so why it is not in master?

At least for me, the question is rather why this issue had already been publicly disclosed!

omron93:
Wow, I am not the right person to answer this :-(

@cmb69 Should I close this PR or something?

cmb69:
> Should I close this PR or something?

Well, the cat is out of the bag anyway, so it might be best to leave this PR open as a reminder for our RMs. :-)

cmb69:
This has now been committed as 4859d69e, so I'm closing the ticket.

PR #374, opened 2017-01-25
Enable test case on Windows

cmb69:
This test case has been disabled for Windows, because even MSVC 14
doesn't know ftello(), which is a Posix extension to the ISO standard.

Actually, ftell() is supposed to work fine for this test too, as we
don't have to deal with a large file, and we're comparing to a fixed
expected length, anyway.

### DIFF
--- a/tests/xbm/github_bug_170.c
+++ b/tests/xbm/github_bug_170.c
@@ -8,24 +8,16 @@
 	See also <https://github.com/libgd/libgd/issues/170>.
 */
 
-#include <inttypes.h>
 #include "gd.h"
 #include "gdtest.h"
-#ifdef _WIN32
 
-int main()
-{
-	/* skip for now */
-	return 0;
-}
-#else
 int main()
 {
 	gdImagePtr im;
 	int black;
 	FILE *outFile;
 	gdIOCtx *out;
-	off_t length;
+	long length;
 
 	/* create the test image */
 	im = gdImageCreate(11, 11);
@@ -39,12 +31,11 @@ int main()
 	gdTestAssert(out != NULL);
 	gdImageXbmCtx(im, "github_bug_170.xbm", 1, out);
 	out->gd_free(out);
-	length = ftello(outFile);
+	length = ftell(outFile);
 	fclose(outFile);
 
 	gdImageDestroy(im);
 
-	gdTestAssertMsg(length == 250, "expected to write 250 bytes; %jd bytes written", (intmax_t) length);
+	gdTestAssertMsg(length == 250, "expected to write 250 bytes; %ld bytes written", length);
 	return gdNumFailures();
 }
-#endif


cmb69:
@vapier You had [recommended to use `ftello()`](https://github.com/libgd/libgd/pull/229#r66795107) when I've submitted this test. Okay to use `ftell()` instead for compatibility with MSVC14?

pierrejoye:
The main differences are:

long ftell(FILE *stream);
off_t ftello(FILE *stream);

I would recommend to use ftello on non Windows and keep ftello on posix or if available.

vapier:
i don't have a MSVC env to check against.  for some systems, you have to define things like `_LARGEFILE_SOURCE` in order to get access to `ftello`.  does that work here ?

if not, stuffing into some internal common header things like:
```
#if defined(whatever windows marker)
#define ftello ftell
#endif
```

pierrejoye:
Yes, that will do it.

cmb69:
Hm, that appears to be confusing. If using `ftello()` I'd expect to get an `off_t`.

vapier:
sure, but if the system doesn't provide `ftello`, i'm not sure what you can do otherwise ?  i agree it's kind of ugly, but imo it's better than sprinkling the ifdef logic around every `ftello` call site.

cmb69:
> sure, but if the system doesn't provide ftello, i'm not sure what you can do otherwise ?

Actually, we don't even need `ftell(o)` for this *test case*. Writing to memory should work as well, and is more efficient, anyway. I'll make another PR ASAP.

vapier:
i'm not looking at just this test.  i'm suggesting we should use `off_t` everywhere instead of `long` which means a portability hack for Windows should be in a common place.

or we can say that gd never supports images larger than 2GiB in size.

pierrejoye:
I agree with Mike. I will try to look at how we could do it on Windows in
away than tests won't be borked because of long.

cmb69:
> i'm suggesting we should use off_t everywhere instead of long which means a portability hack for Windows should be in a common place.

[According to MSDN](https://msdn.microsoft.com/en-us/library/323b6b3k(v=vs.80).aspx) `off_t` is supported by MSVC since a very long time. It might be a 32-bit value, but AFAIK the `sizeof off_t` isn't specified by POSIX, either.

vapier:
POSIX defines `off_t`.  it doesn't require, by design, an explicit size.  it's guaranteed to be signed, and it says it's at least 32-bit, and LFS systems support larger.  but either way i don't see this as a problem ?

> http://pubs.opengroup.org/onlinepubs/9699919799/xrat/V4_xsh_chap02.html
> **off_t**: This type is used to represent a file offset or file size. On systems supporting large files, off_t is larger than 32 bits in at least one programming environment. Other programming environments may use different sizes for off_t, for compatibility or other reasons.

in looking at our API, we have `gdSeek` (which uses an `int`) and `gdTell` (which uses a `long`).  but neither are exported afaict.  so changing those shouldn't be a problem.

we also have `gdIOCtx` in `gd_io.h` with `seek` and `tell` callbacks (also using `int` and `long` respectively).  it's not clear to me whether we expect people to use that structure directly, or we want it to be an opaque handle.  if it's an opaque handle, we can stop installing gd_io.h to enforce the opaqueness, and then we can change the callbacks fine.

PR #375, opened 2017-01-25
Fix and reenable optimized support for reading 1 bps TIFFs

cmb69:
Due to #82 the optimized support for reading 1 bps TIFF files (black &
white) had been disabled. Tony Lew already pointed out a fix in #88.

Furthermore, there was the following missing and improper error handling:
* TIFFReadScanline() returns -1 on error, not 0
* the result of TIFFReadTile() hasn't been checked
* in case of failure of these functions, the error had not been
  propagated

We fix this, and re-enable direct support for 1 bps TIFFs, which is
more memory efficient than the general RGBA support. We also make sure
not to hit any not yet implemented code path.


cmb69:
Any objections to merge that? Should it go to master only, or also into GD-2.2? @pierrejoye 

Note that this "fixes" #88.

pierrejoye:
Master only I think. TIFF is experimental still 🤗

cmb69:
> Master only I think. TIFF is experimental still

Okay. Will `merge --squash` as soon as Travis is happy.

cmb69:
Merged with b4d153ba.

Issue #397, opened 2017-07-21
Missing header(s) from install


galik:
I tried compiling `libgd-2.2.4` using the `<gdpp.h>` header and it complains of a missing include:

>    /usr/include/gdpp.h:31:10: fatal error: gd_io_stream.h: No such file or directory

I am using Fedora package: gd-devel-2.2.4-1.fc25

vapier:
libgd does not use/link against C++ today.  this is an ABI change.  we can't deploy it for gd-2.2 which means we should delete the header from the install there.

for gd-2.3, we can add an optional configure flag.

galik:
Fair enough. I am actually writing my own `C++` wrapper around the `C` API. There is one change I would like to propose that should not break the current `ABI` but would make the library much more friendly to `C++`.

These functions:
```cpp
gdImageStringFT();
gdImageStringTTF();
gdImageStringFTEx();
```
They each take a `char*` input for both the name of the font and the text to be printed. The contents of the passed in string are never modified by the functions (or any of the functions called by them) so could those parameters be changed to `const char*` instead?

Otherwise they are very awkward to use from `C++` being that `C++` almost always uses `const char*` and casting away the `const` causes *undefined behavior* in most cases.

This means that when I call the font functions I have to make a physical copy of the font name and the string to be printed. This is hugely inefficient. So would that be an acceptable change of those function signatures?

Basically this function:
```cpp
		gdImageStringFT(gdImage * im, int *brect, int fg, char *fontlist,
            double ptsize, double angle, int x, int y,
            char *string)
```
would become
```cpp
		gdImageStringFT(gdImage * im, int *brect, int fg, const char *fontlist,
            double ptsize, double angle, int x, int y,
            const char *string)
```

I currently have a patch that makes this change and it doesn't appear to break anything. But unfortunately my patch also includes those missing `C++` headers and sources so I would need to fix that, if you're interested.

vapier:
please file new issues for changes to existing APIs, or just send PRs directly.  i don't see a problem with constifying strings where possible.

pierrejoye:
@galik good morning

Still working on the C++ wrapper? We do have one but I don't have the time or good enough knowledge to provide much support there. There is definitively a room for rock solid C++ API in libgd.

Happy to help as much as I can :)

PR #398, opened 2017-07-21
Added missing header gd_io_stream.h (fix issue #397)

galik:
Offered for consideration. Should fix Issue #397 

Updated `src/Makefile.am` and `src/CMakeLists.txt` to include the missing file.
Also moved header includes from all on one line to separate lines as it was getting a little long.


galik:
The Automake build was also missing the C++ source files leading to missing symbol errors so I have now added these. It seems my Linux distro (Fedora 25) packages this library using Automake.

vapier:
this won't fly as it changes the ABI/C++ linkage unconditionally.  let's keep the discussion in the referenced bug.

PR #400, opened 2017-07-24
Truetype font drawing functions now take `const` strings

galik:
The following API functions now accept the font names and the text to be
printed as `const char*` rater than `char*`. This makes the functions
much more `C++` fiendly. Otherwise they are very awkward to use from `C++` being that `C++` almost always uses `const char*` and casting away the `const` causes *undefined behavior* in most cases.

```cpp
gdImageStringFT();
gdImageStringTTF();
gdImageStringFTEx();
```
**Other functions/types affected:**
```cpp
typdef struct fontkey_t;

any2eucjp();
gdTcl_UtfToUniChar();
DetectKanjiCode();
do_convert();
do_check_and_conv();
```
**Changes made to these files:**
```cpp
gd.h
gdft.c
gdkanji.c
```

### DIFF
--- a/docs/naturaldocs/project/Menu.txt
+++ b/docs/naturaldocs/project/Menu.txt
@@ -65,28 +65,28 @@ Group: Image Formats  {
    File: XPM Input  (no auto-title, gdxpm.c)
    }  # Group: Image Formats
 
+File: Color Quantization  (gd_topal.c)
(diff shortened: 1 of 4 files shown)


vapier:
you'll want to look at the travis logs and fix the failures it found :)

vapier:
thanks!

PR #423, opened 2018-01-24
Avoid accessing uninitialized buffers on short reads

tdsmith:
llvm's memory sanitizer ("MSAN") can detect the use of uninitialized values. In cases where the provided input is very short, the strncmp for the GIF header can access uninitialized memory, which is inconsequential but can prevent the fuzzer from finding more interesting examples.

### DIFF
--- a/src/gd_gif_in.c
+++ b/src/gd_gif_in.c
@@ -53,7 +53,7 @@ static int set_verbose(void)
 
 #define BitSet(byte, bit)	(((byte) & (bit)) == (bit))
 
-#define ReadOK(file, buffer, len) (gdGetBuf(buffer, len, file) > 0)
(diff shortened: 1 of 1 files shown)


vapier:
i don't think this is how we want to address this.  i think this is just making some bugs stop showing up rather than fix it.

i think the real problem is with the `ReadOK` helper.  it returns success as long as some data was read regardless of how many bytes were requested.  looking at ever call site in this file, they're all wrong if we got a short read.  so how about we change `ReadOK` to require the result of `gdGetBuf` be equal to `len`.

tdsmith:
That seems plausible; the tests pass, anyway!

vapier:
thanks!

PR #425, opened 2018-01-26
Avoid undefined behavior when reading negative integers

tdsmith:
Left-shifts into signed integers which cause an overflow are undefined
behavior in C99.

Credit to OSS-Fuzz.

References:
https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=5693
https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=5694
https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=5695


tdsmith:
Oops, jinx.

Issue #426, opened 2018-01-26
Integrate fuzzing targets


tdsmith:
Integrate the fuzzing targets and configuration into the libgd repository as proposed at https://github.com/google/oss-fuzz/pull/1109 and https://github.com/google/oss-fuzz/blob/master/docs/ideal_integration.md.

I anticipate being able to do this work and thought this ticket could be a way to discuss progress; let me know if you'd rather coordinate some other way!

I think some to-dos are:
* [ ] Move the fuzzing target into the libgd repository and point the oss-fuzz build.sh at it
* [ ] Find a place to store fuzzing corpora; copy the relevant parts of the afl corpus into the libgd repository
* [ ] Automatically mix in existing test cases as part of the corpora
* [ ] Integrate with the build system?
* [ ] Build and quickly test the fuzzers against a [standalone driver](https://github.com/llvm-mirror/compiler-rt/tree/master/lib/fuzzer/standalone) for `make check`
* [ ] Suppress logging messages under libFuzzer (generate $OUT/xxx_target.options files containing `[libfuzzer]\nclose_fd_mask = 2\n`)

Thanks!

vapier:
i was thinking we add a configure flag for it so you can do --enable-fuzzing (or whatever the name) and it would build things with the right flags/libs and build the helpers

in the source layout, i'd be fine with just using fuzzers/ like we have with examples/ and tests/.  both the fuzzers and their input corpora can live there.  the corpora should be small (in disk space) of course.

tdsmith:
It's interesting that #383 discovered a bug on output, which suggests to me that maybe the fuzzers should also try emitting images (to /dev/null at least).

cmb69:
> […] that maybe the fuzzers should also try emitting images (to /dev/null at least).

+1

It also might be beneficial to add fuzzers for (some of) the image manipulation routines.

PR #427, opened 2018-01-26
Integrate fuzzer

tdsmith:
Some questions:
* Is this a good place to put the fuzzer targets and build scripts?
* It's nice to have a place to manually put reproducers for discovered bugs, so storing at least part of the corpus explicitly in the repository is useful. OTOH, the AFL gif and bmp corpuses are together about 1300 files and 7mb, which is unwieldy. Maybe I should go back to just fetching the AFL corpus from the web in the oss-fuzz Dockerfile and merging them in during the build. What do you think?
* Should this be more refined before it's merged or is it okay to work on it piecemeal?

cf #426


tdsmith:
These are the matching changes on the oss-fuzz side; https://github.com/google/oss-fuzz/compare/master...tdsmith:libgd-integration?expand=1

vapier:
i'd prefer fuzzers/ just to match examples/ and tests/

is this corpus something you built up or just fetched from somewhere else ?  if it's an existing project that maintains inputs, fetching & unpacking an archive is fine.

for pending commits, just squash them together rather than add files and then modify/shuffle things in follow up changes

should add a fuzzers/README.md to document things

tdsmith:
Sorry for the delay! I made these changes.

tdsmith:
Hi @vapier; is this a good place to merge, or would you like to see some additional progress? Compared to what's running now, this PR adds some additional corpora and mutes stderr, which should already help improve oss-fuzz's efficiency a little.

pierrejoye:
It would be great to have that in the github actions, thoughts?

pierrejoye:
I think we should merge, and add missing ones (all codecs are keys).

I see that libgd is already on ossfuzz, not sure what that means. I need to sit down reading their docs, how we can have that as part of the CI etc. :)

Issue #428, opened 2018-01-29
add a build time knob to disable gd2 image format


vapier:
considering we view the gd2 image format as more of a dev/testing space and not useful to in any real production scenario, should we add a configure flag for this (and have it be disabled by default) ?

cmb69:
That seems to be a prudent first step on the way to finally rid these obsolete formats (.gd2 as well as .gd). Maybe we should start with `disable-gd(2)` and later switch the default (i.e. `enable-gd(2)`). If anybody has a good use for `gdImageCreateFromGd2Part()` (the only reason I can think of for using .gd2; gd makes no sense to me at all), they still could object.

vapier:
i think #429 should do the trick

vapier:
last question: when should we flip the default ?  for gd-2.3 or gd-2.4 ?

cmb69:
> last question: when should we flip the default ? for gd-2.3 or gd-2.4 ?

Depends on our release schedule, IMHO. If we're going to release gd-2.3 rather soonish (say, in some months), I'd stick with the current default for now.

vapier:
we should prob snap gd-2.3 off this year

pghmcfc:
The default as shipped in 2.3.3 is that gd/gd2 image support is DISABLED. I suspect that that was not intentional. It is currently causing the perl GD and GDGraph module test suites to fail as they expect this support to be there. If I rebuild gd using `--enable-gd-formats` then the test suites work again.

PR #429, opened 2018-01-29
make gd/gd2 image formats optional #428

vapier:
We still enable them by default, but we'll probably flip them off
in the next major release series.


cmb69:
LGTM. Thanks!

PR #433, opened 2018-02-25
gd_xbm: fix undefined behavior for ctype functions

rillig:


vapier:
sorry, but what undefined behavior are you referring to here ?

the prototype of ctype funcs are that they take an `int c`.  `name` is of type `char *` which is usually signed.

this function only works on ASCII names which is 7-bit.  all chars in that range should work fine with implicit signed char to signed int conversions.

so are you passing in UTF-8 encoded names or something that otherwise is 8-bit ?

rillig:
Yes, I am doing exactly that. And in such a case, the program should not crash but just ignore these characters. Or even consider them as valid name characters.

vapier:
why would the program crash?  are you saying calling `isupper(-10)` on your system crashes?  that sounds like a bad bug in your system.

would be nice to add or update a unittest to make sure this behavior doesn't regress.

rillig:
No, my system doesn't crash on `isupper(-10)`. But it _may do so, at its own will_. You are using the C programming language, which allows the implementation to crash. And most implementations (glibc, ulibc, other libcs) don't explicitly document that they won't crash, therefore programming according to the C standard is the only safe choice.

vapier:
i'm aware of what the [POSIX & ISO C standards say](http://pubs.opengroup.org/onlinepubs/9699919799/functions/isalnum.html), but that doesn't mean a reasonable system would crash.  it practically costs nothing to handle this sanely in the C runtime.

just to be clear, i'm not against this change.  i would like to have unittest coverage before merging.

cmb69:
Possible unit test:
````c
/**
 * Test non ASCII characters in XBM names
 *
 * See <https://github.com/libgd/libgd/pull/433>
 */


#include <locale.h>
#include <string.h>
#include "gd.h"
#include "gdtest.h"


int main()
{
    gdImagePtr im;
    int fg;
    char *filename;
    FILE *fp;
    gdIOCtx *out;
    char buf[20];

    /* force C locale for best portability */
    gdTestAssert(setlocale(LC_CTYPE, "C") != NULL);

    /* create image */
    im = gdImageCreate(8, 8);
    gdImageColorAllocate(im, 255, 255, 255);
    fg = gdImageColorAllocate(im, 0, 0, 0);

    /* save image */
    filename = gdTestTempFile(NULL);
    fp = fopen(filename, "w+");
    gdTestAssert(fp != NULL);
    free(filename);
    out = gdNewFileCtx(fp);
    gdTestAssert(out != NULL);
    gdImageXbmCtx(im, "t\xc3\xbcr", fg, out); /* UTF-8 tür (= door) */

    /* read and verify */
    out->seek(out, 0);
    out->getBuf(out, buf, sizeof buf);
    gdTestAssert(!memcmp(buf, "#define t__r_width 8", sizeof buf));

    /* clean up */
    out->gd_free(out);
    fclose(fp);
    gdImageDestroy(im);

    return gdNumFailures();
}
````
The `memcmp()` is somewhat brittle; using `sscanf()` might be preferable.

File: src/gd_xbm.c [hunk 1]
```diff
@@ -239,7 +239,8 @@ BGD_DECLARE(void) gdImageXbmCtx(gdImagePtr image, char* file_name, int fg, gdIOC
 	} else {
 		for (i=0; i<l; i++) {
 			/* only in C-locale isalnum() would work */
-			if (!isupper(name[i]) && !islower(name[i]) && !isdigit(name[i])) {
+			unsigned char ch = name[i];
```
vapier, line 242: add a comment like "POSIX says ctype functions only accept values within [0..255]"

vapier (changes requested):
@rillig can you look at integrating the unittest so we can merge this ?

rillig:
I'm no longer interested in fixing these bugs myself.

PR #438, opened 2018-04-05
Close #435: Deprecate gdImageDashedLine()

### LINKED ISSUE #435: Deprecate gdImageDashedLine()?
The [former GD manual notes wrt. gdImageDashedLine()](https://boutell.com/gd/manual2.0.33.html#gdImageDashedLine):

> gdImageDashedLine is provided solely for backwards compatibility with gd 1.0. New programs should draw dashed lines using the normal gdImageLine function and the new gdImageSetStyle function.

However, we do not mention this anymore. This raises the question whether we should stick with `gdImageDashedLine()` or whether we should actually deprecate this function.

Note that downstream projects may already have heeded the deprecation. For instance, the [PHP manual](http://www.php.net/manual/en/function.imagedashedline.php) mentions that the slim wrapper `imagedashedline()` is deprecated (although, calling the function does not raise a deprecation warning).

cmb69:
`gdImageDashedLine()` is a legacy from GD 1, and has been superseded
by the more flexible `gdImageLine()` in combination with
`gdImageSetStyle()`.  We therefore deprecate `gdImageDashedLine()`:

* document the deprecation
* raise a compile time warning for supporting compilers
* raise a GD_NOTICE whenever the function is called

### DIFF
--- a/src/gd.c
+++ b/src/gd.c
@@ -1541,10 +1541,23 @@ static void gdImageVLine(gdImagePtr im, int x, int y1, int y2, int col)
 }
 
 /*
-	Function: gdImageLine
-	
-	Bresenham as presented in Foley & Van Dam.
-*/
+ * Function: gdImageLine
+ *
+ * Draws a line
+ * 
+ * Uses the Bresenham algorithm as presented in Foley & Van Dam.
+ *
+ * Parameters:
+ *   im    - The image.
+ *   x1    - The x-coordinate of the start point.
+ *   y1    - The y-coordinate of the start point.
+ *   x2    - The x-coordinate of the end point.
+ *   y2    - The y-coordinate of the end point.
+ *   color - The color.
+ *
+ * See also:
+ *   - <gdImageSetStyle>
+ */
 BGD_DECLARE(void) gdImageLine (gdImagePtr im, int x1, int y1, int x2, int y2, int color)
 {
 	int dx, dy, incr1, incr2, d, x, y, xend, yend, xdirflag, ydirflag;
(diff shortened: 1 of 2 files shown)


src/gd.c
```diff
@@ -1713,8 +1726,26 @@ static void dashedSet (gdImagePtr im, int x, int y, int color,
...
+ *   y2    - The y-coordinate of the end point.
+ *   color - The color.
+ *
+ * See also:
+ *   - <gdImageLine>
+ *   - <gdImageSetStyle>
+ */
 BGD_DECLARE(void) gdImageDashedLine (gdImagePtr im, int x1, int y1, int x2, int y2, int color)
 {
 	int dx, dy, incr1, incr2, d, x, y, xend, yend, xdirflag, ydirflag;
```
vapier: seems like we should be able to gut this function and make it call `gdImageLine` & `gdImageSetStyle` ?  it'd simplify, shrink the code (compiled & source), and be a stronger proof that this func is indeed fully redundant with those other two.

src/gd.h
```diff
@@ -76,7 +76,16 @@ extern "C" {
...
+#if defined(__GNUC__) || defined(__clang__)
+# define BGD_ATTR_DEPRECATED __attribute__((__deprecated__))
+#elif defined(_MSC_VER)
+# define BGD_ATTR_DEPRECATED __declspec(deprecated)
+#else
+# define BGD_ATTR_DEPRECATED
+#endif
+
 #define BGD_DECLARE(rt) BGD_EXPORT_DATA_PROT rt BGD_STDCALL
+#define BGD_DEPRECATED(rt) BGD_ATTR_DEPRECATED BGD_EXPORT_DATA_PROT rt BGD_STDCALL
```
vapier: can't you write:
`#define BGD_DEPRECATED(rt) BGD_ATTR_DEPRECATED BGD_DECLARE(rt)`

src/gd.h
```diff
@@ -726,7 +735,7 @@ BGD_DECLARE(void) gdImageLine (gdImagePtr im, int x1, int y1, int x2, int y2, in
 
 /* For backwards compatibility only. Use gdImageSetStyle()
    for much more flexible line drawing. */
-BGD_DECLARE(void) gdImageDashedLine (gdImagePtr im, int x1, int y1, int x2, int y2,
+BGD_DEPRECATED(void) gdImageDashedLine (gdImagePtr im, int x1, int y1, int x2, int y2,
                                      int color);
```
vapier: indentation need adjusting ?

src/gd.h
cmb69: Good catch! Will fix that later.

src/gd.h
cmb69: Okay, makes sense.

PR #439, opened 2018-04-18
handle new_width & new_height absolute exception

hg-pyun:
Since the cos range is a real number range, there is a possibility of negative numbers.
- handling fabs new_width & new_height

Reference: https://github.com/libgd/libgd/issues/440

Thanks.


cmb69:
Thanks! Merged via 6653f3b53557a19f2f13e7c3eab532f98894f036.

PR #442, opened 2018-06-20
annotate.c/gdft.c: Replace strncpy with memccpy to fix -Wstringop-tru…

robertlinux:
…ncation.

Fixed for gcc8:
git/src/gdft.c:1699:2: error: 'strncpy' output truncated before terminating nul copying as many bytes from a string as its length [-Werror=stringop-truncation]

Signed-off-by: Robert Yang <liezhi.yang@windriver.com>

### DIFF
--- a/src/annotate.c
+++ b/src/annotate.c
@@ -104,7 +104,7 @@ int main(int argc, char *argv[])
 					fprintf(stderr, "Font maximum length is 1024, %d given\n", font_len);
 					goto badLine;
 				}
-				strncpy(font, st, font_len);
+				memcpy(font, st, font_len);
 			}
 		} else if(!strcmp(st, "align")) {
 			char *st = strtok(0, " \t\r\n");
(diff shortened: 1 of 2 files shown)


cmb69:
Hmm, I'm not fond of replacing `strncpy` with `memcpy` just to rid some bogus compiler warnings. Actually, I'd prefer to use `strlcpy` instead, if available.

vapier:
the annotate code was buggy, before & after this commit.  i rewrote it in commit e9bf9ddf1d612227ef04a363ed1e0ad5e960b857 to fix all of this.

vapier:
the gdft change was OK, but i don't think it was fully analyzed before changing, and didn't change all the users in the file.  i've committed 748578ff3fa4b636513373f0613e158b930f3ed4 to fix it.

so everything in this PR should be handled in master now.  let us know how it goes.

robertlinux:
Hi vapier, thank you very much.

PR #443, opened 2018-06-22
travis: add some source/lint tests

vapier:
We probably should look into cpplint, but until then, these are
some simple hacks to catch common mistakes.

### DIFF
--- a/tests/source/README.md
+++ b/tests/source/README.md
@@ -1,0 +1,3 @@
+These are tests/lints for the source code itself.
+They aren't included in the default `make check` so they aren't run against
+releases, but we run them against Github PRs in Travis.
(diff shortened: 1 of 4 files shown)


cmb69:
Thanks. LGTM!

Before going further, we should adopt some coding style guide.

vapier:
strongly agree.  i'll see about sketching out a md file based on the style that seems to be active in the codebase.

i'm not familiar with any C/C++ linters other than [cpplint](https://github.com/google/styleguide/tree/gh-pages/cpplint).  not sure if we'll be able to configure that in a way that works for our codebase.

cmb69:
I'm not familiar with any C/C++ linters at all, so feel free to introduce whatever you prefer. However, for me it's way more important to have a documented coding style than to be able to automatically check the sources for compliance.

vapier:
i've come across [clang-tidy](http://clang.llvm.org/extra/clang-tidy/) that i think we'd be able to shape for our needs

PR #444, opened 2018-06-27
Remove Git ident attributes

petk:
Hello.

Git ident attributes were in most cases utilized with SVN and keywords substitutions, where $Id$ were replaced with certain revision from the repository. In Git this functionality is different. Each $Id$ needs to
be defined in .gitattributes file to be effective. This patch removes unused and outdated attributes.

Thanks for considering checking this out or merging it.


vapier:
makes sense to me

cmb69:
Thanks!

Issue #169, opened 2015-06-28
gdImageEllipse() has overflow issue for moderately large radii


cmb69:
On architectures where `long`s are signed 32bit values (e.g. LLP64), moderately large values of w and h (e.g. w=h=2200) cause gdImageEllipse() to overflow when [assigning to rx](https://github.com/libgd/libgd/blob/4e701317ff304ab6cb71c9f6c822944fd81bf965/src/gd.c#L1891), what results in broken ellipses. It appears to be sufficient to declare r,rx and ry as `long long`.

See the [bug reported downstream for PHP](https://bugs.php.net/bug.php?id=51498) for some more details.

vapier:
really anything >=2048 overflows due to the cube/squaring of the inputs.

```
// w = h = 2048 = 0x800
a = w >> 1;   // a = 0x400
b = h >> 1;   // b = 0x400
bq = b * b;   // bq = 0x400 * 0x400 = 0x100000
r = a * bq;   // r = 0x400 * 0x100000 = 0x40000000
rx = r << 1;  // rx = 0x80000000; signed overflow
```

i don't believe we use ```long long``` anywhere else so i don't think we want to start here.  my preference would be to use explicit ```uint64_t``` here ... we also don't use that internally, but it's in POSIX/C99, so we should be able to rely on it.

just marking r/rx/ry wouldn't completely solve the problem ... when w or h also get large, simply squaring them would overflow as well.  as would large values of mx/w.

in looking at the math, i feel like there's a way we can do it w/out mul/div, but maybe it'd be way more complicated than simply using 64bit math.

pierrejoye:
We should used standard integer types like int64_t here if necessary. However we do not yet cover the cases where the platform does not support them. Easy on Windows but slightly trickier on linux/unix with both cmake and autoconf.

pierrejoye:
Also for this specific bug, 64biit integer is enough for this function. Using long long will make use of 128bit integer on 64bit Linux, which may cause other issues.

vapier:
autoconf provides `AC_TYPE_INT64_T` and `AC_TYPE_UINT64_T` to quickly detect standard types.  i'm not sure we should bother supporting platforms where 64-bit types aren't available anymore.

i'm not sure what you mean by 128-bit ... `long long` is going to be 64-bit.

pierrejoye:
Yes the detection is easy.  But we have to define them if the miss.

if we use long long on 64bit as well it will be 128bits.

[...]

PR #451, opened 2018-08-20
Remove trailing comma in enum list

petk:
Hello,

Trailing commas in enums with the C99 are legit unless agreed otherwise with the coding style used in the given codebase.

The LLVM coding style guide doesn't use them. Neither does the default clang format tool enforce any particular trailing comma style. The Google C++ uses trailing commas. Also the git differences can be a bit nicer.

Since current code base of the libgd doesn't use trailing commas in enums, this patch syncs this to avoid possible compiler warning: comma at end of enumerator list [-Wpedantic] for C89...

Thank you.

### DIFF
--- a/src/gd_intern.h
+++ b/src/gd_intern.h
@@ -44,7 +44,7 @@
 
 typedef enum {
     HORIZONTAL,
-    VERTICAL,
+    VERTICAL
 } gdAxis;
 
 /* Convert a double to an unsigned char, rounding to the nearest


vapier:
taking a step back, should we formalize and drop support for C89 (i.e. require C99+) ?  realistically no one is testing C89 anymore, and things like `-pedantic` and `-ansi` are dumb nowadays.

i think a similar discussion came up before wrt standards, but i can't find them atm.  i vaguely recall @pierrejoye listing some embedded setups he thought we should keep working.

cmb69:
@vapier Maybe you are referring to https://github.com/libgd/libgd/issues/169#issuecomment-225336212? This was particularly about uint64_t and friends, though.

I'm not sure whether we really should require C99 now/soon, but I think we can stick with the trailing commas in enums anyway.

vapier:
thanks, that was the stream i was looking for.  i opened #454 to track C99.

petk:
Hello, thanks for waiting so long... It seems that this is looking good and logical in the linked discussion above so I think the pick going currently in the direction of C99 is the way to go. It's been several decades since C89 so it shouldn't cause too many issues I think. Cheers.

PR #456, opened 2018-08-30
Add LGTM code quality badges

xcorail:
Hi there!

I thought you might be interested in adding these [code quality badges](https://lgtm.com/projects/g/libgd/libgd/ci/#badges) to your project. They will indicate a very high code quality to potential users and contributors.
You can also check the alerts discovered by [LGTM](https://lgtm.com/projects/g/libgd/libgd).

N.B.: I am on the team behind LGTM.com, I'd appreciate your feedback on this initiative, whether you're interested or not, if you find time to drop me a line. Thanks.

### DIFF
--- a/README.md
+++ b/README.md
@@ -3,6 +3,8 @@
 [![Build Status](https://travis-ci.org/libgd/libgd.svg?branch=master)](https://travis-ci.org/libgd/libgd)
 [![Build Status](https://scan.coverity.com/projects/3810/badge.svg)](https://scan.coverity.com/projects/libgd)
 [![Chat](https://badges.gitter.im/libgd/libgd.svg)](https://gitter.im/libgd/libgd)
+[![Code Quality: Cpp](https://img.shields.io/lgtm/grade/cpp/g/libgd/libgd.svg?logo=lgtm&logoWidth=18)](https://lgtm.com/projects/g/libgd/libgd/context:cpp)
+[![Total Alerts](https://img.shields.io/lgtm/alerts/g/libgd/libgd.svg?logo=lgtm&logoWidth=18)](https://lgtm.com/projects/g/libgd/libgd/alerts)
 
 GD is an open source code library for the dynamic creation of images by
 programmers.


vapier:
i'm not seeing the value in the 'Code Quality' badge.  can you explain what it's trying to show ?

the 'Total Alerts' site seems to have a number of false positives and bad suggestions.
- it doesn't understand the C++ `const` keyword on functions
- it thinks casting an `int` to `size_t` "fixes" multiplication overflow -- on 32bit systems, sizeof(int) == sizeof(size_t), so 0x7fffffff * 0x7fffffff won't fit in a size_t
- it also thinks you can cast an `int` to `unsigned long` to avoid multiplication overflow
- it can't handle while loops correctly -- it only evaluates the condition once at the start and doesn't handle reloading of vars (see the gdkanji.c report)
- the multiple header inclusion checks miss the ifdef protection we already have (although we could tweak them to make the warning not show up)

it did flag some redundant color negative checks i'll clean up, but any decent compiler would have also optimized those away

so i'm not inclined to link to 'Total Alerts' if it's going to encourage people to send us commits with bad/broken changes.  maybe if if the static analyzer improved in quality we could re-evaluate.

xcorail:
Hello @vapier 

First thanks for taking the time for this detailed feedback! I really appreciate it.

> i'm not seeing the value in the 'Code Quality' badge. can you explain what it's trying to show ?

LGTM analyses the tens of thousands of other projects on GitHub, and compare your project with similar ones (language, size ...), to compute these badges. So these badges show that your project's code is of very high quality. I think it's a valuable information for potential users and contributors, it tells them that you care about quality, that this project will easily accept new features, ...
If you have some time, you can read this [blog post](https://lgtm.com/blog/code_quality_1) which shows an amazing correlation between our quality badges and the project popularity.  

> the 'Total Alerts' site seems to have a number of false positives and bad suggestions.

I'll dig into these with our C++ team ASAP.

> it can't handle while loops correctly -- it only evaluates the condition once at the start and doesn't handle reloading of vars (see the gdkanji.c report)

When I look at your code: 
`while (c >= 64 && c != '\0' && whatcode == EUCORSJIS)`
and the alert says:
`Comparison is always true because c >= 64.`
When you mouseover the alert (the giant red box), it highlights the specific comparison that is always true: `c != '\0'`
Whatever happens to c within the while loop, if c >= 64, then c cannot be '\0'. This is what this alert is telling you.

> it did flag some redundant color negative checks i'll clean up

Happy it has been useful for this

> maybe if if the static analyzer improved in quality we could re-evaluate

Our technology allows us to fix false positives very very quickly, and [our analyses are even open source](https://github.com/Semmle/ql) so others can contribute with fixes, so after I review your report, it will be a matter of days to provide a fix if these happens to be false positives.

Thanks again, really appreciated!

jbj:
Hi @vapier. I work on Semmle's C++ analysis, and I'll try to shed some light on what the alerts are saying.

1. As @xcorail writes, `c >= 64 && c != '\0'` is equivalent to `c >= 64`.
2. The const alert is for the function signature `const gdImagePtr GetPtr() const`, where `gdImagePtr` is defined as `typedef gdImage *gdImagePtr;`. The `const` after the function name is fine, but the initial `const` probably doesn't do what a reader might expect. It makes the returned _pointer_ const but not the pointed-to value. In other words, the signature is equivalent to `gdImage *const GetPtr() const`, which is effectively the same as `gdImage *GetPtr() const`. The intended signature was most likely `const gdImage *GetPtr() const` (or, equivalently, `gdImage const *GetPtr() const`).
3. It's true that [the overflow advice](https://lgtm.com/rules/2157860313/) does not prevent overflow on 32-bit systems. Using your knowledge of the particular values being multiplied you may be able to dismiss overflow as impossible (maybe such an image couldn't fit on a 32-bit system in the first place) or as affecting more code than what LGTM highlights. The advice offered by LGTM here is a generic and local fix to the often-overlooked problem that `(long)(x * y)`  may overflow in many cases where `(long)x * y` would not.
4. It's true that our header guard checks could be more lenient, but we've chosen to require header guards very early in the file. This ensures that future maintainers don't accidentally add code before the header guard because it's fallen too far down over time.

vapier:
> you can read this blog post

that post is pretty extensive and the approach sounds reasonable.  i'm OK with adding this badge.

> Whatever happens to c within the while loop, if c >= 64, then c cannot be '\0'. This is what this alert is telling you.

thanks, that should be fixed now

> The const alert is for the function signature const gdImagePtr GetPtr() const.

i'll take another look at this.  mousing over the box highlights the "GetPtr" function itself when it really seems like it should be highlighting the first const.

> It's true that the overflow advice does not prevent overflow on 32-bit systems.

while i certainly agree that figuring out multiplication overflows is not easy and is project-specific (e.g. our gd_nnquant.c code already does an overflow check before the allocation code in question), my point is that the generic advice "cast it to long" is bad for the community.  it misleads people into thinking their code is safe when it absolutely is not.  if you want to reword it to include a suggestion like "use larger types when multiplying", then go for it, but you cannot assume sizeof(long)>sizeof(int) or sizeof(size_t)>sizeof(int).  32-bit is still very much a common architecture with armv7 (and x86, although that's largely been replaced by x86_64 now).

PR #457, opened 2018-09-04
Remove HAVE_ERRNO_H

petk:
The `<errno.h>` header file is part of the standard C89 headers [1] and
on older systems there needed to be also a manual check if header is
present.

Since libgd requires at least C89 or greater, the `HAVE_ERRNO_H`
symbol defined by Autoconf in configure.ac [2], and Cmake build system
files can be removed and simplifed.

Refs:
[1] https://port70.net/~nsz/c/c89/c89-draft.html#4.1.3
[2] https://git.savannah.gnu.org/cgit/autoconf.git/tree/lib/autoconf/headers.m4


vapier:
this sounds good.  unfortunately, seeing your change made me notice that README.JPN had non-UTF-8 encoding.  so i fixed that, but now your change conflicts :).  can you rebase on top of that so we can merge it ?

petk:
Rebased...

vapier:
thanks!

PR #459, opened 2018-09-05
Remove HAVE_LIMITS_H

petk:
The `<limits.h>` header file is part of the standard C89 headers [1] and
on current systems there is no need to manually check if header is
present anymore.

Since the libgd requires at least C89 or greater, the `HAVE_LIMITS_H`
symbol defined by Autoconf in configure.ac [2], and Cmake build system
files can be removed and simplifed.

Refs:
[1] https://port70.net/~nsz/c/c89/c89-draft.html#4.1.2
[2] https://git.savannah.gnu.org/cgit/autoconf.git/tree/lib/autoconf/headers.m4


vapier:
thanks!

PR #461, opened 2018-09-05
Upgrade AX_CFLAGS_WARN_ALL macro file

petk:
This patch bumps current file version from the Autoconf Archive from
15 to 16:
http://git.savannah.gnu.org/cgit/autoconf-archive.git/tree/m4/ax_cflags_warn_all.m4

Changes:
- https links in the file comments


vapier:
hmm, i think i made this http->https fix in autoconf-archive a while back :p

PR #481, opened 2018-11-12
Fix 479: Try to make gdMalloc() return 16-byte aligned memory

kbrow1i:
This is necessary if we're using libimagequant.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -95,6 +95,9 @@ AC_CHECK_HEADERS(iconv.h,
 dnl do we need to specify -lm explicitly?
 AC_CHECK_FUNC(sin,,[AC_CHECK_LIB(m,sin)])
 
+dnl No need to check for posix_memalign if aligned_alloc works.
+AC_CHECK_FUNCS([aligned_alloc posix_memalign], [break])
+
 AX_PTHREAD()
 AX_OPENMP()
 
--- a/src/config.h.cmake
+++ b/src/config.h.cmake
@@ -6,6 +6,9 @@
 /* Whether to support gd image formats */
 #define ENABLE_GD_FORMATS @ENABLE_GD_FORMATS@
 
+/* Define to 1 if you have the `aligned_alloc' function. */
+#cmakedefine HAVE_ALIGNED_ALLOC
+
 /* Define to 1 if you have the <dirent.h> header file. */
 #cmakedefine HAVE_DIRENT_H
 
@@ -66,6 +69,9 @@
 /* Define if OpenMP is enabled */
 #cmakedefine HAVE_OPENMP
 
+/* Define to 1 if you have the `posix_memalign' function. */
+#cmakedefine HAVE_POSIX_MEMALIGN
+
 /* Define if you have POSIX threads libraries and header files. */
 #cmakedefine HAVE_PTHREAD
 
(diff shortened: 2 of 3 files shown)


vapier:
there's no need to check for `posix_memalign` imo.  we can assume it exists and wait for someone to complain.  it's been part of POSIX for over a decade.

if the only thing that cares about 16byte alignment is the liq code, then imo we should keep it scoped to that.  we haven't guaranteed larger alignment for gdMalloc, and people are allowed to override gdMalloc themselves (in which case the liq code would still break).

kbrow1i:
> there's no need to check for `posix_memalign` imo. we can assume it exists and wait for someone to complain. it's been part of POSIX for over a decade.

So should I also drop the check for aligned_alloc and just use posix_memalign?

> if the only thing that cares about 16byte alignment is the liq code, then imo we should keep it scoped to that. we haven't guaranteed larger alignment for gdMalloc, and people are allowed to override gdMalloc themselves (in which case the liq code would still break).

I thought my change to gdMalloc was already scoped to liq.  Or am I misunderstanding you?

cmb69:
> I thought my change to gdMalloc was already scoped to liq. Or am I misunderstanding you?

Something like the following should be sufficient (untested):
````diff
 src/gd_topal.c | 11 ++++++++++-
 1 file changed, 10 insertions(+), 1 deletion(-)

diff --git a/src/gd_topal.c b/src/gd_topal.c
index 5178277..29a6694 100644
--- a/src/gd_topal.c
+++ b/src/gd_topal.c
@@ -1523,6 +1523,15 @@ static void free_truecolor_image_data(gdImagePtr oim)
 	oim->tpixels = 0;
 }
 
+#ifdef HAVE_LIBIMAGEQUANT
+/* liq requires 16 byte aligned heap memory */
+static malloc16(size_t size)
+{
+	void *p;
+	return posix_memalign(&p, 16, size) == 0 ? p : 0;
+}
+#endif
+
 /*
  * Module initialization routine for 2-pass color quantization.
  */
@@ -1601,7 +1610,7 @@ static int gdImageTrueColorToPaletteBody (gdImagePtr oim, int dither, int colors
 #ifdef HAVE_LIBIMAGEQUANT
 	if (oim->paletteQuantizationMethod == GD_QUANT_DEFAULT ||
 	        oim->paletteQuantizationMethod == GD_QUANT_LIQ) {
-		liq_attr *attr = liq_attr_create_with_allocator(gdMalloc, gdFree);
+		liq_attr *attr = liq_attr_create_with_allocator(malloc16, free);
 		liq_image *image;
 		liq_result *remap;
 		int remapped_ok = 0;
````
And we probably should check if `attr` is `NULL` and bail out gracefully in that case.

kbrow1i:
OK, now I understand.  I'll modify my branch accordingly.

I don't have much experience with PRs, so I'm not sure how to best make the change.  Is it OK if I just change my commit locally and then do a forced push?  Or is it better if I make a second commit?

kbrow1i:
> Guess a forced push would be okay. Or maybe even a new PR?

OK, I'll close this PR and make a new one, which will contain essentially your proposed code.  I looked into bailing out if attr is NULL, but I don't think it's necessary.  It seems that the liq functions that are called always check for a NULL attr arg and handle this gracefully.

vapier:
doing a force push or updated your branch with more commits is fine.  GH offers a rebase/squash option so the GH repo would only merge a single change.

PR #482, opened 2018-11-14
Fix #479: Provide a suitable malloc function to liq

### LINKED ISSUE #479: gdMalloc alignment problem on 32-bit Cygwin
The program php_bug_67325 in the libgd test suite crashes on 32-bit Cygwin when libgd is built with libimagequant support.  I traced this to an alignment problem of liq_image objects, which ultimately comes down to the fact that gdMalloc() doesn't necessarily return 16-byte aligned memory.

In more detail, there is a call to liq_attr_create_with_allocator(gdMalloc, gdFree) in gdImageTrueColorToPaletteBody().  But the documentation of liq_attr_create_with_allocator() says, "The malloc function must return 16-byte aligned memory on x86 (and on other architectures memory aligned for double and pointers)."  See https://pngquant.org/lib/#liq_attr_create_with_allocator.

Currently gdMalloc() just calls the system malloc().  Should it call something like posix_memalign() instead?

kbrow1i:
liq requires 16 byte alignment, which gdMalloc() doesn't guarantee.
Define and use a new malloc16() function when building with liq.

This replaces PR#481.

### DIFF
--- a/src/gd_topal.c
+++ b/src/gd_topal.c
@@ -1523,6 +1523,15 @@ static void free_truecolor_image_data(gdImagePtr oim)
 	oim->tpixels = 0;
 }
 
+#ifdef HAVE_LIBIMAGEQUANT
+/* liq requires 16 byte aligned heap memory */
+static void *malloc16(size_t size)
+{
+	void *p;
+	return posix_memalign(&p, 16, size) == 0 ? p : NULL;
+}
+#endif
+
 /*
  * Module initialization routine for 2-pass color quantization.
  */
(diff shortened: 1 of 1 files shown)


vapier:
at this point we've implemented `liq_attr_create`, so seems like we should replace this line with that call and be done (and put a comment explaining why we don't use `gdMalloc`).

cmb69:
@vapier From the [`liq_attr_create_with_allocator` docs](https://pngquant.org/lib/#liq_attr_create_with_allocator):
> The malloc function must return 16-byte aligned memory on x86 (and on other architectures memory aligned for double and pointers). Conversely, if your stdlib's malloc doesn't return appropriately aligned memory, you should use this function to provide aligned replacements.

vapier:
gotcha, thanks

PR #483, opened 2018-11-23
Compute average in gdGuessBackgroundColorFromCorners properly

omron93:
Coverity scanning finds following error in libgd:
```
Error: UNINTENDED_INTEGER_DIVISION:
libgd-2.2.5/src/gd_crop.c:302: integer_division: Dividing integer expressions "(im->trueColor ? (tl & 0xff0000) >> 16 : im->red[tl]) + (im->trueColor ? (tr & 0xff0000) >> 16 : im->red[tr]) + (im->trueColor ? (bl & 0xff0000) >> 16 : im->red[bl]) + (im->trueColor ? (br & 0xff0000) >> 16 : im->red[br])" and "4", and then converting the integer quotient to type "float". Any remainder, or fractional part of the quotient, is ignored.
libgd-2.2.5/src/gd_crop.c:302: rounding_remediation: For the rounding operation to work as intended, change or cast either division operand to type "float".
#  300|   		register int r,b,g,a;
#  301|   
#  302|-> 		r = (int)(0.5f + (gdImageRed(im, tl) + gdImageRed(im, tr) + gdImageRed(im, bl) + gdImageRed(im, br)) / 4);
#  303|   		g = (int)(0.5f + (gdImageGreen(im, tl) + gdImageGreen(im, tr) + gdImageGreen(im, bl) + gdImageGreen(im, br)) / 4);
#  304|   		b = (int)(0.5f + (gdImageBlue(im, tl) + gdImageBlue(im, tr) + gdImageBlue(im, bl) + gdImageBlue(im, br)) / 4);
```

This PR aims to fix this and compute the color average properly.

### DIFF
--- a/src/gd_crop.c
+++ b/src/gd_crop.c
@@ -303,10 +303,10 @@ static int gdGuessBackgroundColorFromCorners(gdImagePtr im, int *color)
 	} else {
 		register int r,b,g,a;
 
-		r = (int)(0.5f + (gdImageRed(im, tl) + gdImageRed(im, tr) + gdImageRed(im, bl) + gdImageRed(im, br)) / 4);
-		g = (int)(0.5f + (gdImageGreen(im, tl) + gdImageGreen(im, tr) + gdImageGreen(im, bl) + gdImageGreen(im, br)) / 4);
-		b = (int)(0.5f + (gdImageBlue(im, tl) + gdImageBlue(im, tr) + gdImageBlue(im, bl) + gdImageBlue(im, br)) / 4);
-		a = (int)(0.5f + (gdImageAlpha(im, tl) + gdImageAlpha(im, tr) + gdImageAlpha(im, bl) + gdImageAlpha(im, br)) / 4);
+		r = (2 + gdImageRed(im, tl) + gdImageRed(im, tr) + gdImageRed(im, bl) + gdImageRed(im, br)) / 4;
+		g = (2 + gdImageGreen(im, tl) + gdImageGreen(im, tr) + gdImageGreen(im, bl) + gdImageGreen(im, br)) / 4;
+		b = (2 + gdImageBlue(im, tl) + gdImageBlue(im, tr) + gdImageBlue(im, bl) + gdImageBlue(im, br)) / 4;
+		a = (2 + gdImageAlpha(im, tl) + gdImageAlpha(im, tr) + gdImageAlpha(im, bl) + gdImageAlpha(im, br)) / 4;
 		*color = gdImageColorClosestAlpha(im, r, g, b, a);
 		return 0;
 	}


cmb69:
Thanks!  The old code is obviously wrong. But what about something like:
````.c
r = (2 + gdImageRed(im, tl) + gdImageRed(im, tr) + gdImageRed(im, bl) + gdImageRed(im, br)) / 4;
````

omron93:
Agree. Omitting casting and floating point operations is better.

> PS: PHP's bundled libgd has the same issue.

Does this mean that this PR has to be handled somehow differently?

cmb69:
> Or I should do something?

No. It's more of a reminder for me to port an eventual change to PHP's bundled libgd.

Anyhow, thanks! The patch looks good to me, but it would be sensible to have a (regression) test (see tests/ directory). If you could provide one, great! Otherwise I'll have a look at it.

omron93:
> but it would be sensible to have a (regression) test (see tests/ directory). If you could provide one, great! Otherwise I'll have a look at it.

I'm not a gd regular user and I guess that such test requires having an example image of previous result and the correct one. And I have no idea how to get them, so I would prefer you to create test. Sorry.

cmb69:
I've tried to come up with a test, but the code which does not properly round is only ever supposed to be run, if all four corners of the image have different colors or palette indexes. However, if all four corners of the image have different colors or palette indexes, nothing will ever be cropped.

There is a loophole, though, namely that `gdGuessBackgroundColorFromCorners()` does not check whether `tr` and `br` are equal, which could be used to provide a test script to show the erroneous rounding behavior. However, this loophole should be closed instead (since it's a bug), and instead of calculating an average color value, `gdImageCropAuto()` should bail out, if `gdGuessBackgroundColorFromCorners()` returns `0`. That is, unless I'm missing something, though.

cmb69:
>There is a loophole, though, namely that gdGuessBackgroundColorFromCorners() does not check whether tr and br are equal, […]

This is fixed with 6613094e5d218dc4d4372757aef5e58c6462a9f7.

willson-chen:
It seems that the issue has been fixed. Shall we close this PR?

cmb69:
> It seems that the issue has been fixed. Shall we close this PR?

The actual issue has not been fixed. We still divide two integer operands, add `0.5` to the result, and truncate to integer; the last two steps don't have the desired effect (actually, they have no effect at all).

willson-chen:
In fact, I can not full understand the meaning of this code block. :(

We want the average by dividing `4`? and why plus `0.5f`? What is the difference between `plus 0.5f then dividing 4` and `dividing 4 then plus 2`?

cmb69:
A simple way to *round* a number, is to add 0.5 and then to truncate to int. This is almost certainly the intention of this code, but it doesn't properly do it. So this PR suggests to add 2 *before* dividing by 4, which looks reasonable.

willson-chen:
I see. Many thanks. @cmb69 

So this PR is fine, but need a test case, right?

cmb69:
Having a regression test is always desireable, but in this case it may not be possible to have one; see https://github.com/libgd/libgd/pull/483#issuecomment-441384845, first paragraph (the second paragraph is obsolete, see https://github.com/libgd/libgd/pull/483#issuecomment-441455660). I think we can merge this PR as is.

willson-chen:
@omron93 bud thanks

Issue #450, opened 2018-08-01
Failed to open 1 bit per pixel bitmap


quarthex:
When loading a monochrome image in Windows bitmap format with 1-bit per pixel encoding, the operation fails.
The bug occurs for some specific image widths like, for instance, 10px.

[example.bmp.gz](https://github.com/libgd/libgd/files/2249634/example.bmp.gz)

PR #487, opened 2018-12-30
Fix loading of 1bpp images

quarthex:
Should fix issue #450.


cmb69:
Thank you very much! Applied as d0859134fcbd8817f87895bd1cb158096568a7a2.

Issue #345, opened 2016-11-15
gdlib-config deprecated so what to use now ?


centminmod:
with gdlib-config deprecated what do i use now to get equivalent output for ? particular the compiled and enabled libs etc

```
/usr/bin/gdlib-config --all
GD library  2.0.34
includedir: /usr/include
cflags:     -I/usr/include
ldflags:    -L/usr/lib64 -Wl,-z,relro
libs:       -lXpm -lX11 -ljpeg -lfontconfig -lfreetype -lpng -lz -lm  
libdir:     /usr/lib64
features:   GD_XPM GD_JPEG GD_FONTCONFIG GD_FREETYPE GD_PNG GD_GIF GD_GIFANIM GD_OPENPOLYGON
```

vapier:
as the commit message said, use pkg-config to query gd settings (like libraries & include paths).  if you don't know how to use pkg-config, consult their homepage for docs:
https://www.freedesktop.org/wiki/Software/pkg-config/

rurban:
This is bogus. The .pc doesn't give us the features, there's even no Defines line with the -DHAVE_GIF ...
This is a severe regression. We need the features list or list of DEFINE's

cmb69:
@vapier Wouldn't it be sensible to add a `feature` variable to gdlib.pc which lists the features like gdlib-config did? This would make it trivial to switch from gdlib-config to pkg-config.

cmb69:
PHP's GD binding offers “introspection” about which features are supported. This is basically been done by detecting supported features with `AC_TRY_LINK`. This won't work for the GD and GD2 formats, though. One likely would have to actually run a test program and check for the error message.

cmb69:
PHP can be built against its bundled libgd or against a system libgd. I believe most (if not all) Linux distros choose the latter. And while I see your point that feature detection should not be done during build-time, it is still necessary with regard to libgd – if libgd doesn't support JPEG, there is no `gdImageCreateFromJpeg()`, so a client must not use it.

vapier:
> if libgd doesn't support JPEG, there is no `gdImageCreateFromJpeg()`

how about we change that ?  other features (like zlib), we always provide the funcs, just let them return errors.

vapier:
actually, i see we already have `gdSupportsFileType` for people to figure out at runtime whether a format is supported.  which means we just have to add the missing gaps for ABI compat.

dimpase:
[...]

PR #491, opened 2019-01-15
add stub ABI funcs when formats are disabled #345

vapier:
We do this for some funcs already, so add jpeg/png/tiff/liq/webp
stubs too.  This keeps our ABI stable and lets code do runtime
detection instead.

### DIFF
--- a/src/gd_jpeg.c
+++ b/src/gd_jpeg.c
@@ -1192,4 +1192,63 @@ void jpeg_gdIOCtx_dest(j_compress_ptr cinfo, gdIOCtx *outfile)
 	dest->outfile = outfile;
 }
 
+#else /* !HAVE_LIBJPEG */
+
+static void _noJpegError(void)
+{
+	gd_error("JPEG image support has been disabled\n");
+}
+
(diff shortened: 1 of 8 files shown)


src/gd_topal.c
```diff
@@ -1800,4 +1800,34 @@ static int gdImageTrueColorToPaletteBody (gdImagePtr oim, int dither, int colors
 	return conversionSucceeded;
 }
 
+#else
```
cmb69: This `#else` appears to match `#ifdef QUANT_2PASS_SUPPORTED`, so this is not related to liq. Actually, liq is optional, see [gdPaletteQuantizationMethod](https://libgd.github.io/manuals/2.2.5/files/gd-h.html#gdPaletteQuantizationMethod). It seems to me we should rather remove the `#ifdef QUANT_2PASS_SUPPORTED`, since we have `#define QUANT_2PASS_SUPPORTED` above, so one would have to modify the file to influence it.

vapier: thanks.  i had checked the diff files to make sure they were being compiled, and this one i couldn't get to trigger.  i just assumed i wasn't using the right configure flags to disable the features.

PR #493, opened 2019-01-17
Fix possible double free bugs in gd_jpeg.c and gd_wbmp.c

y3noor:


cmb69:
Thanks. Applied as 553702980ae89c83f2d6e254d62cf82e204956d0.

PR #494, opened 2019-01-29
Allow GIFs missing both Global and Local Colormaps to still decode

az6667:
Ensure that a GIF without any Global or Local color tables is still
decoded by libgd.

GIF89a spec indicates conforming image files need not have
Global or Local color tables at all.

Spec recommends creating custom color map in that situation, and
that at least Black+White as first two entries, to ensure B&W images
are decoded.

Some commonly used single-pixel GIFs found around the web are
undecoded by libgd otherwise. Test case has been included.

References:
https://www.w3.org/Graphics/GIF/spec-gif89a.txt
http://probablyprogramming.com/2009/03/15/the-tiniest-gif-ever


tests/gif/CMakeLists.txt
```diff
@@ -3,6 +3,7 @@ LIST(APPEND TESTS_FILES
 	bug00181
 	bug00227
 	gif_null
+	git_nocolormaps
```
cmb69: Typo: should be `gif_nocolormaps`.

az6667: well that's embarrasing. thankyou.

src/gd_gif_in.c
```diff
@@ -315,8 +315,9 @@ BGD_DECLARE(gdImagePtr) gdImageCreateFromGifCtx(gdIOCtxPtr fd)
 			ReadImage(im, fd, width, height, localColorMap, BitSet(buf[8], INTERLACE), &ZeroDataBlock);
 		} else {
 			if(!haveGlobalColormap) {
-				gdImageDestroy(im);
-				return 0;
+				ColorMap[CM_RED][1] = 0xff;	//Still a valid gif, apply simple default palette as per spec
```
vapier: can you move the comment above the line instead of being inline ?

also put a space after the `//`

az6667: yup sure

vapier:
thanks!

Issue #445, opened 2018-06-29
FindFreetype.cmake (missing:  FREETYPE_INCLUDE_DIRS)


wahibre:
cmake can't find FreeType on Fedora and CentOS.

```
[root@36cdc16e51ef libgd]# cmake .
-- gd version 2.3.0-dev
-- gd shared lib version 3 (3.0.8)
-- Could NOT find Freetype (missing:  FREETYPE_INCLUDE_DIRS)
-- Configuring done
-- Generating done
-- Build files have been written to: /root/libgd
```

actual path to ftheader.h:
```
/usr/include/freetype2/freetype/config/ftheader.h
```

branch: master
https://github.com/libgd/libgd/blob/08ae745426579be883d7f8aeabcc592b24de0597/cmake/modules/FindFreetype.cmake#L75-L79

vapier:
we should just gut these cmake modules and replace them with pkg-config lookups

harkue:
Fixed, thanks!

wahibre:
Looks OK, thank You.

willson-chen:
> we should just gut these cmake modules and replace them with pkg-config lookups

@vapier I have seen a PR in master branch(https://github.com/libgd/libgd/commit/e0cb1b76c305db68b251fe782faa12da5d357593#diff-eab09c1a70a2c5671ce56ca0f82b0467) ,and it seems solved this issue . But  I am little confuse if it's solved by pkg-config lookups?

vapier:
i don't remember anymore.  that commit was just updating us to a newer cmake modules release.

PR #496, opened 2019-02-09
Use pkg-config path hints in "FindFreeType.cmake" #445

az6667:
In some distros (Fedora/Centos), finding "freetype" would fail.
This asks pkg-config for some help. Recycled the pkg-config
example that was in "FindFontConfig.h"


az6667:
If I'm not doing these pull requests how you'd prefer, please let me know.

vapier:
we're taking FindFreeType.cmake from the cmake project.  can you send your fixes to them ?

az6667:
absolutely. I'll close this pull request.

az6667:
I assumed you meant "please commit to upstream instead"?
It looks like it was fixed in cmake at least as far back as v2.8.12.2

vapier:
i mean we're OK with pulling changes from the cmake project, but we don't want to diverge from them

PR #505, opened 2019-07-31
Fix typo

lance0z:
Make the message start with an upper case letter.


lance0z:
The commit message has been changed.

vapier:
thanks!

Issue #501, opened 2019-05-31
Uninitialized read in gdImageCreateFromXbm (CVE-2019-11038)


carnil:
There is a GD related issue fixed in PHP (7.1.30, 7.2.19 and 7.3.6) which was assigned CVE-2019-11038 and is in the PHP bug https://bugs.php.net/bug.php?id=77973 .

Filling the issue for the correlated issue in the libgd library itself.

mejo-:
While working on a security update of libgd for Debian, I prepared a patch to fix this CVE based on the [PHP gd patch](https://git.php.net/?p=php-src.git;a=commit;h=ed6dee9a198c904ad5e03113e58a2d2c200f5184). You can find it in PR #503.

Here's a simple reproducer:

1. Create a xbm file that triggers the bug:
    ```sh
    printf "23646566696e6520776964746820320a23646566696e652068656967687420320a737461746963206368617220626974735b5d203d7b0a7a7a787a7a" | xxd -r -p - github_bug_501.xbm
   ```

2. A simple C reproducer:
   ```c
   #include "gd.h"
   #include <stdio.h>
   
   int main() {
       gdImagePtr im;
       FILE *xbm_in;
   
       xbm_in = fopen("github_bug_501.xbm", "rb");
       im = gdImageCreateFromXbm(xbm_in);
       fclose(xbm_in);
   
       gdImageDestroy(im);
   }
   ```

Unfortunately I failed to create a test for this bug (under `tests/xbm`). If I understand it correctly, we only could test for the error message after `gdImageCreateFromXbm`, but I don't know how to do that. At least, `gdTest` doesn't provide functions for that. A simple `im == NULL` assertion won't help as that would be true both with and without this fix.

pgajdos:
Note that the test case will segfault in gdImageDestroy() even AFTER, as gdImageCreateFromXbm() returns 0 in case of failure.

wang0z:
PR https://github.com/libgd/libgd/pull/506 is the test case for the CVE.  The CI show it faied because the master has not merge the fix. The case itself is no problem.

PR #506, opened 2019-08-15
Add a test for issue #501 CVE-2019-11038

wang0z:
Hi, 

I added a test case for CVE-2019-11038.


wang0z:
The CI said it failed. Yes of course it should fail. Because the master has not merge the fix, so the case failed. That did NOT mean the patch of case is no right.

wang0z:
Replaced by #507.

PR #509, opened 2019-09-10
Update source download url and bug report url in README.TESTING

willson-chen:
It seems that the old urls can't not be found.

### DIFF
--- a/docs/README.TESTING
+++ b/docs/README.TESTING
@@ -40,11 +40,11 @@ If you use a packaged version of the GD Library (installed from RPM, deb or
 ports, gentoo packages or any other packages format or distribution), please try
 to run the tests using the source releases available at:
 
-http://www.libgd.org/Downloads
+http://github.com/libgd/libgd/release
 
 If the tests fail using our source release, please report a bug here:
 
-http://bugs.libgd.org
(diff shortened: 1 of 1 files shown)


docs/README.TESTING
```diff
@@ -40,11 +40,11 @@ If you use a packaged version of the GD Library (installed from RPM, deb or
 ports, gentoo packages or any other packages format or distribution), please try
 to run the tests using the source releases available at:
 
-http://www.libgd.org/Downloads
+http://github.com/libgd/libgd/release
```
vapier: this doesn't work either :).  it's "releases", not "release".

go with https://libgd.org/pages/downloads.html for now instead though

docs/README.TESTING
```diff
@@ -40,11 +40,11 @@ If you use a packaged version of the GD Library (installed from RPM, deb or
...
 
 If the tests fail using our source release, please report a bug here:
 
-http://bugs.libgd.org
+http://github.com/libgd/libgd/issues
```
vapier: https here

willson-chen:
@vapier, thanks :)
I pushed another commit to correct.

vapier:
thx bud

PR #510, opened 2019-09-20
fixed cmake failed with generators "Visual Studio XX" under windows

willson-chen:
unistd.h (included in src/getopt.c) can't not be found definitely in win. And it is also unnessary for building libgd both by cmake and automake in Linux.


willson-chen:
Included in #512 . Close

PR #512, opened 2019-09-26
fixed optxxx var undefined in src/webpng.c under MSVC

willson-chen:
`unistd.h` can not included by src/webpng.c under MSVC. So optind opterr and related var are unavaible.

### DIFF
--- a/src/getopt.c
+++ b/src/getopt.c
@@ -33,7 +33,6 @@
 #include <stdio.h>
 #include <stdlib.h>
 #include <string.h>
-#include <unistd.h>
 
 int	opterr = 1,		/* if error message should be printed */
 	optind = 1,		/* index into parent argv vector */
--- a/src/getopt.h
+++ b/src/getopt.h
@@ -1,0 +1,20 @@
+#ifndef GETOPT_H
+#define GETOPT_H 1
+
+#ifdef __cplusplus
+extern "C" {
+#endif
+
+extern int opterr;
+extern int optind;
+extern int optopt;
+extern int optreset;
+extern char *optarg;
+
+int getopt(int nargc, char * const nargv[], const char *ostr);
+
+#ifdef __cplusplus
+}
+#endif
+
+#endif
--- a/src/webpng.c
+++ b/src/webpng.c
@@ -2,6 +2,10 @@
 #include "config.h"
 #endif
 
+#ifdef WIN32
+#include "getopt.h"
+#endif
+
 /* Bring in standard I/O and string manipulation functions */
 #include <stdarg.h>
 #include <errno.h>


src/getopt.c
```diff
@@ -33,7 +33,6 @@
 #include <stdio.h>
 #include <stdlib.h>
 #include <string.h>
-#include <unistd.h>
```
willson-chen: getopt.c only included for MSVC. So `unistd.h` can not be found definitely.

src/getopt.h
```diff
@@ -0,0 +1,20 @@
...
+
+#ifdef __cplusplus
+extern "C" {
+#endif
+
+extern int opterr;
+extern int optind;
+extern int optopt;
+extern int optreset;
+extern char *optarg;
```
willson-chen: opterr, optind, optopt, optreset, optarg is global variables under Linux by including `unistd.h`. If src/webpng.c or any other source file want to access them, there should be a header file define them as global vars.

willson-chen:
@vapier hi, could you please spare some time to offer some suggestion? I want improve the build under windows platform.

willson-chen:
Hi @vapier @cmb69, could you please spare some time to review this PR.

I have test this PR on win10 with cmake and it works.

But `getopt` is only used in `webpng.c`, which is not the lib source file, but a tool source file. Beside it, there are `gdtopng.c`, `pngtogd.c`, `gdparttopng`, `gd2topng`, `pngtogd2`, `annotate`, `gd2copyal`, `gd2togif`, `giftogd2` in tool source file list. I don't know if someone still use these tools, but I think it is better to move them to a new folder named `tools`. The build failure of these tool source files block adding windows platform to CI.

What do you think? :)

cmb69:
This looks okay to me. I wonder. though, if it makes sense to maintain these tools; they actually look more like examples to me, and their names are somewhat arbitrary.

willson-chen:
> they actually look more like examples to me

Take `gd2topng.c`for example. It generates a executable program, which is used to transform image from gd2 to png formats. It can be treated as a tools, or an example to show the transformation.

I strongly recommand to move them to a new folder, make `src`dir only contain the source files used to generate the library file. What do you think we should deal with them?

src/getopt.h
```diff
@@ -0,0 +1,20 @@
+#ifndef GETOPT_H
+#define GETOPT_H 1
+
+#ifdef __cplusplus
```
stima, line 4: I'm curious if there is necessity in having cpp guard for c library header

src/webpng.c
```diff
@@ -2,6 +2,10 @@
 #include "config.h"
 #endif
 
+#ifdef WIN32
+#include "getopt.h"
```
stima, line 6: Header is good way to declare functions, but I haven't seen headers in this library and this functionality is used only in one place.

PR #514, opened 2019-09-27
implementation of mkstemp in src/webpng.c under win

willson-chen:
use _mktemp_s and _open to implement mkstemp in src/webpng.c under win.

### DIFF
--- a/src/webpng.c
+++ b/src/webpng.c
@@ -11,6 +11,11 @@
 #ifdef HAVE_UNISTD_H
 #include <unistd.h>
 #endif
+#ifdef WIN32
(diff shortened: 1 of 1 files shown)


willson-chen:
@vapier Hi, I had made some modification according to your suggestion. And #512 which is also created by me is similar to this PR. Any advice on both PRs would be appriciated. Thanks

willson-chen:
There will be serveral functions only defined in Unix-like, like `getline`. Put the windows implementation of these funtions into different source files, or create a single file to gather them all, which one is prefered?

Issue #404, opened 2017-08-16
Travis macOS build failure due to missing libjpeg.8.dylib


cmb69:
Since about a week our Travis build fail on macOS/gcc with [this error](https://travis-ci.org/libgd/libgd/jobs/263169074#L4409-L4411):
````
dyld: Library not loaded: /usr/local/opt/jpeg/lib/libjpeg.8.dylib
  Referenced from: /usr/local/opt/libtiff/lib/libtiff.5.dylib
  Reason: image not found
````
[Apparently](https://stackoverflow.com/questions/32703296/dyld-library-not-loaded-usr-local-lib-libjpeg-8-dylib-homebrew-php/45695529#45695529), this is caused by libjepg having been updated to v9, but libtiff still requires v8.

Any ideas about how to solve this best?

vapier:
afaik, homebrew is installed/maintained in the OS image by the travis folks, so if things are broken/out of sync there, it's pretty much out of our hands.  unless we want to download & build all the deps ourselves which i don't think we do.

how about, instead of disabling macOS altogether, we disable tiff on the system until its tiff install is fixed.

cmb69:
> how about, instead of disabling macOS altogether, we disable tiff on the system until its tiff install is fixed.

Sounds good to me. Can you do that?

PR #517, opened 2019-10-08
Fix issue404

willson-chen:
ref issue #404. Enable Travis build on MacOS because it runs successfully now.


vapier:
really wish there was a POSIX regex word boundary, but i couldn't find one.  at least, one worth the effort considering the src/config.hin source is so limited here.

PR #519, opened 2019-10-08
tests/gdTest/CMakeLists.txt: include readdir.c under win

willson-chen:
@pierrejoye has implemented `opendir`, `closedir` and `readdir` in readdir.c, which is only used under windows platform. These functions are enabled by modifiying the tests/gdTest/CMakeLists.txt


willson-chen:
Something must be wrong about my commit. `354014` is also included in this PR. I will create a new one.

PR #520, opened 2019-10-08
tests/gdtest/CMakeLists.txt: include readdir.c under win

willson-chen:
@pierrejoye has implemented `opendir`, `closedir` and `readdir` in readdir.c, which is only used under windows platform. These functions are enabled by modifiying the tests/gdTest/CMakeLists.txt


tests/gdtest/CMakeLists.txt
```diff
@@ -1,3 +1,7 @@
 add_definitions(-DGDTEST_TOP_DIR="${CMAKE_CURRENT_SOURCE_DIR}/..")
-add_library (gdTest STATIC gdtest.c)
+if(WIN32)
+	add_library (gdTest STATIC gdtest.c readdir.c)
```
vapier: can we define a list for this instead so we don't have to duplicate the full thing ?  maybe like (untested):
```
SET(GDTEST_FILES gdtest.c)
if (WIN32)
  LIST(APPEND GDTEST_FILES readdir.c)
endif()
add_library(gdTest STATIC ${GDTEST_FILES})
```

willson-chen: Sounds reasonable. I will append a commit soon. Thanks to your advice.

willson-chen:
@vapier hi, please review my latest commit. Thanks.

PR #522, opened 2019-10-10
windows/Makefile.vc: Solve the exception that nmake compiles and runs the test report getline does not exist

willson-chen:
Solve the exception that nmake compiles and runs the test report getline does not exist.

### DIFF
--- a/src/gd.h
+++ b/src/gd.h
@@ -110,6 +110,10 @@ extern "C" {
 #include <stdarg.h>
 #include "gd_io.h"
 
+#if defined(_WIN32)
+	 BGD_DECLARE(size_t) getline(char** lineptr, size_t* n, FILE* stream);
+#endif
+
 /* The maximum number of palette entries in palette-based images.
    In the wonderful new world of gd 2.0, you can of course have
    many more colors when using truecolor mode. */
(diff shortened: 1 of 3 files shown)


vapier:
BGD_DECLARE & gd.h are for things the library itself exports.  we def don't want to add getline to libgd's API.  if this is only for programs/tests, we should add them to such internal support libs.

it seems like we keep growing more & more logic to replace things that are in POSIX like your #520 PR.  we need to centralize this somewhere as the ad-hoc logic in src/ and tests/ and gdtest.c is getting out of hand.  i'm somewhat inclined to integrate something like gnulib ...

willson-chen:
> BGD_DECLARE & gd.h are for things the library itself exports. we def don't want to add getline to libgd's API. if this is only for programs/tests, we should add them to such internal support libs.

My mistake. I will fix it soon.

> it seems like we keep growing more & more logic to replace things that are in POSIX like your #520 PR. 

Yeah, I tried to improve windows building recently, including cmake, nmake and mingw, which is an important part of milestone 2.2.6.

> we need to centralize this somewhere as the ad-hoc logic in src/ and tests/ and gdtest.c is getting out of hand. i'm somewhat inclined to integrate something like gnulib ...

Sounds good, could you specify how to do?

pierrejoye:
Could you update CMake and configure as well for getline please? Or is it not needed?

pierrejoye:
we will drop makefiles for windows. CMake is the way. It works for visual studio and mingw (see the CI).

PR #529, opened 2019-10-22
CMakeLists.txt: Exclude libm dependency under win

willson-chen:
libm is unavaible under win


vapier:
in autotools, we just link -lm all the time if it's available.  how about we do that here too.

change the two `ADD_GD_TESTS(m)` to `ADD_GD_TESTS()`, and move the win32/m logic into cmake/modules/gd.cmake.

willson-chen:
> in autotools, we just link -lm all the time if it's available. how about we do that here too.
>
>change the two ADD_GD_TESTS(m) to ADD_GD_TESTS(), and move the win32/m logic into cmake/modules/gd.cmake.
Good idear

vapier:
thx bud

willson-chen:
You are the most professinal committer I have ever met. @vapier

PR #527, opened 2019-10-18
Fix #311: fix wrong reference of marco definition

willson-chen:
Fix wrong reference of marco definition

Issue #311, opened 2016-09-25
gdAssertImageFileEqualsMsg() is broken


cmb69:
The macro gdAssertImageFileEqualsMsg() [resolves to gdTestImageCompareFiles()](https://github.com/libgd/libgd/blob/gd-2.2.3/tests/gdtest/gdtest.h#L68), but there is no such function. Obviously, gdTestImageCompareToFile() is meant.

However, after fixing this, the message is still [ignored](https://github.com/libgd/libgd/blob/gd-2.2.3/tests/gdtest/gdtest.c#L513).

I'm not sure how to solve this issue. Should the message be prepended or appended to the message reported by gdTestImageCompareToImage(), or should it replace it?

Thoughts?

willson-chen:
Is the macro gdAssertImageFileEqualsMsg() useful?

vapier:
we've opted to just delete the macro and wait for a user to show up so we can then reason about correct usage :)

PR #530, opened 2019-10-23
Delete unused marcos #311

willson-chen:
To replace PR #527


willson-chen:
@vapier Thanks. And #311 can be closed.

PR #388, opened 2017-04-12
cmake fixes

rkanavath:
when building libgb with libimagequant it complains about gdstatic target does not exists. first commit fixes that.

second commit uses $(MAKE) which can be gmake or make. This is required when building on bsd oses where 'make' is actually 'bsdmake' and not gnu make.

PR #531, opened 2019-10-24
CMakeLists.txt: Fixed cmake failed with ENABLE_LIQ=1

willson-chen:
Target ${GD_LIB_STATIC} does not exist while BUILD_STATIC_LIBS is off. And the space in the beginning of the lines is replace by tabs.


willson-chen:
I found this PR is similar to PR #388. If this PR can be merge, #388 can be closed as well.

vapier:
if we delete the code as discussed in #534, then this block of code would also be deleted

willson-chen:
> if we delete the code as discussed in #534, then this block of code would also be deleted

yeah, let me close this RP.

willson-chen:
This block of code in this PR is unneeded.

PR #532, opened 2019-10-25
CMakeLists.txt: update cpack vendor and version

willson-chen:


vapier:
i hadn't even noticed these

we set the version dynamically earlier in this file -- see the use of `GD(...)` macros.  we should be able to use those variables here instead of hardcoding the version numbers still.

willson-chen:
I have some puzzles about the version number.
1. Why set the version "2.3.0-dev". The next milestone 2.2.6 will be skiped?
2. What is the rule of libgd dynamic library version number? The version is 3.0.8 now, totally different from the src version "2.3.0-dev".
Could you help me to resolve them? @vapier

vapier:
> Why set the version "2.3.0-dev". The next milestone 2.2.6 will be skiped?

the master branch is 2.3.  we have a dedicated GD-2.2 branch for the 2.2.x releases.

> What is the rule of libgd dynamic library version number? The version is 3.0.8 now, totally different from the src version "2.3.0-dev".

SONAME versioning can be hard to explain.  we follow this:
https://www.gnu.org/software/libtool/manual/html_node/Versioning.html

basically the first `3` should never change regardless of versions.  only when ABI breaks, and we don't like breaking the ABI :).

> The macro GV needs perl env, and GV_LT needs shell env. That is not friendly to windows platform. How to get the version number dynamically under windows?

i don't really have an answer here.  we don't want to duplicate version strings in multiple places ... we used to do that, but then we'd forget to update one of the many.

i'd be fine with replacing the getver.pl with a getver.sh, but that would still need a shell.

willson-chen:
> i don't really have an answer here. we don't want to duplicate version strings in multiple places ... we used to do that, but then we'd forget to update one of the many.

`Powershell`, which is supported and in-builted since Win7, is a solution to resolve the problem. I will try to develop a powershell script according to getver.pl and getlib.sh in some days.

> i'd be fine with replacing the getver.pl with a getver.sh, but that would still need a shell.

Do you think it is a good idea to replace getver.pl with getver.sh even in Linux? That makes libgd **ONLY** depends on shell env. So I think it's a good idea. @vapier

Issue #249, opened 2016-07-09
CMake and Makefiles build broken on Windows


pierrejoye:
@nathanaeljones any appealing reason to have changed the Makefile for something we do not actually control? I would rather stick to the struture we had for cmake and PHP, using binaries/prebuilt deps that are validated and widely used. Any comment to do something different welcome :)

CMake as of now is also broken, 

Using a simple command like:

`cmake -G "Visual Studio 14 2015" -DBUILD_TEST=1 -DENABLE_PNG=1 -DENABLE_JPEG=1 -DENABLE_FREETYPE=1 -DENABLE_WEBP=1 -DCMAKE_LIBRARY_PATH=c:\php-sdk\master\vc14\x64\deps\lib -DCMAKE_INCLUDE_PATH=c:\php-sdk\master\vc14\x64\deps\include`

Which simply requires to copy the deps llibs in ..\deps generates a clean build using all libraries we support. Deps prebuilt can be fetched here:

http://windows.php.net/downloads/php-sdk/

or can be built easily using:

https://github.com/winlibs

This is something I am not keen to change to some other structure or pre built binaries. 

Thoughts and comments more than welcome.

pierrejoye:
It also means we need to restore the cmake/modules directory as currently CMake on Windows is a big mess from a deps point of view.

Installing system wide builds files is an absolute no-go, especially when it comes to create clean and custom builds.

I have to check if we can have it used only for windows and keep using system ones for other platforms, if that helps.

cmb69:
> I would rather stick to the struture we had for cmake and PHP, using binaries/prebuilt deps that are validated and widely used.

+1

vapier:
I think before we seriously consider supporting the Windows builds, we need to get the appveyor CI stuff working. I spent a little time on it, but I've never used it before, and it's been years since I built under Windows (majority of my work has been with mingw).

along those lines, should we look at adding mingw to Travis? or would it not be that useful as we expect the majority of people to build under Windows using an MS toolchain?

pierrejoye:
On Jul 9, 2016 10:45 PM, "Mike Frysinger" <notifications@github.com> wrote:
>
> I think before we seriously consider supporting the Windows builds,

We do. And we used to run tests but things went off on my side after latest
2.1. But we do support windows.

we need to get the appveyor CI stuff working. I spent a little time on it,
but I've never used it before, and it's been years since I built under
Windows (majority of my work has been with mingw).

[...]

PR #534, opened 2019-10-27
FindLIQ.cmake: delete download libimagequant source code logic

willson-chen:
* And delete the LIQ_BUILD logic in CMakeLists.txt correspondingly.
* Fixed some indent errors.


cmake/modules/FindLIQ.cmake
```diff
@@ -4,15 +4,16 @@
 # This module defines
 #  LIQ_INCLUDE_DIR, where to find libimagequant.h
 #  LIQ_LIBRARIES, the libraries to link against to use libimagequant.
-#  LIQ_FOUND, If false, do not try to use libimagequant.
+#  LIQ_FOUND, if false, do not try to use libimagequant.
+#  LIQ_BUILD, if undefined, libgd will not download and build libimagequant
```
vapier: did you mean "will download" ?

i'm kind of the opinion we should delete all this logic ... gd shouldn't be downloading random code & executing it from the internet on people

willson-chen: > did you mean "will download" ?

If LIQ_BUILD is **undefined**, libgd will not download libimagequant. Correct?

> i'm kind of the opinion we should delete all this logic ... gd shouldn't be downloading random code & executing it from the internet on people

So we should not download libimaequant if it is not found, right?

vapier: > If LIQ_BUILD is undefined, libgd will not download libimagequant. Correct?

sorry, you're right of course.  double negatives in programming often throw me off which is why i suggested people avoid them :).

> So we should not download libimaequant if it is not found, right?

imo, yeah, we should delete all of `LIQ_BUILD` and everything in the `ELSE (LIQ_FOUND)` section.

willson-chen: Okey. I will improve it soon.

It would be greatly appreciated that you could spare some to review my rest PRs. I am close to finish building with cmake and ms toolchain under windows platform. And then we can close issue #249, which is an import part of milestone 2.2.6.

vapier:
thanks!

PR #536, opened 2019-10-29
Fix #164: update doc files install directory in CMakeLists.txt

### LINKED ISSUE #164: parity with CMake infrastructure vs autotools

willson-chen:
Change install directory from 'share/docs' to 'share/doc/gd'.


willson-chen:
`share/doc` is wilely used for the software doc files in Linux.

vapier:
`/usr/share/doc/` is fine, but normally packages do `<name>-<version>` rather than just `<name>`.  i think you can get that here with `${PACKAGE_NAME}-${GDLIB_VERSION}`.  i don't know if there's a cmake builtin that'd get you something equiv.

willson-chen:
> `/usr/share/doc/` is fine, but normally packages do `<name>-<version>` rather than just `<name>`. i think you can get that here with `${PACKAGE_NAME}-${GDLIB_VERSION}`. i don't know if there's a cmake builtin that'd get you something equiv.

`GDLIB_VERSION` contains `GDLIB_MAJOR`, `GDLIB_MINOR`, `GDLIB_REVISION` and `GDLIB_EXTRA`. What about `${PACKAGE_NAME}${GDLIB_MAJOR}`? In current version, that equals to `gd2`.
 I think libgd user does not want a new doc dir when a new revision or minor version is released.

vapier:
i'm aware it contains all of those.  it should be fine.

if you wanted to drop GDLIB_REVISION/GDLIB_EXTRA, that's prob fine too.

another option might be to just delete all of these installs and leave it to the distro to handle.  the autotools code isn't installing the doc files and people don't seem to have minded.

willson-chen:
I prefer droping GDLIB_REVISION/GDLIB_EXTRA. :)
Maybe some users need the doc files after intallation. So let's keep these installs.

willson-chen:
I checked the dir `/usr/share/doc`. Most of softwares use `${NAME}${MAJOR}.${MINOR}`, like `python2.7`, `python3.6`. So I delete the dash between ${NAME} and ${MAJOR}.

vapier:
lets stick with `gd-${GDLIB_MAJOR}.${GDLIB_MINOR}`.  python has existing naming that aligns with the dashless form that doesn't really apply to gd.

willson-chen:
> lets stick with `gd-${GDLIB_MAJOR}.${GDLIB_MINOR}`. python has existing naming that aligns with the dashless form that doesn't really apply to gd.

Sounds reasonable. It's done.

thx

PR #538, opened 2019-10-30
Fix #533: Remove cmake modules

### LINKED ISSUE #533: Remove cmake modules that already exist in cmake standard
The following cmake modules exist in cmake standard. So shoud we remove them from `cmake/modules`  and not to maintain them any more?
* FindFontconfig.cmake
* FindFreetype.cmake
* FindIconv.cmake
* FindJPEG.cmake
* FindPNG.cmake
* FindPackageHandleStandarArgs.cmake
* CMakeParseArguments.cmake

I tried to delete `FindJPEG.cmake/FindPNG.cmake/CMakeParseArguments.cmake`, and the building and test under linux still worked. I don't know why the rest modules conld not be found by cmake in its standard modules right now if I delete them, but I will have a further research in it.

willson-chen:
FindFreetype.cmake, FindPNG.cmake and FindJPEG, have been the cmake standard modules since cmake v2.6.


vapier:
if cmake-2.6 includes them, great!

PR #540, opened 2019-11-01
Fix #539: Add RAQM support for cmake

### LINKED ISSUE #539: Raqm support
1. There is `--with-raqm` option in `configure`, but no `ENABLE_RAQM` option for cmake.
2. travis test script doesn't contain any raqm related configuration.

willson-chen:
* Add `ENABLE_RAQM` option to CMakeLists.txt
* Add cmake module FindRAQM.cmake


vapier:
thx!

PR #543, opened 2019-11-06
Update cmake min version to 3.7

willson-chen:


willson-chen:
As we discussed in Issue #533. This PR is created to update cmake min version.
By the way. I should delete `FindPNG`, `FindJPEG` and `FindFreetype` in PR #538 . But I delete `CMakeParseArguments` instead of `FindFreetype` by mistake. So I delete `FindFreetype` and `FindPackageHandleStandarArgs` this time.

PR #554, opened 2019-11-16
Fixed #397: add C++ support

### LINKED ISSUE #397: Missing header(s) from install

willson-chen:
original patch by Galik. Related PR #398

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -114,6 +114,14 @@ fi
 AC_DEFINE_UNQUOTED([ENABLE_GD_FORMATS], [$gd_ac_value], [Whether to support gd image formats])
 AM_CONDITIONAL([ENABLE_GD_FORMATS], test "$gd_enable_gd_formats" = yes)
 
+AC_MSG_CHECKING([whether to support cpp wrapper])
+AC_ARG_ENABLE([cxx],
+  [AS_HELP_STRING([--enable-cxx], [Enable cpp wrapper])],
+  [gd_enable_cxx=$enableval],
+  [gd_enable_cxx=no])
+AC_MSG_RESULT([$gd_enable_cxx])
+AM_CONDITIONAL([ENABLE_CXX], test "$gd_enable_cxx" = yes)
+
 dnl Helper macro for working with external libraries.
 dnl GD_LIB_CHECK([SYM], [FEATURE], [name], [...test...])
 dnl $1 - upper case symbol
@@ -302,6 +310,7 @@ AC_MSG_RESULT([
 ** Configuration summary for $PACKAGE $VERSION:
 
    Support for gd/gd2 images:        $gd_enable_gd_formats
+   Support for cpp wrapper:          $gd_enable_cxx
    Support for Zlib:                 $gd_with_LIBZ
    Support for PNG library:          $gd_with_LIBPNG
    Support for JPEG library:         $gd_with_LIBJPEG
--- a/configure.ac
+++ b/configure.ac
@@ -37,6 +37,7 @@ AM_INIT_AUTOMAKE([1.11 foreign dist-xz -Wall -Werror subdir-objects])
 AC_CONFIG_HEADERS([src/config.h:src/config.hin])
 
 AM_PROG_AR
+AC_PROG_CXX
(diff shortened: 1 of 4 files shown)


vapier:
afaict, this is the same as the previous PR, which means none of the concerns have been addressed.  please see the discussion in #397.

willson-chen:
@vapier hi, two questions:
* should we support cpp in GD-2.3? by inistalling `gd_io_stream.h` and related files.
* we add `gdpp.cxx` and `gd_io_stream.cxx` to `src/CMakeLists.txt`, but exclude them in `src/Makefile.am`. what is the reason?

vapier:
we should have a --enable-cxx option to configure.ac so people can control it

would also be nice if we could link the code into libgd w/out requiring libstdc++, but i'm not sure how easy that'll be to pull off

willson-chen:
> we should have a --enable-cxx option to configure.ac so people can control it

easy to implement. but could you please tell me the reason? although most libgd users do not need c++ wrapper, integrating libgd with c++ sounds resonable considering gcc and g++ are always installed together in Linux.

and reference to my question 2, we should include `gdpp.cxx` and `gd_io_stream.cxx` only when option `--enable-cxx` or `-DENABLE_CXX` is supplied, right?

> would also be nice if we could link the code into libgd w/out requiring libstdc++, but i'm not sure how easy that'll be to pull off

we use iostream in `gd_io_stream.h`, and it depends on libstdc++. AFAIK, libgd with c++ wrapper independent of libstdc++ seems impossible.

willson-chen:
@vapier , hi, i added `--enable-cxx` option in my last commit. is it what we want? please have a review.

thx

vapier:
> integrating libgd with c++ without any option sounds resonable considering the relation between c and c++.

sorry but I don't know what you mean by relation.

libgd today does not require a C++ compiler or C++ runtime. that's good for embedded situations that don't have C++ compilers, and even on regular systems where we don't waste resources loading large and normally unused C++ runtimes.

a common idiom in other projects seems to provide in an entirely sep lib like libgdxx.so or libgdcpp.so. that would hold all the C++ bits and not penalize everyone else.

willson-chen:
> > integrating libgd with c++ without any option sounds resonable considering the relation between c and c++.
> 
> sorry but I don't know what you mean by relation.
> 
> libgd today does not require a C++ compiler or C++ runtime. that's good for embedded situations that don't have C++ compilers, and even on regular systems where we don't waste resources loading large and normally unused C++ runtimes.
> 
> a common idiom in other projects seems to provide in an entirely sep lib like libgdxx.so or libgdcpp.so. that would hold all the C++ bits and not penalize everyone else.

So a better way to support C++ is to generate `libgdcpp.so` when option `--enable-cxx` is on?

configure.ac
```diff
@@ -301,6 +310,7 @@ AC_MSG_RESULT([
 ** Configuration summary for $PACKAGE $VERSION:
 
    Support for gd/gd2 images:        $gd_enable_gd_formats
+   Support for cpp wrapper:          $gd_enable_cxx
```
vapier, line 313: "CPP wrapper" -> "C++ library"

vapier (changes requested):
> So a better way to support C++ is to generate `libgdcpp.so` when option `--enable-cxx` is on?

correct!

> If so, is `libgdcpp.so` the superset of `libgd.so`, or `libgdcpp.so` depends on `libgd.so`?

it would depend on `libgd.so`

pierrejoye:
I see this PR now. Please note I added the ENABLE_CPP option in CMake, so it makes perfect sense to have it in configure as well.

I do agree as well on having a libgdcpp library, depending on libgd.so

pierrejoye:
Are the changes done?

pierrejoye:
Won't add c++ wrapper or similar, it is relatively easy to do it if one needs it.

PR #556, opened 2019-11-18
Remove config.h in gdpp.h

willson-chen:
Reference to Issue #345 and PR #491

### DIFF
--- a/src/gdpp.cxx
+++ b/src/gdpp.cxx
@@ -49,7 +49,6 @@ namespace GD
 		case 0x47: // GIF
 			rtn = CreateFromGif(in);
 			break;
-#ifdef HAVE_LIBJPEG
 		/* JPEG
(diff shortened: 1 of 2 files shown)


vapier:
you'll need to update gdpp.cxx too to not refer to the header or the jpeg defines

willson-chen:
> you'll need to update gdpp.cxx too to not refer to the header or the jpeg defines

sorry. missed that. :)

pls review again.

willson-chen:
umm... the indentations in `gdpp.cxx` is a bit of strange. should i correct them?

```
namespace GD
	{
	bool Image::CreateFrom(FILE * in)
		{
		bool rtn;
		int c = fgetc(in);
```

vapier:
the style in the codebase is often inconsistent :/.  i wouldn't worry about it unless we can come up with some tooling to enforce it & make it easy.

PR #559, opened 2019-11-19
Fixed #369: fix new_a init error in gdImageConvolution()

### LINKED ISSUE #369: Artifacts when convoluting image with transparent background
If gdImageConvolution() is applied to an image with a transparent background, there can be black pixel artifacts (at the upper edge of opaque parts, and at the top left of the image), what appears to be a bug.

### Test program
````
#include <gd.h>


int main()
{
    gdImagePtr im;
    FILE *fp;
    float matrix[3][3] = {
        {1, 0, 1},
        {0, 5, 0},
        {1, 0, 1}
    };

    im = gdImageCreateTrueColor(40, 40);
    gdImageAlphaBlending(im, gdEffectReplace);
    gdImageFilledRectangle(im, 0, 0, 39, 39, 0x7fffffff);
    gdImageFilledEllipse(im, 19, 19, 20, 20, 0x00ff00);
    gdImageSaveAlpha(im, 1);

    gdImageConvolution(im, matrix, 9, 1);

    fp = fopen("47946.png", "wb");
    gdImagePng(im, fp);
    fclose(fp);

    gdImageDestroy(im);

    return 0;
}
````

### PR DESCRIPTION
**Pull request by willson-chen**

### CODE REVIEW

**Comment by willson-chen** [comment]
The result can be seen in #369 .

**Comment by cmb69** [comment]
Ah, thanks! This is certainly a correction of the old behavior (not initialized or wrongly initialized `pxl`). I wonder though, whether the alpha channel should not be convoluted as well (basically, similar calculations as for `new_r` etc.)

**Comment by willson-chen** [comment]
> Ah, thanks! This is certainly a correction of the old behavior (not initialized or wrongly initialized `pxl`). I wonder though, whether the alpha channel should not be convoluted as well (basically, similar calculations as for `new_r` etc.)

IMO, it depends on what we want. Alpha channel is used to reserve transparency information. And there is nothing different between the image with and without alpha convolution. When it comes to alpha compositing, it is hard to predict the result if two images are merged into one. Because alpha convolution will change the alpha value for each pixel.

**Comment by cmb69** [comment]
Well, you're right. At least for best BC, alpha convolution should not be introduced.

**Comment by willson-chen** [comment]
So can this PR be merged, please? @cmb69 @vapier

**Comment by cmb69** [comment]
The fix looks good to me, but I think there should also be an accompanying test case. Could you please add one?

**Comment by willson-chen** [comment]
> The fix looks good to me, but I think there should also be an accompanying test case. Could you please add one?

sure

**Comment by willson-chen** [comment]
@cmb69 hi, i added a test case. please review it. thx!

File: tests/gdimageconvolution/bug00369.c [hunk 1]
```diff
@@ -0,0 +1,35 @@
...
+		{1, 0, 1},
+		{0, 5, 0},
+		{1, 0, 0}
+	};
+
+	im = gdImageCreateTrueColor(40, 40);
+	gdImageAlphaBlending(im, gdEffectReplace);
+	gdImageFilledRectangle(im, 0, 0, 39, 39, 0x7FFFFFF);
+	gdImageFilledEllipse(im, 19, 19, 20, 20, 0x00FF00);
+	gdImageSaveAlpha(im, 1);
```
cmb69: This call is superfluous if the image is not saved (see http://libgd.github.io/manuals/2.2.5/files/gd-c.html#gdImageSaveAlpha).

tests/gdimageconvolution/bug00369.c
```diff
@@ -0,0 +1,35 @@
...
+	gdImageConvolution(im, matrix, 9, 1);
+
+	pxl = gdImageGetPixel(im, 0, 0);
+	if (!gdTestAssert(pxl != 0x7F010101)) {
+		error = 1;
+	}
+
+	gdImageDestroy(im);
+
+	return error;
```
cmb69: Instead of managing an own `error` variable, you *can* simply return `gdNumFailures()`.

willson-chen: reasonable. fixed.

tests/gdimageconvolution/bug00369.c
willson-chen: your are right. i will delete it.

willson-chen:
> Hmm, it seems that test case succeeds with and without the fix. Can you please check that?

Sorry. I made a mistake. But fixed now.

cmb69:
Applied as 4b0f3724022ad5bdf16d86409126436ff5a31afb.

PR #563, opened 2019-11-22
add testcase for gdImageCopyResized

willson-chen:
add testcase for gdImageCopyResized


cmb69:
Sorry for being late, but I think this test case should be improved to actually signal failure if `gdImageCopyResize()` fails to do what it is expected to do. As it is now, one may notice any failure when running the test in a Git clone, because gdimagecopyresized_exp.png would be modified then, but the test suite is not necessarily supposed to be run in a Git clone (and nobody should be required to check for such modifications). Instead, the test should better use `gdAssertImageEqualsToFile()` to verify that the created image is equal to a prepared image file (see e.g. [gdimagecopyresampled/basic_alpha.c](https://github.com/libgd/libgd/blob/master/tests/gdimagecopyresampled/basic_alpha.c)).

willson-chen:
Hi,

Thanks for your suggestion. I am trying to fixed this issue.

BTW, did you receive my last email about what I can do for the new release?
Looking forward to hearing from you.

Best regards,

Wilson

willson-chen:
I will create a new PR due to my operation of `git reset`.

PR #564, opened 2019-11-24
src/CMakeLists.txt: drop header files in add_library

willson-chen:
Header file will expend in source file in precompile stage. So header file is not necessary for `add_library` in CMakeLists.txt

### DIFF
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -1,12 +1,8 @@
 SET (LIBGD_SRC_FILES
-	bmp.h
 	gd.c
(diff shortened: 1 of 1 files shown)


vapier:
the cmake documentation is a bit unclear, but from random internet searching, there seems to be a slight advantage to listing the headers here in that it helps with autogeneration of other projects (e.g. for IDEs).

at the very least, it doesn't seem to cause problems (other than the overhead of listing them in the first place).

willson-chen:
fair enough

PR #565, opened 2019-11-25
Fixed #169: update var type to hold bigger w&h for ellipse

### LINKED ISSUE #169: gdImageEllipse() has overflow issue for moderately large radii

willson-chen:


willson-chen:
`int64_t` is defined in `stdint.h`

Reference to #169. In function `gdImageEllipse` and `gdImageFilledEllipse`, `r` should be signed. So I use `int64_t` instead of `uint64_t`.

vapier:
just assume `int64_t` and related types are available.  if we want glue code, we can do it in headers, not inline code.

willson-chen:
> just assume `int64_t` and related types are available. if we want glue code, we can do it in headers, not inline code.

could you give me more hint, please?

 We include `config.h` in `gd.c`, to see if `HAVE_STDINT` is defined. And then we can use `int64_t`. If we want to define a new type in header, such as `gd.h`, and it depends on `stdint.h`, we include `config.h` in `gd.h`?

vapier:
i'm saying use `int64_t` unconditionally here.  wait for someone to complain and we can figure out a header replacement if/when it happens.  these standards have been around for so long that it's not worth our time imo to splatter #ifdef throughout the codebase.

willson-chen:
> i'm saying use `int64_t` unconditionally here. wait for someone to complain and we can figure out a header replacement if/when it happens. these standards have been around for so long that it's not worth our time imo to splatter #ifdef throughout the codebase.

get it

willson-chen:
@vapier hi, could you please review again and offer more suggestion?

willson-chen:
> Thanks for the PR! Basically, this looks good to me, but I would prefer to check for valid inputs to these functions (e.g. what happens if `w < 0` and/or `h < 0`?), and to let the functions fail early (i.e. by returning `GD_FALSE` before drawing any pixels; changing the return type would be a BC break, though). This would require a thorough range analysis for all values, but if that is done, we likely could choose suitable types for the local variables.
> 
> Also, having (regression) tests appears to be reasonable.

I will add a test case. Since changing return type will cause BC break, shall we leave it to GD3? Or GD2.3 is fined?

willson-chen:
@cmb69 hi, what about improving this PR by just adding test case without modifying the return type to keep API compatible. But I will leave an issue to remind us change it in GD3, and it's better to add it to milestone 3.0.

cmb69:
Adding a return type is better left for GD 3.0. The current PR plus a test case would be fine for me. Thanks.

willson-chen:
Test will failed in 32-bit platform in current master branch of GD, and will success with this patch.
test_master

tests/gdimageellipse/bug00169.c
```diff
@@ -0,0 +1,22 @@
...
+int main()
+{
+	gdImagePtr im;
+	int white = 0;
+
+	im = gdImageCreateTrueColor(2200, 2200);
+	white = gdImageColorAllocate(im, 255, 255, 255);
+
+	gdImageEllipse(im, 1100, 1100, 2200, 2200, white);
+  gdAssertImageEqualsToFile("gdimageellipse/bug00169_exp.png", im);
```
cmb69: Nit: indentation (also in the other test below)

willson-chen: Fixed!

Sorry I used a new computer with vim config auto expanding tab to spaces.

willson-chen: Thanks for your careful review :)

PR #570, opened 2019-11-30
Fixed #558: Move demo files to examples

### LINKED ISSUE #558: Some confusing files in src
Some source file in src dir is confusing. They are neither used for libgd.so nor tools,
but look more like for test.

So should they be move to tests dir or be droped?

File list:
gd_color_map_test.c  gifanimtest.c  circletexttest.c  gddemo.c  gdtest.c ...

willson-chen:


willson-chen:
reference to #558

PR #575, opened 2019-12-04
Fix #301: brect bounds

### LINKED ISSUE #301: gdImageStringFT()'s brect out-of-bounds
I've noticed that a [PHP test case](https://github.com/php/php-src/blob/PHP-7.0.11/ext/gd/tests/bug48732.phpt) fails when PHP has been compiled against an external libgd, but with the same libfreetype.

Consider the following slightly modified transscript of the test:

```
#include "gd.h"
#include "gdtest.h"


int main()
{
    gdImagePtr im;
    int black, red, green;
    char *fontpath = gdTestFilePath2("gdimagestringft", "Tuffy.ttf");
    int brect[8];
    FILE *fp;
    int i;

    im = gdImageCreate(150, 50);
    gdImageColorAllocate(im, 255, 255, 255);
    black = gdImageColorAllocate(im, 0, 0, 0);
    red = gdImageColorAllocate(im, 255, 0, 0);
    green = gdImageColorAllocate(im, 0, 128, 0);
    gdImageLine(im, 10,0, 10,49, green);
    gdImageStringFT(im, brect, black, fontpath, 12, 0, 10, 20, "ABCEDFGHIJKLMN\nopqrstu\n");
    gdImageRectangle(im, brect[6],brect[7], brect[2], brect[3], red);

    fp = fopen("bug00301.png", "wb");
    gdImagePng(im, fp);
    fclose(fp);

    for (i = 0; i < 8; i += 2) {
        printf("%d, %d\n", brect[i], brect[i+1]);
    }
    gdImageDestroy(im);
    gdFree(fontpath);

    return gdNumFailures();
}
```

This results in the following image:
bug00301
The green line is drawn at the given x-coordinate for gdImageStringFT() (i.e. 10), the red rectangle is drawn at the resulting brect coordinates. One can see that the text indeed starts at x=10, but the bounding rectangle extends 2 pixels to the left, what looks wrong to me.

Also note that the test program outputs:

```
8, 41
140, 41
140, 6
8, 6
```

willson-chen:


willson-chen:
libgdAfter
libgdBefore
phpgd

PR #578, opened 2019-12-12
gdcache: fix gdcache test compile cmd

willson-chen:


src/gdcache.c
```diff
@@ -25,7 +25,9 @@
  * John Ellson  (ellson@graphviz.org)  Oct 31, 1997
  *
  * Test this with:
- *               gcc -o gdcache -g -Wall -DTEST gdcache.c
+ *               gcc -o gdcache -g -Wall -DTEST -DNEED_CACHE gdcache.c -lgd
```
willson-chen: libgd must be installed in this case

src/gdcache.c
```diff
@@ -25,7 +25,9 @@
...
+ *               or
+ *               gcc -o gdcache -g -Wall -DTEST -DNEED_CACHE gdcache.c libgd.a
```
willson-chen: we should specify the path of libgd.a in this case

pierrejoye: Is it solved?

willson-chen:
I don't know if someone will use the test code. But the compile cmd is incomplete.

PR #476, opened 2018-09-17
Adjusting CMakeLists.txt for #472

kohldampfer:
This commit should solve the problem described in Issue #472

PR #582, opened 2019-12-27
Fixed #472: Adjusting CMakeLists.txt

### LINKED ISSUE #472: CMakeLists.txt: Required packages when option is enabled
I just tried to compile libgd with JPEG functionality on linux, i.e. calling `cmake -G "Unix Makefiles" -DENABLE_JPEG=ON .`. During the process of CMake, I got this message:

> -- Could NOT find JPEG (missing: JPEG_LIBRARY JPEG_INCLUDE_DIR) 

and I overlooked it because CMake not stopped here. So I saw only later, that this JPEG part was not compiled with libgd.

So, my question here is: As JPEG functionality (and other picture formats) are handled through CMake options (like `ENABLE_JPEG`), why not put an error when someone wants to add this functionality? I.e. when option is on, CMake should fail, if it could not find the packages.

What do you think about this? If there is interest, I can provide a fix and a PR.

willson-chen:
Ported from PR #476. And added `RAQM REQUIRED`.


cmake/modules/FindLIQ.cmake
```diff
@@ -28,6 +28,10 @@ IF (LIQ_FOUND)
   IF (NOT LIQ_FIND_QUIETLY)
     MESSAGE(STATUS "Found LIQ: ${LIQ_LIBRARY} ${LIQ_INCLUDE_DIR}")
   ENDIF (NOT LIQ_FIND_QUIETLY)
+ELSE (LIQ_FOUND)
+  IF (LIQ_FIND_REQUIRED)
+	MESSAGE(FATAL_ERROR "Could not find LIQ library")
+  ENDIF (LIQ_FIND_REQUIRED)
```
willson-chen: Most standard `FindXXX.cmake` cmake module call `FIND_PACKAGE_HANDLE_STANDARD_ARGS` to throw exception when module is not found. The `IF` branch will not terminate cmake process. So I am considering update this code block to `FIND_PACKAGE_HANDLE_STANDARD_ARGS`.

travis/main.sh
```diff
@@ -104,7 +104,6 @@ build_cmake() {
 		-DENABLE_PNG=1
 		-DENABLE_TIFF=1
 		-DENABLE_WEBP=1
-		-DENABLE_XPM=1
```
willson-chen: `libxpm-dev` is unavaible in brew repo, while avaible in apt repo. But I should not delete `-DENABLE_XPM=1` directly. A better way should be enabling different feature according to different OS.

willson-chen: Besides `libxpm-dev`, `libraqm-dev` faces the same problem.

travis/main.sh
```diff
@@ -90,9 +90,14 @@ cmake_args=(
 	-DENABLE_PNG=1
 	-DENABLE_TIFF=1
 	-DENABLE_WEBP=1
-	-DENABLE_XPM=1
 )
 
+# libxpm-dev is unavaible in brew repo
+# Once it gets avaible, please modify this code block.
+if [ "${TRAVIS_OS_NAME}" == "linux" ]; then
```
vapier, line 97: this syntax is incorrect: always use `[[` in bash, not `[`

travis/main.sh
```diff
@@ -90,9 +90,14 @@ cmake_args=(
...
+	cmake_args[${#cmake_args[@]}]="-DENABLE_XPM=1"
```
vapier, line 98: simply append the array: `foo+=( -D... )`

PR #591, opened 2020-01-22
add testcase gdimagecopyresized dev

willson-chen:


willson-chen:
Reference to PR #563. That PR is not good but merged by mistake (thanks for suggestion from @cmb69). I reseted the master branch, improved the test case and created a new PR here.

Please review again

cmb69:
The test case looks okay. Thanks!

Regarding the background color of images: when a new image is created, all its colors are set to zero. For true color images that means the background is initially black; for palette images that means the initial background color is given by the first allocated color. Hence, the `imagecolorallocate(dst_tc, ...)` is useless here.

willson-chen:
> The test case looks okay. Thanks!
> 
> Regarding the background color of images: when a new image is created, all its colors are set to zero. For true color images that means the background is initially black; for palette images that means the initial background color is given by the first allocated color. Hence, the `imagecolorallocate(dst_tc, ...)` is useless here.

Get it. I deleted the useless code. thx! :)

PR #598, opened 2020-02-24
Fix #597: add codecov support

### LINKED ISSUE #597: add codecov support
I think we need a way to show our code coverage. [codecov](https://codecov.io/) is a good choice.

willson-chen:
1. travis/main.sh: generate code coverage report and submit to codecov.io
2. README.md: add codecov.io token


willson-chen:
The last comment from codecov-io shows there is no coverage uploaded. And I think this is because we never upload coverage before for master branch. But I am not sure about that.

willson-chen:
I gonna add some test case. The codecov will show us the improvement. So let me merge it if you don't mind.

PR #601, opened 2020-03-10
Make gd_nnquant.c less likely to introduce duplicate definitions

d0k:
This file contains some accidentally global symbols with generic names.
Give them internal linkage (or remove if unused) to avoid conflicts when
linking another library that happens to use the same generic names.


willson-chen:
LGTM. I will merge it in some days if no more code reviews.

Thanks for your contribution. :)

PR #606, opened 2020-03-20
Call free function

klebertarcisio:
I think in this case we need to call the free function in more places. What do you think?

### DIFF
--- a/tests/gif/bug00227.c
+++ b/tests/gif/bug00227.c
@@ -60,9 +60,20 @@ int main()
 	fp = fopen(file, "rb");
 	if (!fp) return 1;
 	buf = malloc(PROBE_SIZE);
-	if (!buf) return 1;
-	if (PROBE_SIZE != fread(buf, 1, PROBE_SIZE, fp)) return 1;
-	if (buf[PROBE_SIZE-1] & 0x80) return 1;
+	if (!buf) {
+		fclose(fp);		
+		return 1;
+	}
+	if (PROBE_SIZE != fread(buf, 1, PROBE_SIZE, fp)){ 
+		free(buf);
+		fclose(fp);		
+		return 1;
+	}
+	if (buf[PROBE_SIZE-1] & 0x80){ 
+		free(buf);
+		fclose(fp);
+		return 1
+	};
 	free(buf);
 	fclose(fp);
 
--- a/src/gdcache.c
+++ b/src/gdcache.c
@@ -197,7 +197,7 @@ static void cacheRelease(void *map)
 	gdFree((char *)map);
 }
 
-int main(char *argv[], int argc)
+int main(int arg, char **argv)
 {
 	gdCache_head_t *cacheTable;
 	int elem, key;


vapier:
the fix to `main` is good.  we want that.

as for the cleanups, i don't think we want those.  if these were library calls, you'd def be correct that we don't want to leak resources.  but these are unittests and as soon as we exit the process, the kernel will take care of releasing all resources for us.  so adding error checking doesn't gain anything other than making the code more complicated.

klebertarcisio:
Thanks for the answer.
I also agree that the code is complicated, but I have a suggestion. What do you think of joining two conditions in just one? Would that make the code clearer?

```
	if (PROBE_SIZE != fread(buf, 1, PROBE_SIZE, fp) || (buf[PROBE_SIZE-1] & 0x80)){ 
		free(buf);
		fclose(fp);		
		return 1;
	}
```

I believe that stopping calling the free function is not the best solution even in this test case. This was not the developer's initial goal because he called the free function at the end of the code.

willson-chen:
@klebertarcisio hi,
I have merged the fix to main in PR #568. But I find it is merged into branch GD-2.2 instead of master branch.

@vapier hi,
I agree with @klebertarcisio. The kernel will release the resource when the test case terminates, but it is still a good habit to release it manually for a developer. And I have fixed some memory leakage for test case in PR #604 #605. So do you minding I merging it after some improvement, please?

tests/gif/bug00227.c
```diff
@@ -60,9 +60,20 @@ int main()
...
+	if (PROBE_SIZE != fread(buf, 1, PROBE_SIZE, fp)){ 
+		free(buf);
+		fclose(fp);		
+		return 1;
+	}
+	if (buf[PROBE_SIZE-1] & 0x80){ 
+		free(buf);
+		fclose(fp);
+		return 1
+	};
```
willson-chen: I suggest to use `goto` to make code simple. `goto` is not recommanded in lib code, because it will make the code hard to maintain. But considering the low complexity of the test case code, I think it is fine.

willson-chen: And please take attention to the CI failure. :)

vapier:
error code paths are barely, if ever, tested.  i will bet hard currency that no one has ever validated the majority of error code paths in our unittests.

attempting to write them correctly in unittests is a waste of time, both in code maintenance, and in having people think & reason about the code.  it's purely a waste of people's time.

so objectively speaking, they're a negative to the project.  there is no value to manually calling free/close on objects/handles that the kernel discards for you.  it isn't a memory leak.

willson-chen:
@vapier, thanks for your suggestion.

@klebertarcisio, so please close this PR. I will cherry-pickup your commit in PR #568, which also fix the wrong main function.

cmb69:
In my opinion, there is no need to free resources in case the test fails, because the test is not expected to fail in the first place. If it fails, the test failure has to be addressed, and only if the test passes (again), it's useful to check for memory mismanagement (e.g. leaks).

klebertarcisio:
I would like to thank @willson-chen, @vapier and @cmb69 for participating in the discussion. I really believe that the discussion is more important than the result of the pull request. I will continue to contribute to the project. Thanks.

PR #616, opened 2020-03-27
Fix two build configure errors in tests

willson-chen:
tests/webp/bug_double_free should depend on both webp and jpeg.
tests/gdimagesquaretocircle should depend on png.


willson-chen:
@cmb69 Thanks for your careful review. :)

PR #617, opened 2020-03-28
cmake: remove required host includes

neheb:
This breaks cross-compilation. More info:
https://github.com/openwrt/packages/pull/10079


willson-chen:
Hi, @neheb. Could you please specify why make this change? I can not fully understand the link you provided.

neheb:
Instead of target headers being included, host headers get included when cross compiling.

willson-chen:
I see. I will merge this PR if no more review in some days.

Thanks for your contributions. :)

PR #620, opened 2020-04-06
gdlib.pc: use prefixes for pkgconfig file

neheb:
This allows the pkgconfig file to be properly used in cross compile
scenarios by overriding the prefix. Otherwise, it points to host
libraries.

Signed-off-by: Rosen Penev <rosenp@gmail.com>


willson-chen:
LGTM. will merge it in some days if no more reviews.

vapier:
lgtm, thx

PR #622, opened 2020-04-10
src/config.h.cmake: use #cmakedefine01 for ENABLE_GD_FORMATS

willson-chen:
Although we have [docs/README.CMAKE](https://github.com/libgd/libgd/blob/master/docs/README.CMAKE) describ the avaiable options are ENABLE_XXX=1/0, some users are familar with ENABLE_XXX=ON/OF.  If ENABLE_GD_FORMATS=ON is used, it will result that GD formats is not enabled actually.

### DIFF
--- a/src/config.h.cmake
+++ b/src/config.h.cmake
@@ -4,7 +4,7 @@
 #cmakedefine BGDWIN32
 
 /* Whether to support gd image formats */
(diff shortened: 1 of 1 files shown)


vapier:
this header file is generated (like the first line says), so you can't hand edit it.  if you want to change it, you have to change the `bootstrap.sh` instead.

PR #633, opened 2020-05-02
Fix #615: gdImageStringFT() fails for empty strings as of libgd 2.3.0

### LINKED ISSUE #615: gdImageStringFT() fails for empty strings as of libgd 2.3.0
When trying to render an empty string via `gdImageStringFT()`, the function fails as of libgd 2.3.0, while formerly it was just a no-op.

#### Test script
````.c
#include <stdio.h>
#include <gd.h>

int main()
{
    gdImagePtr im;
    int fg;
    int rect[8];
    char *res;

    im = gdImageCreate(100, 100);
    fg = gdImageColorAllocate(im, 255, 255, 255);
    res = gdImageStringFT(im, rect, fg, "/usr/share/fonts/truetype/dejavu/DejaVuSans.ttf", 12, 0, 10, 10, "");
    if (res) printf("%s\n", res);
    return 0;
}
````
#### Expected output
````
okay
````
#### Actual output
````
Problem doing text layout
````

cmb69:
We change the return type of `textLayout()` to `ssize_t`, and signal
failure by returning `-1`, so that laying out an empty string is no
longer handled as failure.  We make sure that no overflow occurs,
assuming that all `int` values can be fully represented as `ssize_t`.


willson-chen:
LGTM. Thanks!

Issue #634, opened 2020-05-03
Build failed with automake


willson-chen:
Env
```
gd version: master branch. commit id`188aa09`
OS: Ubuntu 18.04 64-bit
CPU: x86_64
gcc: v7.5
```

Error Message
```
libtool: compile:  gcc -DHAVE_CONFIG_H -I. -I../../src -I/usr/include/libpng16 -I/usr/include/freetype2 -I/usr/include/libpng16 -I/usr/include/freetype2 -I/usr/include/libpng16 -I/usr/include/freetype2 -I/usr/include/libpng16 -I/usr/include/x86_64-linux-gnu -Werror -g -O2 -fvisibility=hidden -Wall -MT gdft.lo -MD -MP -MF .deps/gdft.Tpo -c ../../src/gdft.c  -fPIC -DPIC -o .libs/gdft.o
In file included from /usr/include/freetype2/freetype/freetype.h:33:0,
                 from ../../src/gdft.c:152:
/usr/include/freetype2/freetype/config/ftconfig.h:61:0: error: "HAVE_UNISTD_H" redefined [-Werror]
 #define HAVE_UNISTD_H 1
 
In file included from ../../src/gdft.c:18:0:
../../src/config.h:88:0: note: this is the location of the previous definition
 #define HAVE_UNISTD_H
 
In file included from /usr/include/freetype2/freetype/freetype.h:33:0,
                 from ../../src/gdft.c:152:
/usr/include/freetype2/freetype/config/ftconfig.h:63:0: error: "HAVE_STDINT_H" redefined [-Werror]
 #define HAVE_STDINT_H 1
 
In file included from ../../src/gdft.c:18:0:
../../src/config.h:76:0: note: this is the location of the previous definition
 #define HAVE_STDINT_H
```

vapier:
how are you building this exactly ?  this is not the behavior i see.

```sh
$ automake --version | head -1
automake (GNU automake) 1.15.1
$ autoconf --version | head -1
autoconf (GNU Autoconf) 2.69
$ ./bootstrap.sh
$ ./configure
$ grep HAVE_UNISTD_H src/config.h
#define HAVE_UNISTD_H 1
```

willson-chen:
The outputs of these commands in my pc are exactly same with yours, including the version of automake and autoconf.

I have cloned a new GD repo in same pc, but can not reproduce this issue.

Maybe this is a exception?

willson-chen:
[...]

PR #635, opened 2020-05-05
src/gdft.c: fix macro redefinition

willson-chen:
HAVE_STDINT_H and HAVE_UNISTD_H is define in ft2 header.
We need to undef these two macro before include ft2 header.


willson-chen:
Ref to issue #634. In fact, I don't think this is a good solution to fix this issue. But I can't find a better way right now. Suggestions are welcom. :)

PR #639, opened 2020-05-16
Retain transparency when using gdImageCopyResized on palette PNG

### LINKED ISSUE #638: gdImageCopyResized can lose transparency when converting palette png
Test case is a palette png, top half transparent white, bottom half black:

src

The result of gdImageCopyResized is an entirely black png, the top half transparency was lost:

dest

Expected:

exp

(NB. I have a test case/possible fix to review in an upcoming PR)

markdingram:
Fixes #638


markdingram:
Ah! Understood. 

I started off down this path investigating this use of [gdImageCopyResampled](https://github.com/nginx/nginx/blob/4bf4650f2f10f7bbacfe7a33da744f18951d416d/src/http/modules/ngx_http_image_filter_module.c#L934), in Nginx's ngx_http_image_filter_module, which ends up calling gdImageCopyResized for Palette PNGs. I'll pursue this further on the Nginx mailing list.

PR #641, opened 2020-05-24
gddemo.c: move from src to examples

willson-chen:
Reference to #558


cmb69:
Moving this file looks overdue, but I wouldn't do it in a revision (and actually not even in a minor release), so I think we should decide how to handle the next release. Will master become 2.3.1 or 2.4?

willson-chen:
> but I wouldn't do it in a revision (and actually not even in a minor release), so I think we should decide how to handle the next release.

So it is better to leave it to GD 3?

> Will master become 2.3.1 or 2.4?

Sure. We have [milestone list](https://github.com/libgd/libgd/milestones).

vapier:
time to merge!

vapier (approved):

PR #644, opened 2020-05-29
fit build error in vs2010

gergul:


src/gdft.c
```diff
@@ -35,6 +35,12 @@
 #ifndef R_OK
 #define R_OK 04			/* Needed in Windows */
 #endif
+
+#define ssize_t int
```
cmb69, line 39: This looks like a bad idea. If we really want to support old compilers, we should define `ssize_t` only if it's not available. And I would also suggest to use a type with the proper size (i.e. `long long` or so for x64).

gergul: Thanks suggest. I have fix it again.

cmb69:
For some reason CI hasn't been run for the latest commit. If it had run, the build had failed because basetsd.h and `SSIZE_T` are only available on Windows (but CI runs on Ubuntu and macOS). We need a *portable* solution.

@gergul, do you have any particular reason to still use VS2010?

gergul:
But I add my code to "#if defined(_WIN32) " section. So I think it's ok :)

cmb69:
I had a closer look, and actually `ssize_t` isn't known by latest MSVC (VS 2019). It may also not be known by other compilers (it's defined by POSIX, but not by C99). So I think detection and fallback definition should go to config.h (generated by ./configure). 

Wrt. the `FT_Pos` issue: that relies on the C99 feature of mixing declarations and statements. It seems to me that is something we want to use (sometimes), and I think we can require a mostly C99 compatible compiler nowadays. For MSVC, that would require VS 2015 (maybe VS 2013 already supports it; VS 2012 definitely doesn't).

PR #649, opened 2020-07-15
Fix #615 using libraqm

### LINKED ISSUE #615: gdImageStringFT() fails for empty strings as of libgd 2.3.0

remicollet:
@cmb69 please review

raqm_get_glyphs fails but this seems expected when len is 0

My first idea was 
```
 	if (!glyphs) {
 		raqm_destroy (rq);
-		return -1;
+		return len ? -1 : 0;
 	}
```

But testing earlier make things faster in all cases.

### DIFF
--- a/src/gdft.c
+++ b/src/gdft.c
@@ -449,6 +449,10 @@ textLayout(uint32_t *text, int len,
 	size_t count;
 	glyphInfo *info;
 
+	if (!len) {
+		return 0;
+	}
+
 #ifdef HAVE_LIBRAQM
 	size_t i;
 	raqm_glyph_t *glyphs;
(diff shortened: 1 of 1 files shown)


remicollet:
Squashed and Merged, thanks

PR #650, opened 2020-07-18
Improve building with CMake

YakoYakoYokuYoku:
Basically this fixes builds that enable the GD formats.


CMakeLists.txt
```diff
@@ -20,6 +20,7 @@ OPTION(ENABLE_FREETYPE "Enable Freetype2 support" 0)
 OPTION(ENABLE_FONTCONFIG "Enable FontConfig support" 0)
 OPTION(ENABLE_WEBP "Enable WebP support" 0)
 OPTION(ENABLE_RAQM "Enable RAQM support" 0)
+OPTION(ENABLE_ALL_OPTIONS "Enable all the options described above" 0)
```
willson-chen, line 23: I think few users will use this option, which means they have to install all the dependencies. We prefer users can enable options one by one, which make them to check the options carefully.

src/gd_gd.c
```diff
@@ -60,7 +60,7 @@
 #include "gd_errors.h"
 
 /* 2.3: gd is no longer mandatory */
-#if ENABLE_GD_FORMATS
+#ifdef ENABLE_GD_FORMATS
```
willson-chen, line 63: `#if ENABLE_GD_FORMATS` works fine in my environment. Did you encounter any problems, please? :)

src/gd_gd2.c
```diff
@@ -85,7 +85,7 @@
 /* 2.03: gd2 is no longer mandatory */
 /* JCE - test after including gd.h so that HAVE_LIBZ can be set in
  * a config.h file included by gd.h */
-#if defined(HAVE_LIBZ) && ENABLE_GD_FORMATS
+#if defined(HAVE_LIBZ) && defined(ENABLE_GD_FORMATS)
```
willson-chen, line 88: Some as above.

willson-chen:
Hi @YakoYakoYokuYoku , thanks for your patch. I have reviewed it and leave some comments. Basically I don't think the improvement is necessary, and discussions are welcom. :)

src/gd_gd.c
YakoYakoYokuYoku: The issue is that the resulting library does not have support for the GD formats, unless this change is done.

willson-chen: When the option `ENABLE_GD_FORMATS` is set, there will be the following code in `src/config.h`:

```
#define ENABLE_GD_FORMATS 1
```

which make `#if ENABLE_GD_FORMATS` equal to `#if 1` and work fine.

YakoYakoYokuYoku:
I submitted a PR fixing an issue that has already been fixed :flushed:, sorry for taking your time.

willson-chen:
It is fine. :)

Thanks for you concern about GD. More issues and PRs are welcome.

Issue #401, opened 2017-07-26
visual studio builds lack getopt


vshulga:
See https://github.com/libgd/libgd/commit/2db153a93e22c9cd53022e131915c1000259eb52

vapier:
we can probably just import a getopt implementation from like FreeBSD for VS builds

cmb69:
Maybe https://gist.github.com/superwills/5815344 is an option.

cmb69:
@vapier I've had a closer look at the getopt replacement, and the one mentioned above appears to work fine. However, MSVC doesn't know about `mkstemp()` so I've employed the following replacement:
````
#ifdef _MSC_VER
/* MSVC doesn't know mkstemp(), sigh ... */
#include <fcntl.h>
int mkstemp(char *template)
{
	int fd;

	do {
		if (!_mktemp(template)) {
			return -1;
		}
		fd = _open(template, _O_RDWR|_O_CREAT|_O_EXCL);
	} while (fd == -1 && errno == EEXIST);
	return fd;
}
#endif
````
This also works fine so far, but apparently ignores the "b" flag in the [fdopen()](https://github.com/libgd/libgd/blob/b7977b574dbbb1ba7d3e109e55ab08b7d76fb42a/src/webpng.c#L275) call, so everytime an image is modified by webpng (`-i` or `-t`) it gets corrupted. Any ideas on how to work around this? Is it even necessary to write to a temporary file in the first place – can't we simply write to `infile` instead? Since `gdImagePng()` doesn't report success/failure, it doesn't seem to make a real difference.

vshulga:
maybe std::tmpfile()?

vapier:
hopefully master should be fixed now for windows builds.  if you're still seeing getopt problems, let us know.  if you're seeing other portability problems (unrelated to getopt), please file a new issue.

PR #659, opened 2020-10-27
Remove unistd.h from getopt.c

stima:
1. Correct Windows cmake build: unistd.h doesn't exist on Windows.
2. Handle getopt missing declaration while building with ENABLE_PNG

### DIFF
--- a/src/getopt.c
+++ b/src/getopt.c
@@ -33,7 +33,6 @@
 #include <stdio.h>
 #include <stdlib.h>
 #include <string.h>
-#include <unistd.h>
 
 int	opterr = 1,		/* if error message should be printed */
 	optind = 1,		/* index into parent argv vector */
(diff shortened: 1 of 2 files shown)


cmb69:
Good catch, but isn't that header required for other platforms?

stima:
getopt.c was added (https://github.com/libgd/libgd/issues/401) to implement missing functionality on windows and it is guarded by cmake if (WIN32).

stima:
there is also missing mkstemp, I'm going to add it but with next PR - wanna to think what is the better replacement.

stima:
Should I do something more to be it merged?

willson-chen:
hi @stima , thanks for your contribution.

This PR is similar to #512, which is created by me. @cmb69 and @vapier, could you please spare some time to offer some suggestion about this issue? :)

cmb69:
It seems to me that PR #512 is more complete, so we should prefer that one.

stima:
Ok, if you don't mind I left few

pierrejoye:
This is fixed in master

PR #660, opened 2020-10-29
changes for oss fuzz with libtiff and libjpegturbo

hardik05:
`by` defailt libtiff and libjpegturbo was disabled in libbgd oss fuzz. enabled it.


pierrejoye:
good point, I cannot remember who added it, but it would be nice to care about it. Or do we have access by default if member of libgd?

pierrejoye:
@hardik05 hello :)

We have gh actions, we can add that to them as part of the CI.

Does it support ARM btw?

hardik05:
hello,

even I dont have any idea.. I just read the doc and added support for libtiff and libjpeg turbo.

PR #668, opened 2021-01-28
release-2.3.1

willson-chen:


CHANGELOG.md
```diff
@@ -4,6 +4,42 @@ All notable changes to this project will be documented in this file.
...
+- webp: support pkg-config file
+- gd_io: replace internal Putchar with gdPutC
+- gd_io: trim unused Putword function
+- Fix indent error: travis/main.sh
+- Fixed #472: Adjusting CMakeLists.txt (#582)
+- Fix #615: gdImageStringFT() fails for empty strings as of libgd 2.3.0 (#633)
+- Fix typo but preserve BC
+- Fixed test case build configure error in issue #626
+- Fix two build configure errors in tests (#616)
+- .travis.yml: close config warning and info (#642)
```
vapier, line 23: the changelog is meant as release notes readable by the end user to know what has changed.  we should omit entries like this that only impact the CI and have no bearing on the release package.

vapier:
seems reasonable

i think we should branch GD-2.3 once this goes out and open up master for GD-2.4.  i see some APIs we deprecated long ago but haven't yet removed.

cmb69:
I don't think that we should remove deprecated functionality in a minor version, but I don't see any compelling reason not to target 3.0 instead of 2.4 after 2.3.1.

vapier:
to be clear, when i say "remove", i mean "remove from header", not "delete from library ABI".  so existing apps will continue to work fine.

vapier:
that said, i'm happy to go to 3.0 as i'm seeing other support that i would like to delete that would be better justified with a larger version increase :)

willson-chen:
So, I will pull the GD-2.3 branch and open up master for GD-3.0

vapier:
lets not do both things at once.  just create 2.3.1 here and do the release from master, then we can branch.

Issue #557, opened 2019-11-18
Avif support ?


Nuranto:
I know the support is still almost inexistant.... Just supported in Windows10 since May, and no browser supports it atm, as far as I know. (There is however a polyfill : https://github.com/Kagami/avif.js)
But the AVIF format was born earlier this year, and support will probably arrive soon everywhere (AOM initiative, from google, mozilla, apple, etc...)
The promise of this format is big, and it should - finally - be a real replacement for our current outdated formats (jpg, ..).

avif spec : https://aomediacodec.github.io/av1-avif/
libavif : https://github.com/AOMediaCodec/libavif

willson-chen:
good suggestion. IMHO, avif has a bright future. what do you think of it? @vapier

vapier:
if there's a standard lib for it that we can find via pkg-config, sounds find to have as optional

Nuranto:
@vapier I'm not familiar with `pkg-config` but could this help ?  
=> https://github.com/AOMediaCodec/libavif/blob/master/libavif.pc.cmake

Nuranto:
Note : There's now a caniuse result for avif support : https://caniuse.com/#feat=avif
(Firefox has implemented avif support in nightly : https://www.ghacks.net/2020/05/04/mozilla-implements-experimental-avif-image-support-in-firefox-nightly/)
It's coming 👍

willson-chen:
Hi @Nuranto. Thanks for your concern and help.

morsssss:
Hey, all! I'm interested in taking this on - possibly with a collaborator at my company.

I know other folks who've integrated [libavif](https://github.com/AOMediaCodec/libavif) into a project without undue trouble :)

I've not contributed to this project before. Are there guidelines a first-time contributor should follow aside from what I see [here](https://github.com/libgd/libgd/blob/master/CONTRIBUTING.md)?

vapier:
that doc should about cover it.  if it's missing something, we should def add it!

morsssss:
Great - thanks @vapier !

morsssss:
Update: with some kind guidance from @vapier , I am making good progress on this. Stretch goal: submit PR by the end of Jan.

cryptomilk:
@morsssss Do you have a work in progress branch somewhere? I've implemented avif support in darktable using libavif. I can offer to do a code review ...

morsssss:
[...]

Issue #395, opened 2017-06-25
Support for HEIF format


alexbilbie:
Apple have announced that all of their devices running iOS 11 and macOS 10.13 will support HEIF (High Efficiency Image File) format.

From [Wikipedia](https://www.wikiwand.com/en/High_Efficiency_Image_File_Format):

> High Efficiency Image File Format (HEIF) is a file format for individual images and image sequences. It was developed by the Moving Picture Experts Group (MPEG) and is defined by MPEG-H Part 12 (ISO/IEC 23008-12).
>
> The HEIF specification also defines the means of storing High Efficiency Video Codec (HEVC)-encoded intra images and HEVC-encoded image sequences in which inter prediction is applied in a constrained manner.
> 
> HEIF files are compatible with the ISO Base Media File Format (ISOBMFF, ISO/IEC 14496-12) and can also include other media streams, such as timed text and audio.

There is an open source C++ encoder and decoder implementation available from here - https://github.com/nokiatech/heif.

Is there potential for libgd to support this new format?

victorstewart:
+1

iipokypatop:
+1

cmb69:
Please do not post `+1` comments, but rather "Add your reaction" to the OP, if you feel the need.

iipokypatop:
+1

810:
What’s new in Build 17123

Introducing High Efficiency Image File Format (HEIF) in Windows 10

The High Efficiency Image File Format (HEIF) has arrived for Windows 10 and the Photos app! HEIF is an image container that leverages modern codecs like HEVC to improve quality, compression, and capabilities compared to earlier formats like JPEG, GIF, and PNG. In addition to traditional single images, HEIF supports encoding image sequences, image collections, auxiliary images like alpha or depth maps, live images & video, audio, and HDR for greater contrast. We have heard your feedback that these features and the ability to share photos easily with other platforms is important to you. This release we are delivering essential viewing support (You cannot edit HEIF images in this release).

exekutive:
Nice. iOS, MacOS, Windows and Android P all support it now. Imagemagick recently implemented it as well.

vapier:
this will be available in the next release

PR #670, opened 2021-02-02
HEIF support through libheif

YakoYakoYokuYoku:
With the adoption of AVIF by Firefox and Chromium based browsers (still in experimental phase), the newer incorporation of HEIF by Canon and Sony in their cameras and the newer support of both of them in modern software like ImageMagick, GIMP and Krita, `gd` haven't seen any endorsement for the formats up until this PR.

Reading and writing is done by `libheif`, with functionality for chroma subsampling (for now `4:2:0`, `4:2:2` and `4:4:4`), quality (with new `200` for lossless) and compression (whether `HEVC` or `AV1`) selection. Apart from the new `gdImageCreateFromHeif*`/`gdImageHeif*` there are `gdImageCreateFromAvif*`/`gdImageAvif*` methods for strict AVIF use.

This was tested with `libheif` version `1.11.0` in my Solus machine.

Also, fixes both #395 and #557. Feedback is much appreciated.


src/gd_heif.c
```diff
@@ -0,0 +1,839 @@
...
+	int    width, height;
+	uint8_t   *filedata = NULL;
+	uint8_t    *rgba = NULL;
+	unsigned char   *read, *temp, *magic;
+	size_t size = 0, n;
+	gdImagePtr im;
+	int x, y;
+	uint8_t *p;
+
+	magic = gdMalloc(GD_HEIF_HEADER);
```
vapier, line 131: need to check for NULL return

then again, does this need to be on the heap ?  `GD_HEIF_HEADER` is only 12 bytes.  just use a stack buffer.

src/gd_heif.c
```diff
@@ -0,0 +1,839 @@
...
+    size            - size of HEIF data in bytes.
+    data            - pointer to HEIF data.
+*/
+BGD_DECLARE(gdImagePtr) gdImageCreateFromHeifPtr (int size, void *data)
+{
+	gdImagePtr im;
+	gdIOCtx *in = gdNewDynamicCtxEx(size, data, 0);
+	if (!in)
+		return 0;
+	im = gdImageCreateFromHeifCtx(in);
```
vapier, line 94: need to check for NULL return

src/gd_heif.c
```diff
@@ -0,0 +1,839 @@
...
+	gdFree(magic);
+	gdSeek(infile, 0);
+
+	do {
+		temp = gdRealloc(filedata, size+GD_HEIF_ALLOC_STEP);
+		if (temp) {
+			filedata = temp;
+			read = temp + size;
+		} else {
+			if (filedata) {
```
vapier, line 147: don't need to check it as `free(NULL)` is safe

src/gd_heif.c
```diff
@@ -0,0 +1,839 @@
...
+	if (!_gdHeifCheckBrand(magic, expected_brand)) {
+		gd_error("gd-heif incorrect type of file\n");
+		gdFree(magic);
+		return NULL;
+	}
+	gdFree(magic);
+	gdSeek(infile, 0);
+
+	do {
+		temp = gdRealloc(filedata, size+GD_HEIF_ALLOC_STEP);
```
vapier, line 142: use spaces around binary operators like `+` and `>` and `!=`

this comes up multiple times in this file

src/gd_heif.c
```diff
@@ -0,0 +1,839 @@
...
+			return NULL;
+		}
+
+		n = gdGetBuf(read, GD_HEIF_ALLOC_STEP, infile);
+		if (n>0 && n!=EOF) {
+			size += n;
+		}
+	} while (n>0 && n!=EOF);
+
+	heif_ctx = heif_context_alloc();
```
vapier, line 160: does this need error checking ?

src/gd_heif.c
```diff
@@ -0,0 +1,839 @@
...
+			}
+			gd_error("gd-heif decode realloc failed\n");
+			return NULL;
+		}
+
+		n = gdGetBuf(read, GD_HEIF_ALLOC_STEP, infile);
+		if (n>0 && n!=EOF) {
+			size += n;
+		}
+	} while (n>0 && n!=EOF);
```
vapier, line 158: if this loop runs more than once, the previous `temp = gdMalloc(...)` will leak

src/gd_heif.c
```diff
@@ -0,0 +1,839 @@
...
+		gdFree(temp);
+		heif_image_release(heif_im);
+		heif_image_handle_release(heif_imhandle);
+		heif_context_free(heif_ctx);
+		gdImageDestroy(im);
+		return NULL;
+	}
+	for (y = 0, p = rgba; y < height; y++) {
+		for (x = 0; x < width; x++) {
+			register uint8_t r = *(p++);
```
vapier, line 217: don't use `register` anymore.  it hasn't been useful in decades.

src/gd_heif.c
```diff
@@ -0,0 +1,839 @@
...
+	gdPutBuf(data, size, outfile);
+
+	err.code = heif_error_Ok;
+	err.subcode = heif_suberror_Unspecified;
+	err.message = "";
+
+	return err;
+}
+
+/* returns 0 on success, 1 on failure */
```
vapier, line 335: use `TRUE` & `FALSE` instead

src/gd_heif.c
```diff
@@ -0,0 +1,839 @@
...
+
+	if (overflow2(gdImageSX(im), 4)) {
+		return 1;
+	}
+
+	if (overflow2(gdImageSX(im) * 4, gdImageSY(im))) {
+		return 1;
+	}
+	
+	heif_ctx = heif_context_alloc();
```
vapier, line 369: error checking ?

src/gd_heif.c
```diff
@@ -0,0 +1,839 @@
...
+		gd_error("gd-heif cannot get image plane\n");
+		heif_image_release(heif_im);
+		heif_encoder_release(heif_enc);
+		heif_context_free(heif_ctx);
+		return 1;
+	}
+	p = rgba;
+	for (y = 0; y < gdImageSY(im); y++) {
+		for (x = 0; x < gdImageSX(im); x++) {
+			register int c;
```
vapier, line 427: drop `register` usage

vapier (changes requested):
overall, pretty well written!

src/gd.h
```diff
@@ -627,6 +667,14 @@ BGD_DECLARE(gdImagePtr) gdImageCreateFromWebp (FILE * inFile);
 BGD_DECLARE(gdImagePtr) gdImageCreateFromWebpPtr (int size, void *data);
 BGD_DECLARE(gdImagePtr) gdImageCreateFromWebpCtx (gdIOCtx * infile);
 
+BGD_DECLARE(gdImagePtr) gdImageCreateFromHeif (FILE * inFile);
+BGD_DECLARE(gdImagePtr) gdImageCreateFromHeifPtr (int size, void *data);
+BGD_DECLARE(gdImagePtr) gdImageCreateFromHeifCtx (gdIOCtx * infile);
+
+BGD_DECLARE(gdImagePtr) gdImageCreateFromAvif (FILE * inFile);
```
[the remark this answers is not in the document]
YakoYakoYokuYoku: I included them to have a much simpler handling with AVIFs, e.g.:

```php
$myavif = gdimagecreatefromavif("test.avif");
```

The other case was if the need for an alternative AVIF library raises (probably ain't gonna happen).

src/gd.h
```diff
@@ -333,6 +333,46 @@ typedef enum {
...
+ *
+ *  GD_HEIF_COMPRESSION_UNKNOWN
+ *  GD_HEIF_COMPRESSION_HEVC
+ *  GD_HEIF_COMPRESSION_AV1
+ *
+ * See also:
+ *  - <gdImageHeif>
+ */
+typedef enum {
+	GD_HEIF_COMPRESSION_UNKNOWN = 0,
```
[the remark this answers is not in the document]
YakoYakoYokuYoku: `libheif` calls them _"compression formats"_, but let's move on with _"codecs"_.

src/gd_heif.c
YakoYakoYokuYoku: Yes, as the library does some validation for the HEIF data.

src/gd_heif.c
YakoYakoYokuYoku: Same as above.

morsssss:
I'm glad to see a proposal to include support for HEIF!

As mentioned in https://github.com/libgd/libgd/issues/557, my PR to support AVIF through libavif is almost complete. As promised, I'll share my branch tomorrow.

src/gd_heif.c
vapier: i mean, don't you need to verify `heif_ctx` is not NULL here

src/gd.h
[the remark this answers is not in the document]
vapier: sorry, but from reading the code, i'm not seeing how this is necessary.  it seems like `gdImageCreateFromHeif("test.avif")` works.  am i missing something ?

src/gd_heif.c
```diff
@@ -0,0 +1,839 @@
...
+BGD_DECLARE(void *) gdImageHeifPtrEx (gdImagePtr im, int *size, int quality, gdHeifCodec codec, gdHeifChroma chroma)
+{
+	void *rv;
+	gdIOCtx *out = gdNewDynamicCtx(2048, NULL);
+	if (out == NULL) {
+		return NULL;
+	}
+	if (_gdImageHeifCtx(im, out, quality, codec, chroma)) {
+        rv = gdDPExtractData(out, size);
+    } else {
```
vapier, line 596: seems like indentation is incorrect here

github makes it hard to double check whitespace, so make sure to use tabs everywhere

src/gd.h
```diff
@@ -1084,6 +1132,18 @@ BGD_DECLARE(void *) gdImageWebpPtr (gdImagePtr im, int *size);
 BGD_DECLARE(void *) gdImageWebpPtrEx (gdImagePtr im, int *size, int quantization);
 BGD_DECLARE(void) gdImageWebpCtx (gdImagePtr im, gdIOCtx * outfile, int quantization);
 
+BGD_DECLARE(void) gdImageHeifEx (gdImagePtr im, FILE * outFile, int quality, gdHeifCodec codec, gdHeifChroma chroma);
+BGD_DECLARE(void) gdImageHeif (gdImagePtr im, FILE * outFile);
+BGD_DECLARE(void *) gdImageHeifPtr (gdImagePtr im, int *size);
+BGD_DECLARE(void *) gdImageHeifPtrEx (gdImagePtr im, int *size, int quality, gdHeifCodec codec, gdHeifChroma chroma);
+BGD_DECLARE(void) gdImageHeifCtx (gdImagePtr im, gdIOCtx * outfile, int quality, gdHeifCodec codec, gdHeifChroma chroma);
+
+BGD_DECLARE(void) gdImageAvifEx (gdImagePtr im, FILE * outFile, int quality, gdHeifChroma chroma);
```
vapier, line 1141: these are creating HEIF container files using the AV1 codec right ?  not the AVIF format ?

either way, these seem redundant when you can call gdImageHeif funcs with the codec setting.

src/gd_heif.c
YakoYakoYokuYoku: Yeah, `heif_ctx` might be `NULL` and `heif_context_read_from_memory_without_copy` will derreference it.

src/gd.h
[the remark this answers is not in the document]
vapier: support for libavif is inbound.  it doesn't seem redundant to me as we cannot guarantee which library a system might have or the builder chooses to include.  i don't see why gd would play favorites with one library over another.

so if `gdImageCreateFromHeif` supports AVIF inputs, lets drop these Avif variants and leave them for the libavif backend.  once both merge, we could take another look at having these Avif entry points use libavif if available, and if not, fallback to calling the Heif entry points.  then the user wouldn't have to care and would be able to pick whatever makes the most sense.

src/gd.h
vapier: thanks for the link.  i thought the container formats were diff.

that said, it seems like they're still redundant with the Heif variants and so should be omitted.  see the `gdImageCreateFromAvif` discussion above.

tests/heif/heif_im2im.c
```diff
@@ -0,0 +1,70 @@
...
+{
+	gdImagePtr src, dst;
+	int r, g, b;
+	void *p;
+	int size = 0;
+	int status = 0;
+	CuTestImageResult result = {0, 0};
+
+	src = gdImageCreateTrueColor(100, 100);
+	if (src == NULL) {
```
vapier, line 20: i think these if statements could be collapsed into a single call to `gdTestAssert` or `gdTestAssertMsg`

you'll want to look at the other tests to see if you can collapse them too

src/gd_heif.c
```diff
@@ -0,0 +1,855 @@
...
+#include <stdio.h>
+#include <math.h>
+#include <string.h>
+#include <stdlib.h>
+#include "gd.h"
+#include "gd_errors.h"
+#include "gdhelpers.h"
+
+#ifdef HAVE_LIBHEIF
+#include "libheif/heif.h"
```
vapier, line 21: use `<>` to include system headers like this

src/gd.h
```diff
@@ -627,6 +667,14 @@ BGD_DECLARE(gdImagePtr) gdImageCreateFromWebp (FILE * inFile);
 BGD_DECLARE(gdImagePtr) gdImageCreateFromWebpPtr (int size, void *data);
 BGD_DECLARE(gdImagePtr) gdImageCreateFromWebpCtx (gdIOCtx * infile);
 
+BGD_DECLARE(gdImagePtr) gdImageCreateFromHeif (FILE * inFile);
```
vapier, line 670: i know our existing code is inconsistent, but we should start writing new stuff correctly.  so omit the space before the `(` with function calls and after the `*` with pointers.

it comes up in this header and in the gd_heif.c file a bit.

src/gd_heif.c
```diff
@@ -0,0 +1,855 @@
...
+		n = gdGetBuf(read, GD_HEIF_ALLOC_STEP, infile);
+		if (n > 0) {
+			size += n;
+		}
+	}
+
+	heif_ctx = heif_context_alloc();
+	if (heif_ctx == NULL) {
+		gd_error("gd-heif could not allocate context\n");
+		gdFree(temp);
```
vapier, line 163: should free `filedata`, not `temp`

comes up below too

vapier (changes requested):

src/gd_heif.c
YakoYakoYokuYoku: Taking your tip!

src/gd.h
YakoYakoYokuYoku: Done and thanks for pointing that out.

src/gd.h
YakoYakoYokuYoku: As I've incorporated HEIF support into `gd_filename.c`, we'll require the `gdImageAvif*` methods as `gdImageHeif` will use HEVC instead of AV1.

vapier: you can drop the Avif calls from that file.  i'm not seeing it being required here.

vapier:
you'll want to update `Supported Image Formats` in the top level README.md file to cover HEIF formats & libheif usage

src/gd.h
YakoYakoYokuYoku: Should I write an AVIF reader function inside `gd_filename.c` (or another file) then?

vapier: let's just omit it entirely at this point.  as i noted, we can reconcile the libheif support for .avif once libavif support is merged.

cryptomilk:
I haven't looked at the code, but make sure to check if libheif has been compiled with HEIF support enabled. libheif wont be compiled with HEIF support by any Linux distribution ...

You will only get HEIF by installing libheif from 3rd party repositories ...

vapier:
can that be done without executing code ?  we don't like configure tests that try to execute to detect features.

the API is the same right ?  it just returns an error at runtime ?  that's prob the best we can hope for ... just error out saying HEIF support is disabled in libheif.

YakoYakoYokuYoku:
> I haven't looked at the code, but make sure to check if libheif has been compiled with HEIF support enabled. libheif wont be compiled with HEIF support by any Linux distribution ...
>
> You will only get HEIF by installing libheif from 3rd party repositories ...

[Solus](https://dev.getsol.us/source/libheif/browse/master/package.yml;945c95a6f367d55cfd2a32d43a6ef0ac7721ac8b), [Debian](https://salsa.debian.org/multimedia-team/libheif/-/blob/debian/1.11.0-1/debian/control) and [Archlinux](https://github.com/archlinux/svntogit-packages/commit/b3b1b3fb8c6d7e2d598782cc08a6115b66da8e9c#diff-3e341d2d9c67be01819b25b25d5e53ea3cdf3a38d28846cda85a195eb9b7203a) bundles the encoders and decoders for HEIF (and for AVIF too), so it _might_ be rare if any other Linux distro decides not to include them.

> the API is the same right ? it just returns an error at runtime ? that's prob the best we can hope for ... just error out saying HEIF support is disabled in libheif.

libheif requests the encoder with the `heif_context_get_encoder_for_format` and it returns an (not yet appropriate) error if it's not found with my API. While decoding libheif tries to get the best decoder when `heif_decode_image` is called and same situation.

> let's just omit it entirely at this point. as i noted, we can reconcile the libheif support for .avif once libavif support is merged.

As for AVIF support I'll trim it off for another PR (but reading _could_ still be done by `gdCreateFromHeif` and friends), but idek when libavif will stabilize.

src/gd_heif.c
```diff
@@ -0,0 +1,855 @@
...
+	}
+
+	heif_ctx = heif_context_alloc();
+	if (heif_ctx == NULL) {
+		gd_error("gd-heif could not allocate context\n");
+		return NULL;
+	}
+	err = heif_context_get_encoder_for_format(heif_ctx, codec, &heif_enc);
+	if (err.code != heif_error_Ok) {
+		gd_error("gd-heif image creation failed\n");
```
vapier, line 388: since this can fail if the codec is not supported in the host env, how about tweaking the message like:

 gd_error("gd-heif image creation failed (missing codec support?)\n");

src/gd_heif.c
```diff
@@ -0,0 +1,855 @@
...
+
+/*
+  Function: gdImageHeifCtx
+
+    Write the image as HEIF data via a <gdIOCtx>. See <gdImageHeifEx>
+    for more details.
+
+  Variants:
+
+    <gdImageAvifCtx> AVIF method counterpart.
```
vapier, line 485: copy & paste typo

vapier:
> libheif requests the encoder with the heif_context_get_encoder_for_format and it returns an (not yet appropriate) error if it's not found with my API. While decoding libheif tries to get the best decoder when heif_decode_image is called and same situation.

sounds fine then (and looks like you already handle that).  it's the system configurator's problem if they install a libheif that is missing heif support and then tries to use heif gd funcs.  we can't (and shouldn't) try to detect all the misconfigurations in a distro.

README.md
```diff
@@ -41,6 +41,8 @@ It also has optional support for more formats via external libraries:
 
 * [JPEG](https://en.wikipedia.org/wiki/JPEG) via [IJG/libjpeg](http://www.ijg.org/) or [libjpeg-turbo](http://libjpeg-turbo.virtualgl.org/)
   * Does not include [JPEG 2000](https://en.wikipedia.org/wiki/JPEG_2000)
+* [HEIF](https://en.wikipedia.org/wiki/High_Efficiency_Image_File_Format) via [libheif](https://github.com/strukturag/libheif/)
```
vapier, line 44: "H" before "J" :)

configure.ac
```diff
@@ -285,6 +285,14 @@ GD_LIB_PKG_CHECK([LIBWEBP], [WEBP], [webp], [libwebp], [
   ])
 ])
 
+dnl Check for heif support.
+GD_LIB_PKG_CHECK([LIBHEIF], [HEIF], [heif], [libheif], [
```
vapier, line 289: if there's a specific min version you want to enforce, you'd want to write `[libheif >= x.y.z]` here

src/gd_heif.c
```diff
@@ -0,0 +1,585 @@
...
+BGD_DECLARE(gdImagePtr) gdImageCreateFromHeif(FILE *inFile)
+{
+	gdImagePtr im;
+	gdIOCtx *in = gdNewFileCtx(inFile);
+
+	if (!in)
+		return NULL;
+	im = gdImageCreateFromHeifCtx(in);
+	in->gd_free(in);
+	if (!im)
```
vapier, line 70: you can omit this `if` statement.  returning `im` below will be the same.

src/gd_heif.c
```diff
@@ -0,0 +1,585 @@
...
+BGD_DECLARE(gdImagePtr) gdImageCreateFromHeifPtr(int size, void *data)
+{
+	gdImagePtr im;
+	gdIOCtx *in = gdNewDynamicCtxEx(size, data, 0);
+
+	if (!in)
+		return NULL;
+	im = gdImageCreateFromHeifCtx(in);
+	in->gd_free(in);
+	if (!im)
```
vapier, line 95: you can omit this `if` statement.  returning `im` below will be the same.

src/gd_heif.c
```diff
@@ -0,0 +1,585 @@
...
+	int    width, height;
+	uint8_t   *filedata = NULL;
+	uint8_t    *rgba = NULL;
+	unsigned char   *read, *temp, magic[GD_HEIF_HEADER];
+	size_t size = 0, n = GD_HEIF_ALLOC_STEP;
+	gdImagePtr im;
+	int x, y;
+	uint8_t *p;
+
+	gdGetBuf(magic, GD_HEIF_HEADER, infile);
```
vapier, line 132: should check the return value so you don't read uninitialized `magic` data

vapier (commented):
can you add yourself to CONTRIBUTORS too ?

configure.ac
YakoYakoYokuYoku: I'll target version [`1.7.0`](https://github.com/strukturag/libheif/releases/tag/v1.7.0) because it is the release that brought AVIF support (and the encoder options).

README.md
YakoYakoYokuYoku: Completely overlooked that, but it's :+1: now.

YakoYakoYokuYoku:
> can you add yourself to CONTRIBUTORS too ?

Of course!

tests/heif/heif_read.c
```diff
@@ -0,0 +1,23 @@
...
+ */
+
+
+#include "gd.h"
+#include "gdtest.h"
+
+int main()
+{
+	gdImagePtr im;
+    FILE *fp;
```
vapier, line 13: looks like the file has mixed indentation.  you want tabs everywhere.

vapier (commented):
one last tweak and we can merge this

tests/heif/heif_read.c
YakoYakoYokuYoku: Yeah, seems that both me and VSCode didn't caught that.

vapier (approved):

vapier:
thanks for sticking through my annoying feedback.  i've pushed this to the 2.3 branch too so it'll be in the next release and not just languish in master for a while.

YakoYakoYokuYoku:
> thanks for sticking through my annoying feedback.
No problem whatsoever.
> i've pushed this to the 2.3 branch too so it'll be in the next release and not just languish in master for a while.
I'd probably enable the HEIF tests before the next release.

Also thanks again for the reviews. We can take a look again to the AVIF API in the future.

PR #672, opened 2021-02-21
cmake: fix finding libwebp when cross compiling

neheb:
libwebp already contains .cmake files for this. Just use them.

This requires changing the case of the variables to match.

Signed-off-by: Rosen Penev <rosenp@gmail.com>


vapier (changes requested):
unfortunately, it seems libwebp only installs the cmake module when you use cmake to build it, and no distro is doing that afaict (Gentoo/Ubuntu/Debian/Fedora).  which means this is going to fail to build on all of them as you can already see in the Travis run there.

i file https://crbug.com/webp/509 to fix their autotools install.  but until that is merged and distros pick it up, we can't rely on the system providing it.

if you wanted to delete our custom webp cmake module and copy the upstream one into our tree and use it, that'd be fine.

neheb:
OK I'll keep it local.

PR #680, opened 2021-03-09
Fix for libavif v0.8.2

### LINKED ISSUE #677: Test failure avif/compare_avif_to_png with libavif-0.8.2
```
GD Warning: avif error - Could not parse image: Truncated dataavif/compare_avif_to_png.c:52: baboon: gdImageCreateFromAvifPtr failed
GD Warning: avif error - Could not parse image: Truncated dataavif/compare_avif_to_png.c:52: dice_with_alpha: gdImageCreateFromAvifPtr failed
GD Warning: avif error - Could not parse image: Truncated dataavif/compare_avif_to_png.c:52: plum_blossom_12bit: gdImageCreateFromAvifPtr failed
GD Warning: avif error - Could not parse image: Truncated dataavif/compare_avif_to_png.c:52: sunset: gdImageCreateFromAvifPtr failed
FAIL avif/compare_avif_to_png (exit status: 4)

```

morsssss:
Don't return AVIF_RESULT_TRUNCATED_DATA, as this is normal for libavif <= 0.8.2. In our tests, this makes tests pass with libavif 0.8.2.

Plus, we did a few things to stop compiler warnings - and added a newline to error output.

thanks @wantehchang for the collaboration here!

This fixes #677 .

### DIFF
--- a/src/gd_avif.c
+++ b/src/gd_avif.c
@@ -165,9 +165,6 @@ static avifBool isAvifError(avifResult result, const char *msg) {
 
 	We ignore readFlags, just as the avifIO*ReaderRead() functions do.
 
-	In libavif 0.8.2, it's normal to get 0 bytes on the last read.
-	Thus we don't complain if charsRead < size.
-
 	If there's a problem, this returns an avifResult error.
 	If things go well, return AVIF_RESULT_OK.
 	Of course these AVIF codes shouldn't be returned by any top-level GD function.
@@ -176,6 +173,11 @@ static avifResult readFromCtx(avifIO *io, uint32_t readFlags, uint64_t offset, s
 {
 	void *dataBuf = NULL;
 	gdIOCtx *ctx = (gdIOCtx *) io->data;
+
+	// readFlags is unsupported
+	if (readFlags != 0) {
+		return AVIF_RESULT_IO_ERROR;
+	}
 
 	// TODO: if we set sizeHint, this will be more efficient.
 
--- a/src/gd_avif.c
+++ b/src/gd_avif.c
@@ -148,7 +148,7 @@ static avifBool isAvifSrgbImage(avifImage *avifIm) {
 */
 static avifBool isAvifError(avifResult result, const char *msg) {
 	if (result != AVIF_RESULT_OK) {
-		gd_error("avif error - %s: %s", msg, avifResultToString(result));
+		gd_error("avif error - %s: %s\n", msg, avifResultToString(result));
 		return AVIF_TRUE;
 	}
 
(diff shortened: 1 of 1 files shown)


src/gd_avif.c
```diff
@@ -165,6 +165,9 @@ static avifBool isAvifError(avifResult result, const char *msg) {
 
 	We ignore readFlags, just as the avifIO*ReaderRead() functions do.
 
+	In libavif 0.8.2, it's normal to get 0 bytes on the last read.
+	Thus we don't complain if charsRead < size.
```
wantehchang, line 169: We don't need to add this comment because this is the specified behavior. (Also, this comment is overly specific.)

Here is the specification from avif.h:

```
// This function should return a block of memory that *must* remain valid until another read call to
// this avifIO struct is made (reusing a read buffer is acceptable/expected).
//
// * If offset exceeds the size of the content (past EOF), return AVIF_RESULT_IO_ERROR.
// * If offset is *exactly* at EOF, provide a 0-byte buffer and return AVIF_RESULT_OK.
// * If (offset+size) exceeds the contents' size, it must truncate the range to provide all
//   bytes from the offset to EOF.
// * If the range is unavailable yet (due to network conditions or any other reason),
//   return AVIF_RESULT_WAITING_ON_IO.
// * Otherwise, provide the range and return AVIF_RESULT_OK.
typedef avifResult (*avifIOReadFunc)(struct avifIO * io, uint32_t readFlags, uint64_t offset, size_t size, avifROData * out);
```
Note the third and the fifth bullet points. The truncation of the requested range mentioned in the third bullet point is done by the ctx->getBuf() call.

src/gd_avif.c
```diff
@@ -165,6 +165,9 @@ static avifBool isAvifError(avifResult result, const char *msg) {
 
 	We ignore readFlags, just as the avifIO*ReaderRead() functions do.
 
+	In libavif 0.8.2, it's normal to get 0 bytes on the last read.
```
wantehchang, line 168: The comment at line 166 is incorrect. The avifIO*ReaderRead() functions require readFlags to be zero:

```
    if (readFlags != 0) {
        // Unsupported readFlags
        return AVIF_RESULT_IO_ERROR;
    }
```

src/gd_avif.c
```diff
@@ -477,7 +480,7 @@ static avifBool _gdImageAvifCtx(gdImagePtr im, gdIOCtx *outfile, int quality, in
 
 	uint32_t val;
 	uint8_t *p;
-	int x, y;
+	uint32_t x, y;
```
wantehchang, line 483: Note that tests/avif/bad_input.c has an unused variable `retval`.

wantehchang (changes requested):

src/gd_avif.c
morsssss: I see your point. I thought of `readFlags` as simply a thing to ignore, but I see that the functions in `libavif/io.c` consistently throw an error. I should do the same.

src/gd_avif.c
morsssss: good catch. I'll deal with that in a separate PR.

src/gd_avif.c
wantehchang: I see. You can quote the specification of avifIOReadFunc as comments in this function.

For example, put the third bullet point as a comment before the ctx->getBuf() call, and put the fourth and fifth bullet points as a comment before the final return statement.

remicollet:
OK, tested with 0.8.2, and test suite pases

remicollet (approved):

PR #683, opened 2021-03-11
Set default AVIF encoding speed to 6

morsssss:
Turns out that, in many cases, AVIF's default encoding speed (`AVIF_SPEED_DEFAULT`) would be 0, the slowest possible speed.

The team's been working on making speed 6 a great compromise. We should make this our default as well.

/cc @wantehchang

### DIFF
--- a/src/gd_avif.c
+++ b/src/gd_avif.c
@@ -33,7 +33,10 @@
 			We need more testing to really know what quantizer settings are optimal,
 			but teams at Google have been using maximum=30 as a starting point.
 		QUALITY_DEFAULT: following gd conventions, -1 indicates the default.
-		SPEED_DEFAULT: AVIF_SPEED_DEFAULT is -1. This simply tells the AVIF encoder to use the default speed.
+		SPEED_DEFAULT:
+		  AVIF_SPEED_DEFAULT is simply the default encoding speed of the AV1 codec.
+			This could be as slow as 0. So we use 6, which is currently considered to be a fine default.
+		
 */
 
 #define CHROMA_SUBSAMPLING_DEFAULT AVIF_PIXEL_FORMAT_YUV420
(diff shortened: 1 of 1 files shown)


src/gd_avif.c
```diff
@@ -497,7 +500,7 @@ static avifBool _gdImageAvifCtx(gdImagePtr im, gdIOCtx *outfile, int quality, in
 		return 1;
 	}
 
-	if (speed != AVIF_SPEED_DEFAULT)
+	if (speed != SPEED_DEFAULT)
```
wantehchang, line 503: Note that AVIF_SPEED_DEFAULT is defined as an invalid value of -1:

`#define AVIF_SPEED_DEFAULT -1`

If you want to allow the callers of _gdImageAvifCtx() to pass speed=-1 to indicate that they want your default speed, then the code here should read:

```
    if (speed == -1)
        speed = SPEED_DEFAULT;
    else
        speed = CLAMP(speed, AVIF_SPEED_SLOWEST, AVIF_SPEED_FASTEST);
```

Otherwise, we should get rid of the if and just say:

`        speed = CLAMP(speed, AVIF_SPEED_SLOWEST, AVIF_SPEED_FASTEST);`

morsssss: That's funny. I was going to include a comment about this, but I didn't :)

I added that conditional originally around that `CLAMP` originally because, indeed, `AVIF_SPEED_DEFAULT` was `-1`, and we didn't want to clamp that value up to `0`. I left it there because I thought it might still be useful for just this reason - just in case someone, someday, changes the default back to `AVIF_SPEED_DEFAULT`.

Would you prefer that I remove the conditional and add this as a comment instead?

I don't think callers will try `-1` because we're not going to document that that's a default value.

morsssss: I'd personally add a comment in case someone, someday, wants to set the speed back to AVIF_SPEED_DEFAULT, which will probably still be -1.

If you feel strongly that I shouldn't, I could always add this comment in the commit instead :)

wantehchang: Hi Ben,

If you want to prepare for that possibility, it is best to revert to the original code:

```
	if (speed != AVIF_SPEED_DEFAULT)
		speed = CLAMP(speed, AVIF_SPEED_SLOWEST, AVIF_SPEED_FASTEST);
```

morsssss: That would be appealing - except that it would allow users to pass speed=-1. Looking at the PHP interface, I reconsidered this idea - but I realized it would be confusing, since end users will already have a default speed of 6. A second default would be undesirable.

I ended up removing this conditional and adding a comment about this instead.

vapier (approved):

Issue #678, opened 2021-03-08
[2.3.2] segfault in heif tests


remicollet:
```
../config/test-driver: line 107: 2351333 Aborted                 (core dumped) "$@" > $log_file 2>&1
FAIL: heif/heif_im2im

../config/test-driver: line 107: 2351482 Segmentation fault      (core dumped) "$@" > $log_file 2>&1
FAIL: heif/heif_read

```

Backtrace for heif_im2im:

```
#0  0x00007f7d2e1169d5 in raise () from /lib64/libc.so.6
#1  0x00007f7d2e0ff8a4 in abort () from /lib64/libc.so.6
#2  0x00007f7d2e159177 in __libc_message () from /lib64/libc.so.6
#3  0x00007f7d2e160e6c in malloc_printerr () from /lib64/libc.so.6
#4  0x00007f7d2e16278e in _int_free () from /lib64/libc.so.6
#5  0x00007f7d2e3509de in std::_Sp_counted_ptr_inplace<heif::HeifPixelImage, std::allocator<heif::HeifPixelImage>, (__gnu_cxx::_Lock_policy)2>::_M_dispose() ()
   from /lib64/libheif.so.1
#6  0x00007f7d2e375705 in heif::convert_colorspace(std::shared_ptr<heif::HeifPixelImage> const&, heif_colorspace, heif_chroma, std::shared_ptr<heif::color_profile_nclx const>, int) () from /lib64/libheif.so.1
#7  0x00007f7d2e3677f7 in heif::HeifContext::Image::encode_image_as_hevc(std::shared_ptr<heif::HeifPixelImage>, heif_encoder*, heif_encoding_options const*, heif_image_input_class) () from /lib64/libheif.so.1
#8  0x00007f7d2e37d6f3 in heif::HeifContext::encode_image(std::shared_ptr<heif::HeifPixelImage>, heif_encoder*, heif_encoding_options const*, heif_image_input_class, std::shared_ptr<heif::HeifContext::Image>&) [clone .constprop.0] () from /lib64/libheif.so.1
#9  0x00007f7d2e35e288 in heif_context_encode_image () from /lib64/libheif.so.1
#10 0x00007f7d2e98b0a1 in _gdImageHeifCtx (im=im@entry=0x872750, outfile=outfile@entry=0x87e9b0, quality=quality@entry=200, codec=codec@entry=GD_HEIF_CODEC_HEVC, 
    chroma=chroma@entry=0x403120 "444") at gd_heif.c:375
[...]

PR #685, opened 2021-03-16
raise libheif minimal version to 1.9.0

remicollet:
See #678

Segfault with 1.8.0
Tested, OK with 1.9.1, 1.10.0 and 1.11.0

ping @YakoYakoYokuYoku

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -286,7 +286,7 @@ GD_LIB_PKG_CHECK([LIBWEBP], [WEBP], [webp], [libwebp], [
 ])
 
 dnl Check for heif support.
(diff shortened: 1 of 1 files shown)


remicollet:
Of course a better fix will be to have a patch fixing build with older libheif.
Especially as version 1.9.0 is quite recent (Sept 2020)

vapier:
iiuc, we don't need this anymore.  feel free to reopen if i'm mistaken.

remicollet:
@YakoYakoYokuYoku OK for me, build and test suite OK with 1.8.0

PR #686, opened 2021-03-16
Permit compilation with libheif version 1.7.0

YakoYakoYokuYoku:
`libheif` versions that came before `1.9.0` don't support changing the output image chroma. I did not notice that and it resulted with tests failures across other OSes that don't have a much newer `libheif`.

See #678. Supersedes #685.

### DIFF
--- a/src/gd_heif.c
+++ b/src/gd_heif.c
@@ -321,6 +321,7 @@ static int _gdImageHeifCtx(gdImagePtr im, gdIOCtx *outfile, int quality, gdHeifC
 		return GD_FALSE;
 	}
 
+#if LIBHEIF_HAVE_VERSION(1, 9, 0)
 	err = heif_encoder_set_parameter_string(heif_enc, "chroma", chroma);
 	if (err.code != heif_error_Ok) {
 		gd_error("gd-heif invalid chroma subsampling parameter\n");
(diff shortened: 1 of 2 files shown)


src/gd_heif.c
```diff
@@ -321,13 +321,17 @@ static int _gdImageHeifCtx(gdImagePtr im, gdIOCtx *outfile, int quality, gdHeifC
 		return GD_FALSE;
 	}
 
+#if LIBHEIF_HAVE_VERSION(1, 9, 0)
```
vapier, line 324: if the api version is available at runtime (and it seems so with your test changes), then lets do that here too.

```c
if (LIBHEIF_HAVE_VERSION(1, 9, 0) || heif_get_version_number() >= LIBHEIF_MAKE_VERSION(1, 9, 0)) {
```

YakoYakoYokuYoku: Done.

src/gd_heif.c
```diff
@@ -321,12 +321,16 @@ static int _gdImageHeifCtx(gdImagePtr im, gdIOCtx *outfile, int quality, gdHeifC
...
+	if (heif_get_version_number_major() >= 1 && heif_get_version_number_minor() >= 9) {
+		err = heif_encoder_set_parameter_string(heif_enc, "chroma", chroma);
+		if (err.code != heif_error_Ok) {
+			gd_error("gd-heif invalid chroma subsampling parameter\n");
+			heif_encoder_release(heif_enc);
+			heif_context_free(heif_ctx);
+			return GD_FALSE;
+		}
+	} else {
+		(void)chroma;
```
vapier, line 333: you don't need this anymore ...

if you were just trying to avoid unused compiler warnings, by avoiding cpp above, you've done that now too

vapier:
thx!

Issue #655, opened 2020-09-05
gdimagestringft/gdimagestringft_bbox test fails due to image being slightly off


kloczek:
Up to the commit 9ed64276
```
PASS: bmp/bmp_null
PASS: bmp/bug00275
PASS: bmp/bug00450
PASS: bmp/bmp_im2im
PASS: bmp/bug00276
PASS: fontconfig/basic
PASS: freetype/bug00132
PASS: gd/gd_versiontest
PASS: gd/bug00383
PASS: gd/gd_null
PASS: gd/gd_num_colors
PASS: gd/gd_im2im
PASS: gd2/bug_289
PASS: gd2/bug00209
PASS: gd2/bug00309
PASS: gd2/bug00354
PASS: gd2/bug00383
PASS: gd2/createimagefromgd2part
PASS: gd2/createimagefromgd2partptr
PASS: gd2/gd2_empty_file
PASS: gd2/php_bug_72339
PASS: gd2/gd2_null
PASS: gd2/gd2_read_corrupt
PASS: gd2/too_few_image_data
PASS: gd2/gd2_im2im
PASS: gd2/gd2_read
PASS: gdimagearc/bug00079
PASS: gdimagebrightness/basic
PASS: gdimageclone/bug00300
PASS: gdimageclone/style
PASS: gdimagecolor/basic
PASS: gdimagecolorclosest/gdimagecolorclosest
PASS: gdimagecolorclosesthwb/gdImageColorClosestHWB
PASS: gdimagecolordeallocate/gdimagecolordeallocate
PASS: gdimagecolorexact/gdimagecolorexact
PASS: gdimagecolormatch/cve_2019_6977
PASS: gdimagecolormatch/gdimagecolormatch
PASS: gdimagecolorreplace/gdimagecolorreplace
PASS: gdimagecolorresolve/gdimagecolorresolve
PASS: gdimagecolortransparent/gdimagecolortransparent
PASS: gdimagecompare/gdimagecompare
PASS: gdimagecontrast/basic
PASS: gdimageconvolution/bug00369
PASS: gdimageconvolution/basic
PASS: gdimagecopy/bug00007
PASS: gdimagecopy/bug00081
PASS: gdimagecopymerge/gdimagecopymerge
PASS: gdimagecopymergegray/gdimagecopymergegray
PASS: gdimagecopyresampled/exact_alpha
PASS: gdimagecopyresampled/basic
PASS: gdimagecopyresampled/basic_alpha
[...]

PR #689, opened 2021-03-17
fix test when libraqm is used

remicollet:
The difference is really minor (1 pixel)

Another way is to always allow 1 pixel difference.

### DIFF
--- a/tests/gdimagestringft/gdimagestringft_bbox.c
+++ b/tests/gdimagestringft/gdimagestringft_bbox.c
@@ -1,5 +1,4 @@
 #include "gd.h"
-#include "config.h"
 #include <stdio.h>
 #include <stdlib.h>
 #include <math.h>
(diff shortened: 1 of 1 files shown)


vapier:
this fixes #655 right ?

i'd be amenable to having the `if (brect[j] != EXPECT[i][j])` check accept a one-pixel diff everywhere.  wdyt ?

```
// Allow a little fuzz diff between the reference and the active font renderer.
if (brect[j] < EXPECT[i][j] - 1 || brect[j] > EXPECT[i][j] + 1) {
```

remicollet:
> i'd be amenable to having the if (brect[j] != EXPECT[i][j]) check accept a one-pixel diff everywhere. wdyt ?

Yes, if was my "Another way" proposal

> this fixes #655 right ?

Yes

remicollet:
Simpler way pushed

remicollet:
Merged

PR #692, opened 2021-04-08
Application-supplied memory allocators

### LINKED ISSUE #335: Application-supplied memory allocators
I think we should provide a way for libgd clients to use their own memory allocators. That would be very useful for PHP at least, where there's currently a dichotomy between the bundled and external libgd, because the bundled libgd defines macros to [replace gdMalloc() and friends](https://github.com/php/php-src/blob/PHP-7.0.12/ext/gd/libgd/gdhelpers.h#L11-L22) with Zend memory manager functions, but obviously that doesn't work for system libgd, so PHP's [memory_limit](http://php.net/manual/en/ini.core.php#ini.memory-limit) is not obeyed in the latter case.

I haven't though about the API yet – suggestions are welcome.

YakoYakoYokuYoku:
Because of the usage of the Zend memory manager in the PHP interpreter, `gd` has a situation with having many options for memory management and no control over it.

This PR brings an API for the management that corresponds to the following methods:

``` c
void gdSetMemoryCallocMethod(gdCallocf callocf);
void gdSetMemoryMallocMethod(gdMallocf mallocf);
void gdSetMemoryReallocMethod(gdReallocf reallocf);
void gdSetMemoryFreeMethod(gdFreef freef);
void gdSetMemoryAllocationMethods(gdCallocf callocf, gdMallocf mallocf, gdReallocf reallocf, gdFreef freef);
```

Each one of these sets an specific allocator for the task and there is a much more general method that sets the allocation functions at once.
The underlying magic is setting static variables for every method.
All of `gdMalloc` and its friends had to be modified to use the variables.

Marked as draft because discussions may take place first. Fixes #335.


cmb69:
Thank you! Looks generally good to me, but maybe it's better to split the CMake/pthread changes into a separate PR?

YakoYakoYokuYoku:
You welcome.

Pinging @vapier for review.

YakoYakoYokuYoku:
Maybe `reallocf` is not a [good name](https://linux.die.net/man/3/reallocf).

cmb69:
It may make sense to use the same terms: gdSetMemoryCallocMethod vs. gdCallocf. Why not gdCallocMethod? Or gdSetMemoryCallocFunc and gdCallocFunc.

docs/naturaldocs/project/Menu.txt
```diff
@@ -79,7 +79,7 @@ File: gd_io_dp.c  (gd_io_dp.c)
 File: gd_ss.c  (gd_ss.c)
 File: gd_version.c  (gd_version.c)
 File: gdColorMapLookup  (gd_color_map.c)
-File: gdFree  (gdhelpers.c)
+File: gdhelpers.c  (gdhelpers.c)
```
vapier, line 82: erm, something has gone wrong here ?

src/gd.h
```diff
@@ -1678,6 +1678,35 @@ BGD_DECLARE(int) gdTransformAffineBoundingBox(gdRectPtr src, const double affine
...
+typedef void *(*gdCallocMethod)(size_t nmemb, size_t size);
+typedef void *(*gdMallocMethod)(size_t size);
+typedef void *(*gdReallocMethod)(void *ptr, size_t size);
+typedef void (*gdFreeMethod)(void *ptr);
+
+BGD_DECLARE(void) gdSetMemoryCallocMethod(gdCallocMethod calloc_method);
+BGD_DECLARE(void) gdSetMemoryMallocMethod(gdMallocMethod malloc_method);
+BGD_DECLARE(void) gdSetMemoryReallocMethod(gdReallocMethod realloc_method);
+BGD_DECLARE(void) gdSetMemoryFreeMethod(gdFreeMethod free_method);
+BGD_DECLARE(void) gdSetMemoryAllocationMethods(gdCallocMethod calloc_method, gdMallocMethod malloc_method, gdReallocMethod realloc_method, gdFreeMethod free_method);
```
vapier, line 1709: drop this -- see the bug for info

src/gd.h
```diff
@@ -1678,6 +1678,35 @@ BGD_DECLARE(int) gdTransformAffineBoundingBox(gdRectPtr src, const double affine
...
+    typedef: gdFreeMethod
+
+    Memory release method function pointer type.
+*/
+typedef void *(*gdCallocMethod)(size_t nmemb, size_t size);
+typedef void *(*gdMallocMethod)(size_t size);
+typedef void *(*gdReallocMethod)(void *ptr, size_t size);
+typedef void (*gdFreeMethod)(void *ptr);
+
+BGD_DECLARE(void) gdSetMemoryCallocMethod(gdCallocMethod calloc_method);
```
vapier, line 1705: if we look at `ErrorMethod` as the existing pattern to replicate, we'll want gdClear{Calloc,Malloc,Realloc,Free}Method's too.  especially since we aren't accepting `NULL` as a proxy for this.

tests/gdmem/malloc.c
```diff
@@ -0,0 +1,77 @@
+#include "gd.h"
```
vapier, line 1: add a header block explaining the purpose of this test

tests/gdmem/malloc.c
```diff
@@ -0,0 +1,77 @@
...
+#include "gdtest.h"
+
+void *testcalloc(size_t nmemb, size_t size)
```
vapier, line 4: should mark all these funcs as static

tests/gdmem/malloc.c
```diff
@@ -0,0 +1,77 @@
...
+{
+	char *ptr;
+
+	ptr = calloc(nmemb, size);
+
+	for (size_t i = 0; i < size; i++)
+		*(ptr + i) = 0;
```
vapier, line 11: `calloc()` is documented as doing this, so why do we need to ?

tests/gdmem/malloc.c
```diff
@@ -0,0 +1,77 @@
...
+}
+
+void *testmalloc(size_t size)
+{
+	char *ptr;
+
+	ptr = malloc(size);
+
+	for (size_t i = 0; i < size; i++)
+		*(ptr + i) = 0;
```
vapier, line 23: `malloc()` returns uninitialized memory, not zero-ed memory.  if you're trying to poison things, maybe use 0xaa or something ?

tests/gdmem/malloc.c
```diff
@@ -0,0 +1,77 @@
...
+
+	return (void *)ptr;
+}
+
+void *testrealloc(void *ptr, size_t size)
+{
+	char *ret;
+
+	ret = realloc(ptr, size);
```
vapier, line 32: you can one-line this func

tests/gdmem/malloc.c
```diff
@@ -0,0 +1,77 @@
...
+
+	return (void *)ret;
+}
+
+void testfree(void *ptr)
+{
+	if (ptr != NULL)
```
vapier, line 39: no need for this `if`.  `free(NULL)` is valid.

tests/gdmem/malloc.c
```diff
@@ -0,0 +1,77 @@
...
+	gdSetMemoryMallocMethod(testmalloc);
+	gdSetMemoryReallocMethod(testrealloc);
+	gdSetMemoryFreeMethod(testfree);
+	fseek(fp, 0, SEEK_SET);
+	im2 = gdImageCreateFromPng(fp);
+	if (!gdTestAssert(im2 != NULL))
+		goto door2;
+
+	gdTestImageDiff(im1, im2, NULL, &result);
+	gdTestAssertMsg(result.pixels_changed == 0, "pixels changed: %d\n", result.pixels_changed);
```
vapier, line 67: this doesn't really test that the allocators worked, just that the methods above didn't crash.  maybe a better test would be to have each test func above increment a global variable and then we assert that ?

```
static int free_called = 0;
void testfree(void *ptr)
{
    free_called++;
    free(ptr);
}
...
	im1 = gdImageCreateFromPng(fp);
	if (!gdTestAssert(im1 != NULL))
		goto door1;

        ... assert that free_called == 0 and all other funcs too ...

	gdSetMemoryCallocMethod(testcalloc);
	gdSetMemoryMallocMethod(testmalloc);
	gdSetMemoryReallocMethod(testrealloc);
	gdSetMemoryFreeMethod(testfree);
...
        ... assert that free_called != 0 and all other funcs too ...
```

vapier (changes requested):

docs/naturaldocs/project/Menu.txt
YakoYakoYokuYoku: Alphabetical order ig.

src/gd.h
YakoYakoYokuYoku: Yeah, I also was unsure about this.

tests/gdmem/malloc.c
YakoYakoYokuYoku: It was done for my tests, but I'd rather go to a much proper check.

tests/gdmem/malloc.c
YakoYakoYokuYoku: Done much like this, but instead of counting the calls I went with only checking if it was called.

vapier (approved):
lot of boilerplate, but not sure we can do better :/.  splattering a lot of defines feels like it might be fewer LOC, but harder to understand.

YakoYakoYokuYoku:
Running the test with CMake works just fine but not with Autotools 🤔 .

pierrejoye:
Very good PR , thank you! :)

It could be a tat bit better using:

common struct to store the f

- unction ptr,
- a global struct data, surrounded by our lock mechanism
- must be only called once per process (not sure we can ensure it but then doc issue, big mess if it changes accross the same process)

What do you think?

YakoYakoYokuYoku:
From what was discussed in the issue we can go either ways with the static struct or the vars, but I'll stick with the latter for the moment.
Speaking about locking I find it kinda pointless, the methods are obviously not thread safe and will cause disastrous alterations with the memory managers if handled incorrectly.
Maybe doing checks on another global variable is an option. Or maybe we could not checking anything and document that calling twice is invalid usage (kinda like the `vkBeginCommandBuffer` [docs](https://www.khronos.org/registry/vulkan/specs/1.2-extensions/man/html/vkCreateInstance.html)).

pierrejoye:
I think it would make sense to have a struct, it will also helps CPU cache, especially in loaders where quite a few alloc/free may happen. Or in GD3 with the Paths APIs.

Taste thing, I find it cleaner as well :)

What do you think?

pierrejoye:
Besides that, I am all for it and sounds like a good addition for 2.4.0 :) Well done!!

YakoYakoYokuYoku:
Only thing about this PR is that testing `gdmem` with Autotools and shared `libgd` leads to the functions not being used in the test. Although it works with an static archive. From there idek what's happening.

cmb69:
> I think it would make sense to have a struct, it will also helps CPU cache, especially in loaders where quite a few alloc/free may happen. Or in GD3 with the Paths APIs.

I don't have a strong opinion on this, but sounds reasonable.

Most important for me is that we ship custom allocator support ASAP. I wouldn't want to unbundle libgd from php-src before this isn't available.

> Only thing about this PR is that testing gdmem with Autotools and shared libgd leads to the functions not being used in the test. Although it works with an static archive.

Thanks for reporting. I think this needs closer investigation.

Anyhow, if you have some time, could you please resolve the merge conflicts?

YakoYakoYokuYoku:
No further comments @pierrejoye, @cmb69, @vapier?

pierrejoye:
@YakoYakoYokuYoku in principle I am good with it. I was waiting @remicollet feedback on the API.

While I still like to provide the atomicity handling on our side tho'. That can be done post merge.

remicollet:
> I was waiting @remicollet feedback on the API.

I don't really care about the API ;)   (as soon as it is there)

BTW, a single setter may be enough (I don't think changing only some make sense, and a single call with 4 pointers will avoid forgetting one).

Perhaps a safety bell checking those allocators have not be used yet before allowing the switch...

YakoYakoYokuYoku:
@cmb69, @pierre, @remicollet, @vapier I'm deeply sorry for being late with this PR, was occupied with things IRL and I didn't know how to fix this PR up until now.

So if anyone of you can merge this I'd be glad for it. :pray_tone1:

YakoYakoYokuYoku:
@vapier, since you've approved this can you take another look?

cmb69:
Thank you @YakoYakoYokuYoku for all the work you've put into this! However, a related issue came up recently regarding PHP and libgmp, see https://github.com/php/php-src/pull/16507#issuecomment-2423824754. So unless the functions would be stored in TLS (thread-local storage), PHP probably can't use the feature. Note that I'm not talking about true thread-safety (that might not be necessary), but rather that each thread should have its own custom allocators.

Another issue is, that is should somehow forbidden to change/clear the allocators after any memory already been allocated. It might be sufficient to track whether any allocation has happend, and in this case let the calls to change/clear the allocators fail or silently do nothing.

YakoYakoYokuYoku:
In theory adding thread local storage specifiers should suffice though tell me if I'm wrong. On the other hand, the allocators guard could be implemented using atomic booleans and use `gd_error` to print messages, but I think that a warning in the documentation is enough, unless there's something elso to take into account.

cmb69:
> In theory adding thread local storage specifiers should suffice though tell me if I'm wrong.

I think that is sufficient, if supported on all supported platforms. A bit hard to verify, since most of CI is failing anyway.

> On the other hand, the allocators guard could be implemented using atomic booleans and use `gd_error` to print messages, but I think that a warning in the documentation is enough, unless there's something elso to take into account.

Thinking about that (see also https://github.com/php/php-src/pull/16609#pullrequestreview-2397576515), there actually should be locks when setting/retrieving the custom allocators, since you never know when a thread switch happens. If we only support a single change of the custom allocators (i.e. set them all at once), that lock could be completely implemented in libgd; if we support individual setters, we likely need to provide some API which needs to be called by the client who wants to change the custom allocators.

And to avoid havoc even when using a single-threaded process, we need to make it very clear in the documentation, that clients need to restore the old allocators when done with their own. And even that would not be sufficient, if libgd allows to install callbacks.

It is most unfortunate, that a simple and highly desireable feature may cause so many issues.

Issue #185, opened 2015-09-09
Add support for 4 byte UTF-8 characters and HTML entities over `&#xFFFF;` and `&#99999;`


stil:
Currently, GD supports only UTF-8 characters of 1-3 bytes. Four byte characters are unsupported. HTML entities higher than  `&#xFFFF;` and `&#99999;` are also unsupported.

It makes impossible to draw emojis, which are Unicode characters higher than U+1F601. UTF-8 representation of emojis are at least 4 bytes long. I thought I could bypass this issue using HTML entities. However it seems, that they are bugged too.

Invalid code lies in `gdft.c` file, `gdTcl_UtfToUniChar()` function.

Sources:
* [Only 1-3 byte UTF-8 characters in `gdft.c`](https://github.com/libgd/libgd/blob/master/src/gdft.c#L340-L342)
* [Code responsible for HTML entities parsing in `gdft.c`](https://github.com/libgd/libgd/blob/master/src/gdft.c#L293-L314)
* [More information on the bug in PHP bugtracker](https://bugs.php.net/bug.php?id=51135)
* [Proposed patch from PHP bugtracker](https://bugs.php.net/patch-display.php?bug=51135&patch=bundled-libgd-non-bmp-entities&revision=1267806282&display=1)

Possibly related issue
---
Someone at PHP bugtracker suspects, that due to integer overflow, Unicode characters higher than 65536 might also be broken.
> The responses to this report seem to imply that ImageTtfText WIIL work correctly for characters in the range U+10000 - U+1869F (65536 - 99999). However, this does not seem to be the case. When I call it with `&#65537;` (U+10001), it produces an "missing" glyph, even using a font that contains the character (CODE2001.ttf).

vapier:
can you attach a small file that illustrates the issue for us ?

stil:
4 byte UTF-8 characters issue
----

Below you can see glyphs of Arial Unicode font.

![arial_unicode](https://i.imgur.com/w8GTBS1.png)

And here you see, how GD draws consecutive glyphs starting from `0xFFD0` and ending at `0x1001F`.

![drawing_glyphs](https://i.imgur.com/edden2l.png)

You can clearly see, that characters after 0xFFFF are incorrectly parsed.

Check UTF-8 byte representation of 0xFFFF and 0x10000 Unicode characters here:
* http://unicode-table.com/en/FFFF/
* http://unicode-table.com/en/10000/

vapier:
[...]

PR #695, opened 2021-04-23
Issue-185 changes

sterlingpickens:
These changes should resolve https://github.com/libgd/libgd/issues/185
Please, test and let me know if any other changes need to be made.
It is my intent not to break things for other people.


vapier:
please undo all the whitespace changes.  they don't look necessary or related and make it very hard to understand what the patch is actually doing.

sterlingpickens:
I completely replaced 1 function with 3, but I can add the extra whitespaces everywhere if you like ?

sterlingpickens:
I'll try to clarify.
entities.h was updated to include all current html5 entities instead of only html4.
The gdTcl_UtfToUniChar function was removed. The comp_entities function was removed.
gdTcl_UtfToUniChar became gd_Entity_To_Unicode (no longer requires comp_entities), gd_JISX0208_To_Unicode (2-byte encoding not utf-8), and gd_UTF8_To_Unicode (now accepts 4-byte utf8 for the official unicode limit).
The reasoning behind splitting the functions is because JISX0208 and UTF8 are mutually exclusive and overlap.
The entities all start with &, so it kind of piggy backs regardless of the encoding, i've maintained that behavior.

I considered adding JISX0213, but that's a different matter.

vapier:
the indentation in src/entities.h is incorrect.

it's unclear how you're generating this, but it isn't with the entities.tcl that we've used historically.  however it's done, entities.tcl needs to be updated or replaced.

sterlingpickens:
I wrote a C program to do it, as the format of the website changed and it no longer works.

vapier:
we want that logic in the repo so we can verify & keep in sync easier.  and we don't want to keep around old scripts that are known to be outdated.

sterlingpickens:
I'd have to add the header portion of the printing and make it output the file as a whole.
Right now it is just outputting the relevant lines.
I suppose I can do that

sterlingpickens:
There is another way I can write gd_UTF8_To_Unicode that uses half as many lines of code as well.

sterlingpickens:
Ok, i've got the entities_gen.c to replace entities.tcl.
What else should I do ?

My use of FT_Select_Charmap(face, charmap->encoding); is not ideal.
When the user calls the gdImageStringFTEx function with a selected encoding, would the use of FT_Select_Charmap(face, FT_ENCODING_UNICODE); break things for anyone ?

It might be best to rewrite that entire section to map the selected encoding to an FT_Select_Charmap call
ie: FT_Select_Charmap(face, FT_ENCODING_ADOBE_CUSTOM);
Then we can check the return for error.

I think only FT_Set_Charmap existed when gdft.c was written.

sterlingpickens:
If you want to cancel this pull request. I can reopen again later once i'm done with all the other changes I need to make. It'll likely be in a broken state for a few days.

vapier:
*shrug* it doesn't cost us anything to leave open

sterlingpickens:
I think it works now. I'll have to do some extensive testing and look everything over, but i'm close.

sterlingpickens:
As far as I know this is done, unless someone else sees a problem.

src/entities.py
```diff
@@ -0,0 +1,94 @@
+#!/usr/bin/env python
```
vapier, line 1: let's use `python3`

src/entities.py
```diff
@@ -0,0 +1,94 @@
...
+from __future__ import print_function
```
vapier, line 2: delete this since we know it's python3

src/Makefile.am
```diff
@@ -42,8 +42,8 @@ endif
 EXTRA_DIST = \
 	bdftogd \
 	demoin.png \
-	entities.html \
-	entities.tcl \
+	entities.json \
```
vapier, line 45: instead of checking it into git (since it's like 150KB), how about we have entities.py fetch it on the fly & cache it ?

so if src/entities.json doesn't exist, download https://html.spec.whatwg.org/entities.json and save it there.  then read it from that path.

you'll also want to update .gitignore.

there is somewhat an argument to be made about not being able to recreate this file reliably depending on the git version, but i'm not sure we really care about that when we're committing the generated header output to the repo.

wdyt ?

```
import urllib.request
...
if not os.path.exists("entities.json"):
  with open("entities.json", "wb") as fp:
    fp.write(urllib.request.urlopen("https://...").read())
```

src/entities.py
```diff
@@ -0,0 +1,94 @@
...
+import json
```
vapier, line 3: add a module docstring before the imports that provides a quick overview of the purpose of this script.  that way when other devs open this file, they can quickly figure out what it's for.

src/entities.py
```diff
@@ -0,0 +1,94 @@
...
+import re
+
+### Declare headers
```
vapier, line 6: only need one `#` for comments

src/entities.py
```diff
@@ -0,0 +1,94 @@
...
+
+#include <stdlib.h>
+#include <stdint.h>
+
+//html entity strings are entity prefix + string + suffix limited
+//hex and dec should be limited to current unicode spec + entity prefix + suffix
+"""
+
+header_d = """
+struct entities_s{
```
vapier, line 36: space before the `{`

src/entities.py
```diff
@@ -0,0 +1,94 @@
...
+                  	char *name;
```
vapier, line 37: only indent with a single tab here (and below)

src/entities.py
```diff
@@ -0,0 +1,94 @@
...
+                        uint32_t unicode_a;
+                        uint32_t unicode_b;
+                };
+extern const struct entities_s entities[NR_OF_ENTITIES];
+
+#endif
+"""
+
+### Open files
+file_json = open("entities.json", mode="r")
```
vapier, line 47: use a context manager

```
with open(entities.json", mode="rb") as fp:
    data = json.load(fp)
```

src/entities.py
```diff
@@ -0,0 +1,94 @@
...
+file_ent_h = open("entities.h", mode="w")
+file_ent_c = open("entities.c", mode="w")
+
+### Initialize some objects
+x = json.load(file_json)
```
vapier, line 52: single letter vars are fine when the code is narrow in scope, but when there's many spread across the file, it gets hard to keep track of what's what

how about calling this `entities`

src/entities.py
```diff
@@ -0,0 +1,94 @@
...
+curline = 1
+totallines = 0
+len_name_max = 0
+p = re.compile("&\S+;")
```
vapier, line 56: use `r` strings

how about naming it `name_matcher` ?

src/entities.py
```diff
@@ -0,0 +1,94 @@
+#!/usr/bin/env python
+from __future__ import print_function
+import json
+import re
+
```
vapier, line 5: this script assumes it's always run with cwd==src.  we should enforce that.

```
import os
from pathlib import Path
...
SRCDIR = Path(__file__).resolve().parent
os.chdir(SRCDIR)
```

src/entities.py
```diff
@@ -0,0 +1,94 @@
...
+### Declare headers
+header_a = """/*
+ * Generated file - do not edit directly.
+ * This file was generated from:
+ *     https://html.spec.whatwg.org/entities.json
+ * by means of entities.py in the libgd src dir
+*/
+"""
+
+header_b = """
```
vapier, line 15: this is only ever used with `header_a`.  lets just merge them and call it `HEADER_C_FILE` (for `.c`).

src/entities.py
```diff
@@ -0,0 +1,94 @@
...
+#include "entities.h"
+
+const struct entities_s entities[NR_OF_ENTITIES] = {"""
+
+header_c = """/*
```
vapier, line 20: this would be `HEADER_H_FILE`

src/entities.py
```diff
@@ -0,0 +1,94 @@
...
+x = json.load(file_json)
+curline = 1
+totallines = 0
+len_name_max = 0
+p = re.compile("&\S+;")
+
+### Print file headers
+file_ent_c.write(header_a + header_b + "\n")
+file_ent_h.write(header_c)
+file_ent_h.write("#define GD_ENTITIES_H 1\n")
```
vapier, line 61: why not move to the header constant ?

src/entities.py
```diff
@@ -0,0 +1,94 @@
...
+file_ent_c.write("\t\t\t};" + "\n")
+file_ent_h.write("#define NR_OF_ENTITIES " + str(totallines) + "\n");
+file_ent_h.write("#define ENTITY_NAME_LENGTH_MAX " + str(len_name_max) + "\n");
+file_ent_h.write("#define ENTITY_HEX_LENGTH_MAX 10\n")
+file_ent_h.write("#define ENTITY_DEC_LENGTH_MAX 10\n")
+file_ent_h.write(header_d)
+### Close files
+file_json.close()
+file_ent_h.close()
+file_ent_c.close()
```
vapier, line 94: at the end of the file we should print a message like:

```
print("entities.h & entities.c updated from entities.json")
```

this way it's clear to devs what this thing just did

src/entities.py
```diff
@@ -0,0 +1,94 @@
...
+ * This file was generated from:
+ *     https://html.spec.whatwg.org/entities.json
+ * by means of entities.py in the libgd src dir
+*/
+#ifndef GD_ENTITIES_H
+
+#include <stdlib.h>
+#include <stdint.h>
+
+//html entity strings are entity prefix + string + suffix limited
```
vapier, line 31: spaces after `//`

src/entities.py
```diff
@@ -0,0 +1,94 @@
...
+#include "entities.h"
+
+const struct entities_s entities[NR_OF_ENTITIES] = {"""
+
+header_c = """/*
+ * Generated file - do not edit directly.
+ * This file was generated from:
+ *     https://html.spec.whatwg.org/entities.json
+ * by means of entities.py in the libgd src dir
+*/
```
vapier, line 25: space before this `*/`

src/entities.py
```diff
@@ -0,0 +1,94 @@
...
+            string = str(x[e]["codepoints"][0]) + ", 0}"
+
+        if curline == totallines:
+            file_ent_c.write(string + "\n")
+        else:
+            file_ent_c.write(string + "," + "\n")
+        curline+=1
+### Print file ends
+file_ent_c.write("\t\t\t};" + "\n")
+file_ent_h.write("#define NR_OF_ENTITIES " + str(totallines) + "\n");
```
vapier, line 86: use f-strings

```
file_ent_h.write(f"#define NR_OF_ENTITIES {totallines}\n");
```

src/entities.py
```diff
@@ -0,0 +1,94 @@
...
+file_ent_h.write("#define ENTITY_NAME_LENGTH_MAX " + str(len_name_max) + "\n");
+file_ent_h.write("#define ENTITY_HEX_LENGTH_MAX 10\n")
```
vapier, line 88: should move these constant strings to the single header constant to keep it simple

src/entities.py
```diff
@@ -0,0 +1,94 @@
...
+len_name_max = 0
+p = re.compile("&\S+;")
+
+### Print file headers
+file_ent_c.write(header_a + header_b + "\n")
+file_ent_h.write(header_c)
+file_ent_h.write("#define GD_ENTITIES_H 1\n")
+
+### Count how many matching lines we have
+for e in x:
```
vapier, line 64: i think this for loop can be one line:
```
totallines = sum(p.match(x) for x in entities)
```

then again, we can delete this loop entirely -- see below

src/entities.py
```diff
@@ -0,0 +1,94 @@
...
+    if p.match(e):
+        totallines+=1
+### Print json entities to struct
+for e in x:
+    if p.match(e):
+        if len(e) > len_name_max:
```
vapier, line 70: use the `max` func to collapse this `if` statement

src/entities.py
```diff
@@ -0,0 +1,94 @@
...
+            len_name_max = len(e)
+        string = "\t\t\t" + "{\"" + e.replace("&", "") + "\", "
```
vapier, line 72: entries should only be indented by a single tab

src/entities.py
```diff
@@ -0,0 +1,94 @@
...
+        file_ent_c.write(string.replace(";", ""))
+        if len(x[e]["codepoints"]) > 1:
```
vapier, line 74: since you refer to codepoints many times, just pull it out:

```
codepoints = x[e]["codepoints"]
```

src/entities.py
```diff
@@ -0,0 +1,94 @@
...
+            string = str(x[e]["codepoints"][0]) + ", " + str(x[e]["codepoints"][1]) + "}"
```
vapier, line 75: this for loop could be simplified with a little data manipulation

```
codepoints = x[e]["codepoints"]
# If there's only one codepoint, make sure the 2nd one is zero.  Otherwise this is ignored.
codepoints.append(0)
file_ent_c.write(str(codepoints[0]) + ", " + str(codepoints[1]) + "},\n")
```

src/gdft.c
```diff
@@ -267,168 +260,172 @@ gdMutexDeclare (gdFontCacheMutex);
...
+	static const uint8_t bound_low_lcase = '`';
+	static const uint8_t bound_hi_lcase = 'g';
+	static const uint8_t prefix0 = '&';
+	static const uint8_t prefix1 = '#';
+	static const uint8_t prefix2 = 'x';
+	static const uint8_t prefix3 = 'X';
+	static const uint8_t suffix = ';';
+	static const uint8_t ten = 10;
+
+	if(b0 != prefix0)
```
vapier, line 289: there should be a space between `if (`

src/gdft.c
```diff
@@ -267,168 +260,172 @@ gdMutexDeclare (gdFontCacheMutex);
...
+		return 0;
+
+	b0=(uint8_t)str[i++];
```
vapier, line 292: spaces around binary operators like `=`

src/gdft.c
```diff
@@ -267,168 +260,172 @@ gdMutexDeclare (gdFontCacheMutex);
...
-				}
-			} else {
-				for (i = 2; i < 8; i++) {
-					byte = *((unsigned char *) (str + i));
-					if (byte >= '0' && byte <= '9')
-						n = (n * 10) + (byte - '0');
-					else
-						break;
-				}
+static int gd_Entity_To_Unicode(const char *str, uint32_t *chPtr){
```
vapier, line 263: space between `) {`

src/gdft.c
```diff
@@ -267,168 +260,172 @@ gdMutexDeclare (gdFontCacheMutex);
...
+	// str is the UTF8 next character pointer, or a pointer to a string of 4 different kinds of html entity notation
+	// chPtr is the int for the result
+	// return value is the offset from str to the char for the next search to start
+	// function falls through and returns 0 if no valid entity detected, leaving chPtr alone.
+	// HTML4.0/5.0 entities in hexadecimal form (e.g. &#x00021; and &#X00021;), in decimal form (e.g. &#33;),  or in string form (e.g. &excl;)
+	char entity_name_buf[ENTITY_NAME_LENGTH_MAX];
+	uint32_t n = 0;
+	uint32_t i = 1;
+	uint32_t b0 = *(uint8_t *)str;
```
vapier, line 272: you immediately clobber this assignment below, so delete this

src/gdft.c
```diff
@@ -1278,7 +1278,8 @@ BGD_DECLARE(char *) gdImageStringFTEx (gdImage * im, int *brect, int fg, const c
 		}
 	}
 	if (encodingfound) {
-		FT_Set_Charmap(face, charmap);
+		FT_Select_Charmap(face, charmap->encoding);
+		//FT_Set_Charmap(face, charmap);
```
vapier, line 1282: delete dead code, don't leave it commented out

src/gdft.c
```diff
@@ -1293,25 +1294,43 @@ BGD_DECLARE(char *) gdImageStringFTEx (gdImage * im, int *brect, int fg, const c
 			any2eucjp (tmpstr, string, BUFSIZ);
 			next = tmpstr;
 		} else {
-			next = string;
+			next = (char *)string;
```
vapier, line 1297: something is broken if we're casting the const away.  please undo these changes.

vapier (changes requested):

sterlingpickens:
Wouldn't be easier to just merge then we can resolve the minor stuff ?

src/gdft.c
sterlingpickens: n needs to be 0.

vapier:
having 30 tiny CLs that just shuffle style & bits around is not an improvement.  having a single commit that implements a single thing makes history management a hell of a lot easier.

so yes, there is a lot of style feedback mixed in, but that's because this PR is going to be squashed into a single commit when it's merged.

src/gdft.c
vapier: i'm not referring to `n`, i'm referring to `b0`.

that said, i was thinking `b0` was a pointer you were using to walk `str` rather than a char you've pulled out, so nm me here.

sterlingpickens:
I guess i'm just getting kinda slow or worn out on this thing at this point.
Looks like i'll have to work up the energy for the python updates tomorrow.

src/Makefile.am
sterlingpickens: One concern I have here is if they update the json entities to have 3 unicode, or more, libgd would need modified to support that. It seems all the current ones are 1 or 2 for now.
I like the idea of keeping the json file on git, then if a dev decides to delete it and let the script snag a new one they can. I dunno how often they add to the spec, but most fonts i've tried are missing half the glyphs anyways.

sterlingpickens:
I think I addressed all/most of those requests.
I guess we need a decisions on src/Makefile.am and .gitignore.

sterlingpickens:
utf8-to-uni-01.c is 24% faster than utf8-to-uni-02.c when both compiled with -O2 -march=native
utf8-to-uni-01.c is 31% faster than utf8-to-uni-02.c when both compiled with -O3 -march=native

also 15 lines less code.
Do you think I should swap out that function ? 

http://sterlingdesktops.com/pub/test/utf8/utf8-to-uni-01.c
http://sterlingdesktops.com/pub/test/utf8/utf8-to-uni-02.c

It seems kinda silly, the function probably only takes less than 1% of the runtime in libgd.
The entities function is the one that can be made much better.

sterlingpickens:
Ok, what do you think now ?

src/Makefile.am
vapier: i don't think we have to worry about this too much.  we'll have the generated source files checked in, so releases would never run the script.  we aren't guaranteeing that every release will be able to generate the latest entities.json, or even that it even needs to run at all.

src/entities.py
```diff
@@ -0,0 +1,129 @@
...
+ * This file was generated from:
+ *     https://html.spec.whatwg.org/entities.json
+ * by means of entities.py in the libgd src dir
+ */
+
+#include "entities.h"
+
+const struct entities_s entities[NR_OF_ENTITIES] = {
+"""
+H_FILE_HEAD = """ /*
```
vapier, line 43: omit the space before the `/`

src/entities.py
```diff
@@ -0,0 +1,129 @@
...
+import os
+import os.path
+import re
+import urllib.request
+from pathlib import Path
+
+__author__ = "Sterling Pickens"
+__copyright__ = "Copyright 2021, The libgd Project"
+__credits__ = ["Sterling Pickens", "Mike Frysinger"]
+__license__ = "GPLv3 or later"
```
vapier, line 21: this is OK if you want to keep it, but libgd's license isn't GPL.  it's more BSD/MIT-ish.  so normally here we'd say like "MIT; see the COPYING file".

but again, if you want to keep it GPLv3+, that's OK if we put a disclaimer like "the script is under the GPLv3, but the generated header is under libgd's license".

src/entities.py
```diff
@@ -0,0 +1,129 @@
...
+__maintainer__ = "Sterling Pickens"
+__email__ = "sterling_pickens@yahoo.com"
+__status__ = "Production"
+
+# Ensure we are only in src dir
+SRCDIR = Path(__file__).resolve().parent
+os.chdir(SRCDIR)
+
+# Declare File content strings
+C_FILE_HEAD = """ /*
```
vapier, line 32: omit the space before the `/`

src/entities.py
vapier: looks like these members are still using spaces for indentation.  we just want one tab.

src/entities.py
```diff
@@ -0,0 +1,129 @@
...
+#define ENTITY_DEC_LENGTH_MAX 10
+
+// html entity strings are entity prefix + string + suffix limited
+// hex and dec should be limited to current unicode spec + entity prefix + suffix
+
+struct entities_s {
+                  	char *name;
+                        uint32_t unicode_a;
+                        uint32_t unicode_b;
+                };
```
vapier, line 68: don't indent this line at all

src/entities.py
```diff
@@ -0,0 +1,129 @@
...
+# Load json obj
+with open("entities.json", "rb") as file_json:
+    entities = json.load(file_json)
+
+# Initialize some objects
+curline = 1
+name_matcher = re.compile(r"&\S+;")
+
+# Sum of total matching entities
+total_entities = sum(1 if name_matcher.match(key) else 0 for key in entities)
```
vapier, line 90: bool values count as 0/1, so you can simplify this a little:

`total_entities = sum(bool(name_matcher.match(key)) key in entities)`

src/entities.py
```diff
@@ -0,0 +1,129 @@
...
+
+# Find longest entity
+len_name_max = max(len(key) if name_matcher.match(key) else 0 for key in entities)
```
vapier, line 93: this one we'd shuffle so that the generator doesn't produce anything if it doesn't match

`len_name_max = max(len(key) for key in entities if name_matcher.match(key))`

src/entities.py
```diff
@@ -0,0 +1,129 @@
...
+    file_ent_h.write(H_FILE_TAIL)
+
+# Write entities.c file
+with open("entities.c", mode="w") as file_ent_c:
+    file_ent_c.write(C_FILE_HEAD)
+    # Write json entities to struct
+    for key in entities:
+        if name_matcher.match(key):
+            string = "\t" + "{\"" + key.replace("&", "") + "\", "
+            string = string.replace(";", "")
```
vapier, line 109: just move this to the line above:

`string = "\t{\"" + key.replace("&", "").replace(";", "") + "\", "`

src/entities.py
```diff
@@ -0,0 +1,129 @@
...
+            codepoints = entities[key]["codepoints"]
+            string = string + str(codepoints[0]) + ", "
+            if len(codepoints) > 1:
+                string = string + str(codepoints[1]) + "}"
+            else:
+                string = string + "0}"
+            if len(codepoints) > 2:
+                print(f"Warning: entity with >2 codepoints detected")
+            if curline == total_entities:
```
vapier, line 118: just omit this check.  dangling commas are fine.  just write:

`string += ",\n"`

src/entities.py
```diff
@@ -0,0 +1,129 @@
...
+                string = string + "\n"
+            else:
+                string = string + ",\n"
+            file_ent_c.write(string)
+            curline+=1
```
vapier, line 123: spaces around the `+=`

src/entities.py
```diff
@@ -0,0 +1,129 @@
...
+    # Write file end
+    file_ent_c.write(f"\t}};")
```
vapier, line 125: no need for `f` here

also, don't want the `\t`

src/entities.py
```diff
@@ -0,0 +1,129 @@
...
+
+# Print Summary
+print(f"entities.h & entities.c updated from entities.json")
```
vapier, line 128: don't need the `f` here

src/gdft.c
```diff
@@ -267,168 +260,172 @@ gdMutexDeclare (gdFontCacheMutex);
...
-				}
-			} else {
-				for (i = 2; i < 8; i++) {
-					byte = *((unsigned char *) (str + i));
-					if (byte >= '0' && byte <= '9')
-						n = (n * 10) + (byte - '0');
-					else
-						break;
-				}
+static int gd_Entity_To_Unicode(const char *str, uint32_t *chPtr) {
```
vapier, line 263: move the `{` to the next line by itself

src/gdft.c
```diff
@@ -267,168 +260,172 @@ gdMutexDeclare (gdFontCacheMutex);
...
-	/*
-	 * Unroll 1 to 3 byte UTF-8 sequences, use loop to handle longer ones.
-	 */
-
-	byte = *((unsigned char *) str);
 #ifdef JISX0208
-	if (0xA1 <= byte && byte <= 0xFE) {
-		int ku, ten;
 
+static int gd_JISx0208_To_Unicode(const char *str, uint32_t *chPtr) {
```
vapier, line 364: move the `{` to the next line by itself

src/entities.py
```diff
@@ -0,0 +1,129 @@
...
+C_FILE_HEAD = """ /*
+ * Generated file - do not edit directly.
+ * This file was generated from:
+ *     https://html.spec.whatwg.org/entities.json
+ * by means of entities.py in the libgd src dir
+ */
+
+#include "entities.h"
+
+const struct entities_s entities[NR_OF_ENTITIES] = {
```
vapier, line 41: since this is being moved to a sep module instead of a header file, we need to take care of symbol names.  in a static libgd.a, `entities` will bleed out.  you'll want to write `gd_entities` instead.

src/gdft.c
```diff
@@ -267,168 +260,172 @@ gdMutexDeclare (gdFontCacheMutex);
...
+		return 0;
+
+	b0 = (uint8_t)str[i++];
+	if (b0 == prefix1) {
+		// hex or decimal
+		b0 = (uint8_t)str[i++];
+		if (b0 == prefix2 || b0 == prefix3) {
+			// is hex
+			b0 = (uint8_t)str[i++];
+			while (i<ENTITY_HEX_LENGTH_MAX) {
```
vapier, line 299: spaces around the `<`

src/gdft.c
```diff
@@ -267,168 +260,172 @@ gdMutexDeclare (gdFontCacheMutex);
...
-				if (byte == ';') {
-					*p++ = '\0';
-					res = bsearch(&key, entities, NR_OF_ENTITIES,
-					              sizeof(entities[0]), *comp_entities);
-					if (res) {
-						*chPtr = (Tcl_UniChar) res->value;
-						return ++i;
-					}
+			// is dec
+			while (i<ENTITY_DEC_LENGTH_MAX) {
```
vapier, line 318: spaces around the `<`

src/gdft.c
```diff
@@ -267,168 +260,172 @@ gdMutexDeclare (gdFontCacheMutex);
...
+			if (b0 == suffix) {
+				*chPtr = n;
+				chPtr[1] = 0;
+				return i;
+			}
+		}
+	} else {
+		// string format alphanumeric
+		// copy str into buffer untill '\0' or ;
+		while (i<ENTITY_NAME_LENGTH_MAX) {
```
vapier, line 335: spaces around the `<`

src/entities.py
```diff
@@ -0,0 +1,129 @@
...
+"""
+H_FILE_TAIL = """
+#define ENTITY_HEX_LENGTH_MAX 10
+#define ENTITY_DEC_LENGTH_MAX 10
+
+// html entity strings are entity prefix + string + suffix limited
+// hex and dec should be limited to current unicode spec + entity prefix + suffix
+
+struct entities_s {
+                  	char *name;
```
vapier, line 65: should change this to `const char *name` too

src/gdft.c
```diff
@@ -267,168 +260,172 @@ gdMutexDeclare (gdFontCacheMutex);
...
+				break;
+			}
+			entity_name_buf[n++] = b0;
+			b0 = (uint8_t)str[i++];
+		}
+		// not an entity, too long
+		if (b0 != suffix)
+			return 0;
+		for (n = 0; n < NR_OF_ENTITIES; n++) {
+			if (strcmp((const char *)entities[n].name, (const char *)entity_name_buf) == 0) {
```
vapier, line 351: you shouldn't need the `const char *` casts

src/gdft.c
```diff
@@ -267,168 +260,172 @@ gdMutexDeclare (gdFontCacheMutex);
...
+				chPtr[0] = (uint32_t)entities[n].unicode_a;
```
vapier, line 352: `unicode_a` and `unicode_b` are `uint32_t` already, so no need to cast

src/entities.py
```diff
@@ -0,0 +1,129 @@
...
+                        uint32_t unicode_a;
```
vapier, line 66: how about naming `codepoint1` & `codepoint2` instead.  "unicode" is ambiguous as to what it means.

src/gdft.c
```diff
@@ -267,168 +260,172 @@ gdMutexDeclare (gdFontCacheMutex);
...
+		b2 = (uint8_t)str[2];
+		b3 = (uint8_t)str[3];
+		if (((b1 & 0xC0) == 0x80) && ((b2 & 0xC0) == 0x80) && ((b3 & 0xC0) == 0x80)) {
+			*chPtr = ((b0 & 0x7) << 18) | ((b1 & 0x3F) << 12) | ((b2 & 0x3F) << 6) | (b3 & 0x3F);
+			return 4;
+		}
+		*chPtr = b0;
+		return 1;
+	}
+	printf("WARNING: utf-8 above 21-bit unicode range\n");
```
vapier, line 421: you want to include `gd_errors.h` and use `gd_error()` here instead

src/gdft.c
```diff
@@ -1105,12 +1102,15 @@ BGD_DECLARE(char *) gdImageStringFTEx (gdImage * im, int *brect, int fg, const c
 	FT_UInt glyph_index;
 	double sin_a = sin (angle);
 	double cos_a = cos (angle);
-	int  i, ch;
+	uint32_t  i, ch;
```
vapier, line 1105: only once space before `i`

src/gdft.c
```diff
@@ -1303,15 +1303,33 @@ BGD_DECLARE(char *) gdImageStringFTEx (gdImage * im, int *brect, int fg, const c
...
 	while (*next) {
 		int len;
 		ch = *next;
 		switch (encoding) {
 		case gdFTEX_Unicode: {
-			/* use UTF-8 mapping from ASCII */
-			len = gdTcl_UtfToUniChar (next, &ch);
+			// html entity
+			len = gd_Entity_To_Unicode(next, ch_entity);
+			if(len == 0){
```
vapier, line 1315: spaces: `if (len == 0) {`

src/gdft.c
```diff
@@ -1303,15 +1303,33 @@ BGD_DECLARE(char *) gdImageStringFTEx (gdImage * im, int *brect, int fg, const c
...
+#ifdef JISX0208
+				len = gd_JISx0208_To_Unicode(next, &ch);
+#else
+				// UTF-8
+				len = gd_UTF8_To_Unicode(next, &ch);
+#endif
+			}else{
+				if(ch_entity[1] == 0){ // single codepage entity
+					ch = ch_entity[0];
+				}else{ // dual codepage entity
```
vapier, line 1325: spaces: `} else {`

src/gdft.c
```diff
@@ -1303,15 +1303,33 @@ BGD_DECLARE(char *) gdImageStringFTEx (gdImage * im, int *brect, int fg, const c
...
+			// html entity
+			len = gd_Entity_To_Unicode(next, ch_entity);
+			if(len == 0){
+#ifdef JISX0208
+				len = gd_JISx0208_To_Unicode(next, &ch);
+#else
+				// UTF-8
+				len = gd_UTF8_To_Unicode(next, &ch);
+#endif
+			}else{
```
vapier, line 1322: spaces: `} else {`

src/gdft.c
```diff
@@ -1303,15 +1303,33 @@ BGD_DECLARE(char *) gdImageStringFTEx (gdImage * im, int *brect, int fg, const c
...
+				if(ch_entity[1] == 0){ // single codepage entity
```
vapier, line 1323: spaces: `if (ch_entity[1] == 0) {`

src/gdft.c
```diff
@@ -1303,15 +1303,33 @@ BGD_DECLARE(char *) gdImageStringFTEx (gdImage * im, int *brect, int fg, const c
...
+					ch = ch_entity[0];
+				}else{ // dual codepage entity
+					if(charmap->encoding == FT_ENCODING_MS_SYMBOL)
```
vapier, line 1326: space after the `if`

src/gdft.c
```diff
@@ -1378,8 +1396,12 @@ BGD_DECLARE(char *) gdImageStringFTEx (gdImage * im, int *brect, int fg, const c
 			next++;
 			break;
 		}
-		text[i] = ch;
-		i++;
+		// avoid out of bounds
+		if ((i << 2) > (text_size-2)) {
```
vapier, line 1400: spaces around the `-`

src/gdft.c
```diff
@@ -1378,8 +1396,12 @@ BGD_DECLARE(char *) gdImageStringFTEx (gdImage * im, int *brect, int fg, const c
...
+			text_size += text_step;
+			text = (uint32_t *)gdRealloc((void *)text, text_size);
```
vapier, line 1402: don't need any of these casts

src/gdft.c
```diff
@@ -267,168 +260,172 @@ gdMutexDeclare (gdFontCacheMutex);
...
-				*chPtr = (Tcl_UniChar) (((byte & 0x0F) << 12)
-				                        | ((str[1] & 0x3F) << 6) | (str[2] & 0x3F));
-				return 3;
-			}
-			/*
-			 * A three-byte-character lead-byte not followed by
-			 * two trail-bytes represents itself.
-			 */
+static int gd_UTF8_To_Unicode(const char *str, uint32_t *chPtr) {
+	// reads the contents pointed to by str and outputs a unicode codepage into object pointed to by chPtr
```
vapier, line 382: put a period at the end of the sentence so we know where the next one starts

vapier, line 382: func API docs should be outside/above the func rather than in its body.  this is where readers expect it when scanning code.

src/gdft.c
```diff
@@ -267,168 +260,172 @@ gdMutexDeclare (gdFontCacheMutex);
...
+	// return is the bytes to offset for the first char of a subsequent read
```
vapier, line 383: this is a bit hard to understand.  how about:

`// Returns the number of bytes read from |str|.`

src/entities.py
sterlingpickens: So you'd prefer replacing that line with the following ?
__license__ = "MIT; see the COPYING file"

Should the .c/.h files also contain the same line ?

src/gdft.c
sterlingpickens: Do you want just a warning or a fatal error ?
Previous behavior would continue untill the entire loop finished.
So any bad lead byte would and does make any following bytes nonsense.
imo a warning seems like enough, the user is likely to notice.

sterlingpickens:
That should address all those requests except the gd_error use.

sterlingpickens:
Should we remove the old "* gdTcl_UtfToUniChar is borrowed from Tcl ..." comment as well ?
That is 41 lines.
Or are we still sharing enough to mention it ?

src/entities.py
vapier: if you're OK with that change, that would make everything easier

src/gdft.c
vapier: as a library, we should very very rarely be fatal.  it makes users a lot harder to be reliable.

a warning in this specific case sounds fine to me.

vapier:
> Should we remove the old "* gdTcl_UtfToUniChar is borrowed from Tcl ..." comment as well ?

i'm indifferent.  if you think the amount of borrowed code is still significant, leaving the comment sounds reasonable.

sterlingpickens:
Is this done now ?

pierrejoye:
What is the status on this PR?

sterlingpickens:
It works for me, although there have been a handfull of changes in master since april.
I don't know if there are issues for other people.

pierrejoye:
@sterlingpickens could you apply a change (whatever, white space etc) to run the CI again? Last one it was broken. We also improved CI now using github actions on Windows, Mac and linux, intel and ARM. Thanks :)

sterlingpickens:
I tried to sync with main, hopefully that didn't break it.
I really don't understand this CI stuff.
A problem with gd_entities ?
Maybe something that got merged wasn't supposed to happen.
This entire PR has dragged out too long and i'm lost. It might have to be started over from scratch.

sterlingpickens:
Ok, it looks like adding entities.c to CMakeLists.txt was needed.

src/entities.py
```diff
@@ -0,0 +1,115 @@
...
+Delete entities.json if you want to fetch the latest version.
+The current version has 1 and 2 codepoint entities.
+If the spec is updated to include 3 or more codepoints per entity, then this script along with
+gdImageStringFTEx and gd_Entity_To_Unicode (from gdft.c) will need to be modified.
+A warning is issued for any such entities found and only the first 2 codepoints are used.
+"""
+
+import json
+import os
+import os.path
```
vapier, line 13: delete this as it's redundant with the import above

src/entities.py
```diff
@@ -0,0 +1,115 @@
...
+import re
+import urllib.request
+from pathlib import Path
```
vapier, line 16: this should come before `import re`

src/entities.py
```diff
@@ -0,0 +1,115 @@
...
+__credits__ = ["Sterling Pickens", "Mike Frysinger"]
+__license__ = "MIT; see the COPYING file"
+__version__ = "1.0.0"
+__maintainer__ = "Sterling Pickens"
+__email__ = "sterling_pickens@yahoo.com"
+__status__ = "Production"
+
+# Ensure we are only in src dir
+SRCDIR = Path(__file__).resolve().parent
+os.chdir(SRCDIR)
```
vapier, line 29: shouldn't need to chdir.  use SRCDIR to construct paths.  see below.

src/entities.py
```diff
@@ -0,0 +1,115 @@
...
+#define ENTITY_DEC_LENGTH_MAX 10
+"""
+H_FILE_TAIL = """
+extern const struct gd_entities_s gd_entities[NR_OF_ENTITIES];
+
+#endif
+"""
+
+# Check for entities.json and fetch if needed
+if not os.path.exists("entities.json"):
```
vapier, line 69: if you create a var like `entitites_json = SRCDIR / 'entities.json'`, then a lot of this code is simplified.

```
if entities_json.exists():
    ....
    entities_json.write_bytes(urllib.request.urlopen(....
...
entities = json.load(entities_json.read_bytes())
```

src/entities.py
```diff
@@ -0,0 +1,115 @@
...
+
+# Find longest entity
+len_name_max = max(len(key) for key in entities if name_matcher.match(key))
+
+# Write entities.h file
+with open("entities.h", mode="w") as file_ent_h:
+    file_ent_h.write(H_FILE_HEAD)
+    file_ent_h.write(f"#define NR_OF_ENTITIES {total_entities}\n");
+    file_ent_h.write(f"#define ENTITY_NAME_LENGTH_MAX {len_name_max}\n\n");
+    file_ent_h.write("struct gd_entities_s {\n\tconst char *name;\n\tuint32_t codepoint1;\n\tuint32_t codepoint2;\n};\n");
```
vapier, line 91: this content should just be part of `H_FILE_TAIL`

src/entities.py
```diff
@@ -0,0 +1,115 @@
...
+        if name_matcher.match(key):
+            string = "\t" + "{\"" + key.replace("&", "").replace(";", "") + "\", "
+            codepoints = entities[key]["codepoints"]
+            string = string + str(codepoints[0]) + ", "
+            if len(codepoints) > 1:
+                string = string + str(codepoints[1]) + "}"
+            else:
+                string = string + "0}"
+            if len(codepoints) > 2:
+                print("Warning: entity with >2 codepoints detected")
```
vapier, line 108: this should be an assert, not a warning

src/entities.py
```diff
@@ -0,0 +1,115 @@
...
+    file_ent_h.write("struct gd_entities_s {\n\tconst char *name;\n\tuint32_t codepoint1;\n\tuint32_t codepoint2;\n};\n");
+    file_ent_h.write(H_FILE_TAIL)
+
+# Write entities.c file
+with open("entities.c", mode="w") as file_ent_c:
+    file_ent_c.write(C_FILE_HEAD)
+    # Write json entities to struct
+    for key in entities:
+        if name_matcher.match(key):
+            string = "\t" + "{\"" + key.replace("&", "").replace(";", "") + "\", "
```
vapier, line 100: transform the key in place:
`key = key.replace("&", "").replace(";", "")`

src/entities.py
```diff
@@ -0,0 +1,115 @@
...
+            codepoints = entities[key]["codepoints"]
```
vapier, line 101: pad codepoints out in place:
```
codepoints = [str(x) for x in entities[key]["codepoints"]]
if len(codepoints) == 1:
    codepoints.append("0")
```

src/entities.py
```diff
@@ -0,0 +1,115 @@
...
+            string = string + str(codepoints[0]) + ", "
+            if len(codepoints) > 1:
+                string = string + str(codepoints[1]) + "}"
+            else:
+                string = string + "0}"
+            if len(codepoints) > 2:
+                print("Warning: entity with >2 codepoints detected")
+            file_ent_c.write(string + ",\n")
```
vapier, line 109: once you do the above tweaks, the final code gets a lot simpler and easier to read.  the string concats you've written are very hard to follow, but now you can write:

```
file_ent_c.write("\t{\"" + key + "\", " + ",".join(codepoints) + "},\n")
```

src/entities.py
```diff
@@ -0,0 +1,115 @@
...
+extern const struct gd_entities_s gd_entities[NR_OF_ENTITIES];
+
+#endif
+"""
+
+# Check for entities.json and fetch if needed
+if not os.path.exists("entities.json"):
+    print("entities.json not found, attempting to download...")
+    with open("entities.json", "wb") as file_json:
+        file_json.write(urllib.request.urlopen("https://html.spec.whatwg.org/entities.json").read())
```
vapier, line 72: you use this URL twice.  pull it out into a sep constant please.

.gitignore
```diff
@@ -24,6 +24,7 @@ CTestTestfile.cmake
 libimagequant
 libimagequant-prefix
 src/deps.mk
+src/entities.json
```
vapier, line 27: listing this in gitignore is good, but then you've actually include the json file below.  you should delete the file from the PR.

vapier (changes requested):

src/entities.py
sterlingpickens: If I make this an assert, instead of a warning, then the python script will require updating if it encounters >2 codepoints. I'll do this, but just keep that in mind.

src/entities.py
sterlingpickens: I don't understand this.

src/entities.py
sterlingpickens: I'll have to think about this for a few hours. Everything else is done.

pierrejoye:
Thanks @sterlingpickens :)

About the CI, ono every new PR, or push to master or a PR, the builds&tests are ran on the various configurations. You can see the result in the PR or directly [here](https://github.com/libgd/libgd/actions).

pierrejoye:
@sterlingpickens And my apologies, it should have been merged long ago. I can give you a hand to get it done. 

@vapier what are the remaining issues? The Python script, as long as it works, is not very important. It can always be improved later or use jq to create the header file from Json :)

pierrejoye:
I approved it for now so the tests will be executed :)

sterlingpickens:
Had some mistakes in that last commit.

sterlingpickens:
I have to insert a 0 as a placeholder for the ones with only one codepage, that likely contributes to the "hard to read" aspect of the entities.py main loop.

pierrejoye:
image

Well done!

The only thing I would like is some tests for the multibyte entities, then we are good to go I think. Thoughts?

sterlingpickens:
Do you mean inside entities.py ?

pierrejoye:
> Do you mean inside entities.py ?

the resulting C entities used from freetype functions. I suppose some will never disappear, so we can add a test to ensure we don't break it in future releases. Does it make sense?

sterlingpickens:
Yes, if the entites.py hits the assert with an error. Then anything it generates will be garbage.
The way I had it before today was to just continue and warn, only taking the 1 and 2 codepage entities to generate the C file.

sterlingpickens:
So, maybe create a backup, and/or refuse to overwrite, the .c/.h entities if any problem is detected.

pierrejoye:
backup is in git so we don't need it, no?

The tests should ensure two things:

- the generated entities c file.is correct and build
- the GD functions handle multibytes are handles correctly

sterlingpickens:
I don't think freetype will have an issue with any codepages we give it(0 to UINT32_MAX), it'll just print a placeholder glyph if it's not valid. For the entities.c any entity of the form  {char *, uint32_t, uint32_t} should be fine.
ie: {char *, uint32_t, uint32_t, uint32_t} would be a problem or NR_OF_ENTITIES being wrong. We shouldn't have to worry about that if entities.py works right.

I guess you would like something in the tests directory which validates entities.c/.h ?

pierrejoye:
yes :)

pierrejoye:
and also the code who handles the input texts and pass the right codepoints
to freetype.

sterlingpickens:
Ok, I wrote several of those little test programs back in april. I'll try to work up the energy to repurpose and clean some code up for a tests subdir.
ie:
extract all supported supported glyphs from 1 or 2 common fonts and run it through libgd.
Then create a entities.txt, containing all current known entities, and run that through.
Also cycle through all supported unicode ranges (some are reserved), just to verify there are no overruns or segfaults.

pierrejoye:
Thank you @sterlingpickens !! Awesome work!

I will do a 2.3.3 release today or later this week, then master will become 2.4 and we can merge this PR. finally, thank you for the hard work and patience :)

src/entities.py
pierrejoye: @vapier I think it is ok as it is now. This script should be ran by maintainers only anyway, before a release etc. 

If you don't see any other thing, are you good to go with this PR? (I am :)

maximepvrt:
@pierrejoye Is merging this pull request still part of the project?

srett:
So after @sterlingpickens addressed all the remarks and issues, the PR just goes completely ignored for 3 years while the earliest threads on Stackoverflow and random forums regarding emoji rendering in gd/php go back over a decade?
☹️

PR #698, opened 2021-05-05
Support writing lossless WebP

adamsilverstein:
Originally https://github.com/libgd/libgd/pull/327

Lossless WebP is a rather interesting alternative to PNG, and already
supported by `gdImageCreateFromWebp*()`, so we add support for
`gdImageWebp*()`, too.

We can stick with the existing API, using the quality parameter to
request lossless encoding if it is set to `gdWebpLossless`, which we
define to `PHP_INT` (to avoid adding a new dependency to gd.h, we hard-
code the value – we're assuming `sizeof(int)==4` anyway).


adamsilverstein:
@cmb69 I'm going to test this out in WordPress, can you confirm that for previous versions of libgd, if I pass a quality value of 101 to `imagewebp` will it use 100 instead (or would I get an error)?

cmb69:
The quality is passed to WebPEncodeRGBA(); not sure how that function handles out of range values.

cmb69:
Thank you!

There is a caveat, though: WebPEncodeLosslessRGBA() is only [available as of libwebp 1.2.0](https://github.com/webmproject/libwebp/commit/c2416c9b610c49f473acea30c9e57ac80bd588e7), but we do not require any particular libwebp version so far. Either we raise the requirements, or we must use WebPEncodeLosslessRGBA() only if available.

vapier:
libwebp-1.2.0 was only released a few months ago, so we can't require it :/

cmb69:
Sorry, I've been confused by the GH UI. Actually, it should be available as of 0.2.0, released 2012.

cmb69:
> I'll see if I can figure out how to add that in configure, thanks for the lead. hopefully I can find a similar capability check already in use for some other part.

See, for instance, https://github.com/libgd/libgd/blob/82d260950589563a1af9c56f4ce5fde843a695ae/CMakeLists.txt#L128-L133

Note that GD also has an autotools build chain, which also needs respective checkks. I think we can ignore the Windows build chains (at least for now).

adamsilverstein:
> Of course, it would be better to check during configure that libwebp >= 0.2 is available

To clarify - do you mean adding that as a requirement in the [existing `ENABLE_WEBP` check](https://github.com/libgd/libgd/blob/82d260950589563a1af9c56f4ce5fde843a695ae/CMakeLists.txt#L120-L122), or adding a new check that sets something like `ENABLE_WEBP_LOSSLESS`?

vapier:
requiring libwebp 0.2.0 is fine, so just make it the min version all the time and call it a day

adamsilverstein:
I took a pass, however I am working pretty blindly here - meaning I'm not sure what I am doing 🤷🏼 😄 . I question if this change works because I tried higher version number and configure/make still worked for me locally.

cmb69:
Did you run `bootstrap.sh`?

adamsilverstein:
> Did you run bootstrap.sh?

Aha!

configure.ac
```diff
@@ -279,7 +279,7 @@ GD_LIB_PKG_CHECK([LIBTIFF], [TIFF], [tiff], [libtiff-4], [
 ])
 
 dnl Check for webp support.
-GD_LIB_PKG_CHECK([LIBWEBP], [WEBP], [webp], [libwebp], [
+GD_LIB_PKG_CHECK([LIBWEBP], [WEBP], [webp], [libwebp >= 0.2.0], [
```
adamsilverstein, line 282: @vapier is this all that is required to ensure libwebp >= 0.2.0 is required?

vapier: yep

vapier (approved):

pierrejoye:
good morning :)

Thanks for this PR!

anything left here? Need support?

morsssss:
@adamsilverstein , @pierrejoye , it strikes me that we should port this change over to PHP right away. I don't know if we'll get it in before feature freeze, which is tomorrow, but it's worth a try...

https://wiki.php.net/todo/php81

cmb69 (approved):
In my opinion, this is still good for PHP-8.1 even after feature freeze (should happen before ABI freeze, though).

morsssss:
> In my opinion, this is still good for PHP-8.1 even after feature freeze (should happen before ABI freeze, though).

what is ABI freeze? (I see the next milestone looks like Beta 1 on 22 July)

cmb69:
> what is ABI freeze?

That would be tagging of RC1 (Aug, 31st)

morsssss:
On the libgd side... is this ready to merge?

PR #701, opened 2021-05-26
Minor AVIF fixes

morsssss:
These are inspired by the work I've done to propagate libgd's AVIF support into PHP's bundled gd fork.

One of these changes fixes a small memory leak.

In the other - in the original PR, I hadn't realized that `gd_error()` outputs a `LOG_WARNING`. I use `gd_error()` throughout `gd_avif.c`, except for the case where we notice that an image's color profile isn't sRGB. This message is significant, but not as much as the errors that occur when we can't process the user's request at all. So I'd explicitly used `LOG_WARNING`.

That makes no sense, though, if we'd be outputting a `LOG_WARNING` level anyway! So I've just downgraded this one to `LOG_NOTICE`.

Others will have more experience than I do with libgd error levels; I'd cheerfully defer to them here.

### DIFF
--- a/src/gd_avif.c
+++ b/src/gd_avif.c
@@ -374,7 +374,7 @@ BGD_DECLARE(gdImagePtr) gdImageCreateFromAvifCtx (gdIOCtx *ctx)
 		goto cleanup;
 
 	if (!isAvifSrgbImage(decoder->image))
-		gd_error_ex(LOG_WARNING, "Image's color profile is not sRGB");
+		gd_error_ex(LOG_NOTICE, "Image's color profile is not sRGB");
 
 	// Set up the avifRGBImage, and convert it from YUV to an 8-bit RGB image.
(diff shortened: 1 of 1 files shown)


vapier:
seems reasonable, thanks

vapier:
hmm, my build is broken now w/libavif-0.9.0

```
gd_avif.c: In function ‘readFromCtx’:
gd_avif.c:194:19: error: implicit declaration of function ‘avifMalloc’; did you mean ‘avifAlloc’? [-Werror=implicit-function-declaration]
  194 |         dataBuf = avifMalloc(size);
      |                   ^~~~~~~~~~
      |                   avifAlloc
gd_avif.c:194:17: error: assignment to ‘void *’ from ‘int’ makes pointer from integer without a cast [-Werror=int-conversion]
  194 |         dataBuf = avifMalloc(size);
      |                 ^
```

morsssss:
Argh, sorry. Making a followup PR to fix this.

morsssss:
Please look at https://github.com/libgd/libgd/pull/702 ...

PR #709, opened 2021-07-15
Set GD_LIB to gd on mingw

activegregb:
Setting GD_LIB to libgd on MINGW builds results in
liblibgd.a and liblibgd.dll.a being built.


pierrejoye:
Thanks @activegregb ! Good catch :)

pierrejoye:
Merged, thanks again for the fix :)

Issue #697, opened 2021-04-30
Read out of bound in TGA files


me22bee:
hello,
[this issue](https://github.com/libgd/libgd/issues/247#issue-163166532) is showing a read out of bound for a corrupted TGA [test.txt](https://github.com/libgd/libgd/files/6407187/test.txt) which is patched by adding some checks for ` gdGetBuf`.

although the patch prevents occurring this vulnerability I saw that this function (` gdGetBuf`) is used in `read_header_tga` too which there is no check for its return value again.

I changed the header of the file which was used for the previous CVE-2016-6132. In fact, I changed the first byte to `ff` which is assigned to `tga->identsize`. 
[file(it is a tga, not a really a txt) ](https://github.com/libgd/libgd/files/6407191/test.txt)

```
...
tga->identsize = header[0];
...
gdGetBuf(tga->ident, tga->identsize, ctx);
...
```

when I run the test with this input file ASAN shows this:
```
AddressSanitizer:DEADLYSIGNAL
=================================================================
==12378==ERROR: AddressSanitizer: SEGV on unknown address 0x000000000000 (pc 0x7f685ae4a7a0 bp 0x7fff6bb8b740 sp 0x7fff6bb8b720 T0)
==12378==The signal is caused by a READ memory access.
==12378==Hint: address points to the zero page.
    #0 0x7f685ae4a79f in gdImageDestroy /tmp/libgd/src/gd.c:390
    #1 0x55c05b670432 in main /tmp/libgd/tests/tga/bug00084.c:11
    #2 0x7f685aa3709a in __libc_start_main ../csu/libc-start.c:308
    #3 0x55c05b670339 in _start (/tmp/libgd/build/Bin/test_tga_bug00084+0x2339)

AddressSanitizer can not provide additional info.
SUMMARY: AddressSanitizer: SEGV /tmp/libgd/src/gd.c:390 in gdImageDestroy
==12378==ABORTING
```

Is it showing another vulnerability?

me22bee:
any comment?

cmb69:
There might be a bug there, but who needs to read TGAs in … checks calendar … 2021?

pierrejoye:
@me22bee could you upload the full test case or do you use only bug00084.c?

me22bee:
> do you use only bug00084.c?

yes.

pierrejoye:
@me22bee could you attach the patched test file? Should be an easy fix. TGA is still used, indeed not for web but our targets go further than web devs :)

[...]

PR #711, opened 2021-07-19
fix read out-of-bounds in reading tga header file

me22bee:
Update src/gd_tga.c file for possible read out of bound in `gdGetBuf` call site.
Mentioned in this issue #697.


pierrejoye:
Thank you @me22bee  :)

Issue #700, opened 2021-05-26
gdImageGdPtr memory leak


meweez:
Hello,

I found that 'gdImageGdPtr' in gd_gd.c and 'gdImageWebpPtr' in gd_webp.c are similar functions for different picture formats. You have changed 'gdImageWebpPtr' because of CVE-2016-6912 (double free), So It seems that you need to change 'gdImageGdPtr' too.
I run two test files with ASAN, and the result is shown below. The test files are located in the 'tests/webp' folder.

Test1:
```
#include "gd.h"
#include "gdtest.h"


int main()
{
gdImagePtr im1, im2;
int size;

im1 = NULL;
im2 = gdImageGdPtr(im1, &size);
gdTestAssert(im2 == NULL);

gdImageDestroy(im1);

return gdNumFailures();
}
```
ASAN result:
```
AddressSanitizer:DEADLYSIGNAL
=================================================================
==4303==ERROR: AddressSanitizer: SEGV on unknown address 0x000000001c50 (pc 0x7f368bd976d8 bp 0x7ffd77fc7d70 sp 0x7ffd77fc7d60 T0)
==4303==The signal is caused by a READ memory access.
#0 0x7f368bd976d7 in _gdPutHeader /libgd/src/gd_gd.c:353
#1 0x7f368bd977d6 in _gdImageGd /libgd/src/gd_gd.c:370
#2 0x7f368bd97aec in gdImageGdPtr /libgd/src/gd_gd.c:403
#3 0x557763c46493 in main /libgd/tests/webp/gd_mytest.c:12
#4 0x7f368b68309a in __libc_start_main ../csu/libc-start.c:308
#5 0x557763c46339 in _start (/libgd/build/Bin/test_webp_gd_mytest+0x2339)

AddressSanitizer can not provide additional info.
SUMMARY: AddressSanitizer: SEGV /libgd/src/gd_gd.c:353 in _gdPutHeader
==4303==ABORTING
```
Test2:
```
#include "gd.h"
#include "gdtest.h"


int main()
{
gdImagePtr im1, im2;
FILE *fp;
int size;

fp = gdTestFileOpen2("webp", "bug_double_free.jpg");
gdTestAssert(fp != NULL);
im1 = gdImageCreateFromJpeg(fp);
gdTestAssert(im1 != NULL);
fclose(fp);


im2 = gdImageGdPtr(im1, &size);
gdTestAssert(im2 == NULL);

gdImageDestroy(im1);

return gdNumFailures();
[...]

PR #712, opened 2021-07-19
gdImageGdPtr memory leak

me22bee:
discussed in issue #700.

### DIFF
--- a/src/gd_gd.c
+++ b/src/gd_gd.c
@@ -400,8 +400,11 @@ BGD_DECLARE(void *) gdImageGdPtr (gdImagePtr im, int *size)
 	void *rv;
 	gdIOCtx *out = gdNewDynamicCtx (2048, NULL);
 	if (out == NULL) return NULL;
-	_gdImageGd (im, out);
-	rv = gdDPExtractData (out, size);
+	if (_gdImageGd (im, out)) {
+		rv = NULL;
+	} else {
+		rv = gdDPExtractData(out, size);
+	}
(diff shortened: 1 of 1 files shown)


pierrejoye:
Thank you :)

src/gd_gd.c
```diff
@@ -400,8 +400,11 @@ BGD_DECLARE(void *) gdImageGdPtr (gdImagePtr im, int *size)
 	void *rv;
 	gdIOCtx *out = gdNewDynamicCtx (2048, NULL);
 	if (out == NULL) return NULL;
-	_gdImageGd (im, out);
-	rv = gdDPExtractData (out, size);
+	if (_gdImageGd (im, out)) {
```
cmb69, line 403: @pierrejoye, `_gdImageGd()` is a void function; checking its return value doesn't make sense, and likely causes a compile failure or warning. I think we need to setup CI (building and running the tests) to catch such issues.

pierrejoye: Thanks @cmb69!

Damned, misread the signature, sorry :/

me22bee: sorry, my mistake :(
@pierrejoye can you please revert this?

pierrejoye: Reverted. Not that critical as the GD format is really really for dev/testing only. However open for an improved fix :)

Also, as @cmb69 mentioned, GD/GD2 formats are going to be deprecated and removed for 3.0

PR #713, opened 2021-07-19
gdImageGd2Ptr memory leak

me22bee:
We discussed it in issue #700.

There are some conditions in `_gdImageGd2` which lead it to fail1. So if it goes through these fail situations it means that the `out` isn't allocated correctly and calling the gdDPExtractData leads to double free.
so I add return value to `_gdImageGd2` and checked it in `gdImageGd2Ptr`.

### DIFF
--- a/src/gd_gd2.c
+++ b/src/gd_gd2.c
@@ -1,4 +1,4 @@
- /*
+/*
    * gd_gd2.c
    *
    * Implements the I/O and support for the GD2 format.
--- a/src/gd_gd2.c
+++ b/src/gd_gd2.c
@@ -1135,11 +1135,13 @@ BGD_DECLARE(void *) gdImageGd2Ptr (gdImagePtr im, int cs, int fmt, int *size)
 	void *rv;
 	gdIOCtx *out = gdNewDynamicCtx (2048, NULL);
 	if (out == NULL) return NULL;
+	
 	if (_gdImageGd2(im, out, cs, fmt)) {
 		rv = NULL;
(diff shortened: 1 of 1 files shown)


pierrejoye:
Thanks for the updated PR!

Could you push a change (no matter which) to see if all tests pass on all platforms? We finally have gh actions with Linux/MacOS/Windows (ARM and co coming too).

me22bee:
> Could you push a change (no matter which) to see if all tests pass on all platforms?

sorry, I don't understand that what to do now?

pierrejoye:
like commit and push a white space change to trigger the github actions :)

pierrejoye:
Seems some weird non printable bytes made it in :)

me22bee:
there is an error in " Windows 2019 x64"  with this message "libheif version found is too old".

pierrejoye:
let me fix it. Did you see the Linux one?

me22bee:
> Did you see the Linux one?

I fixed that and remove non-printable bytes.
Now there is a problem only with windows.

pierrejoye:
it should be good now. I re run the jobs :)

me22bee:
> it should be good now

Again the same error with libheif.

pierrejoye:
It should be ok. That HEIF issue works now and is not related to your PR.

me22bee:
> is not related to your PR.

Ok, so can you merge this PR?

pierrejoye:
thank you :)

I will merge it later today :)

PR #730, opened 2021-08-21
Version check with CMake for libheif

### LINKED ISSUE #729: HEIF builds fail with latest distros
We need to either update configure/cmake to check for the min version or wrap the 8bit conversion part with whatever was there before. Only recent HEIF have it and most if not all distros do not distribute these versions yet.

YakoYakoYokuYoku:
Performs a version check with the PkgConfig configuration or the libheif version (`heif_version.h`).

Fixes #729.


pierrejoye:
Looks good, better than the one I just committed :)

Thanks!

PR #735, opened 2021-08-24
fix #661, restore correct clamping, fixing alpha artifacts (these ones ar…

### LINKED ISSUE #661: imagecopyresampled() produce artifacts on transparent PNG
imagecopyresampled() produce many artifacts on transparent PNG.
I added a yellow layer to the output image to make the artifacts easier to see.
In the test code below a small resolution (thumbnail) is used on the output image, but similar artifacts also appear in higher resolutions.
Everything worked well on previous GD version (2.0 I think), but after the upgrade on v 2.2.5, the artifacts began to appear.
I was looking for a solution on the internet, but I did not find anything functional, it is probably a bug in the function imagecopyresampled().

Test code:
```
$img_src = imagecreatefrompng('source.png');
$img_dest = imagecreatetruecolor(225, 225);
imagealphablending($img_dest, false);
imagesavealpha($img_dest, true);
imagecopyresampled($img_dest, $img_src, 0, 0, 0, 0, 225, 225, 2000, 2000);
imagepng($img_dest, 'imagecopyresampled1.png');

```

Input image:
source

Output image:
imagecopyresample1

pierrejoye:
This fixes the clamping and rounding bug which creates artifact. The most visible ones are with alpha, black spots or incorrect colors/alpha when all are max values in the interpolations areas.


src/gd.c
```diff
@@ -3539,20 +3539,12 @@ BGD_DECLARE(void) gdImageCopyResampled (gdImagePtr dst,
...
-			}
-			if (blue > 255.0f) {
-				blue = 255.0;
-			}
-			if (alpha > gdAlphaMax) {
-				alpha = gdAlphaMax;
-			}
-			gdImageSetPixel(dst, x, y, gdTrueColorAlpha ((int) red, (int) green, (int) blue, (int) alpha));
+			/* Round up closest next channel value and clamp to max channel value */
+			red = red > 255.5 ? 255 : red+0.5;
```
cmb69, line 3543:
```suggestion
			red = red >= 255.5 ? 255 : red+0.5;
```

pierrejoye: both produces the same:

https://godbolt.org/z/ofr5znobc 

Basically the compiler realizes what we do, and ensure the val+0.5 clamping up:)

pierrejoye: Also good thing is we will be able to unroll this loop (no worry, not for 2.3 and for what I see gcc/clang and VC nicely parallelize it :)

cmb69: Try:
````.c
#include <stdio.h>

int main() {
    float red = 255.5;
    printf("%d\n", (int) (red > 255.5 ? 255 : red+0.5));
    printf("%d\n", (int) (red >= 255.5 ? 255 : red+0.5));
    return 0;
}
````
output:
````
256
255
````

pierrejoye: Oh! damned you right :)

pierrejoye: done :)

I will add a test for that specific case using only the gdTrueColorAlpha.

PR #737, opened 2021-08-25
Fix #405, -1 aims to reset the transparent color; refactor a bit; add test to existing test for -1

### LINKED ISSUE #405: Transparent color can't be reset
`gdImageColorTransparent(im, -1)` is supposed to reset the transparent color index. That doesn't work anymore since 118fc70, because the function bails out early, if `color < 0` even though the following code explicitly deals with `color != -1`. The commit message refers to [PHP bug 72519](https://bugs.php.net/bug.php?id=72519), but actually [PHP bug 72512](https://bugs.php.net/bug.php?id=72512) is meant. Still, it appears that the fix is not quite correct.

pierrejoye:
Fix #405, -1 aims to reset the transparent color; refactor a bit; add test to existing test for -1

### DIFF
--- a/src/gd.c
+++ b/src/gd.c
@@ -895,19 +895,24 @@ BGD_DECLARE(void) gdImageColorDeallocate (gdImagePtr im, int color)
  */
 BGD_DECLARE(void) gdImageColorTransparent (gdImagePtr im, int color)
 {
-	if (color < 0) {
+	// Reset ::transparent
+	if (color == -1) {
+		im->transparent = -1;
+	}
+	if (color < -1) {
+		return;
+	}
+	if (im->trueColor) {
+		im->transparent = color;
 		return;
 	}
 
-	if (!im->trueColor) {
-		if (color >= gdMaxColors) {
-			return;
-		}
-		if (im->transparent != -1) {
-			im->alpha[im->transparent] = gdAlphaOpaque;
-		}
-		im->alpha[color] = gdAlphaTransparent;
+	// Palette Image
+	if (color >= gdMaxColors) {
+		return;
 	}
+	im->alpha[im->transparent] = gdAlphaOpaque;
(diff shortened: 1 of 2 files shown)


src/gd.c
```diff
@@ -895,19 +895,24 @@ BGD_DECLARE(void) gdImageColorDeallocate (gdImagePtr im, int color)
...
+	if (color == -1) {
+		im->transparent = -1;
+	}
+	if (color < -1) {
+		return;
+	}
+	if (im->trueColor) {
+		im->transparent = color;
 		return;
 	}
```
cmb69, line 908: This looks superfluous; at least previously, we didn't do anything to truecolor images.

src/gd.c
```diff
@@ -895,19 +895,24 @@ BGD_DECLARE(void) gdImageColorDeallocate (gdImagePtr im, int color)
...
-		if (im->transparent != -1) {
-			im->alpha[im->transparent] = gdAlphaOpaque;
-		}
-		im->alpha[color] = gdAlphaTransparent;
+	// Palette Image
+	if (color >= gdMaxColors) {
+		return;
 	}
+	im->alpha[im->transparent] = gdAlphaOpaque;
+	im->alpha[color] = gdAlphaTransparent;
```
cmb69, line 915: This doesn't look right. `color` may be `-1` here.

pierrejoye: if (color == -1)  is first in the function, can't be -1 there.

src/gd.c
pierrejoye: For clarity and simplify paths. Can be removed but it provides clarity and clear return here. I prefer that that what we had before. Thoughts?

src/gd.c
pierrejoye: Damned, return missing there. Which is the whole point.... good catch :)

PR #751, opened 2021-09-05
Close #334: gdColorMatch(): doubtful algorithm

### LINKED ISSUE #334: gdColorMatch(): doubtful algorithm
I've noticed that the behavior of PHP's bundled libgd differs from libgd wrt. to imagecropauto() which is used as wrapper for gdImageCropThreshold() and filed a [respective ticket](https://bugs.php.net/bug.php?id=73291).

However, none of both algorithms appear to be reasonable. I would suggest to calculate the Euclidean distance of the percentages of the channel saturation and compare that to the threshold, i.e. something like:
````
int gdColorMatch(gdImagePtr im, int col1, int col2, float threshold)
{
	const float dr = (gdImageRed(im, col1) - gdImageRed(im, col2)) / (float) gdRedMax;
	const float dg = (gdImageGreen(im, col1) - gdImageGreen(im, col2)) / (float) gdGreenMax;
	const float db = (gdImageBlue(im, col1) - gdImageBlue(im, col2)) / (float) gdBlueMax;
	const float da = (gdImageAlpha(im, col1) - gdImageAlpha(im, col2)) / (float) gdAlphaMax;
	const float dist = sqrt(dr * dr + dg * dg + db * db + da * da);

	return dist < threshold;
}
````
As the values of `dist` range from 0..2 with this algorithm, `threshold` might be doubled to fit the percentage range.

Of course, changing the algorithm in such a way would cause a BC break, so this probably would have to wait for GD 3.

### PR DESCRIPTION
**Pull request by cmb69**
We change the algorithm of `gdColorMatch()` to calculate the Euclidean
distance of the color component values, what may require to change some
thresholds in existing code to match the new algorithm, as can be seen
by the need to adapt gdimagecolorreplace.c.

### DIFF
--- a/src/gd_color.c
+++ b/src/gd_color.c
@@ -7,6 +7,10 @@
 #include <math.h>
 
 /**
+ * Checks whether two given colors match according to a given threshold in
+ * range [0..100]. dist is the euclidian distance of the color channels in
+ * range [0..2].
+ *
  * The threshold method works relatively well but it can be improved.
  * Maybe L*a*b* and Delta-E will give better results (and a better
  * granularity).
--- a/src/gd_color.c
+++ b/src/gd_color.c
@@ -4,6 +4,7 @@
 
 #include "gd.h"
 #include "gd_color.h"
+#include <math.h>
 
 /**
  * The threshold method works relatively well but it can be improved.
@@ -12,13 +13,13 @@
  */
 int gdColorMatch(gdImagePtr im, int col1, int col2, float threshold)
 {
-	const int dr = gdImageRed(im, col1) - gdImageRed(im, col2);
-	const int dg = gdImageGreen(im, col1) - gdImageGreen(im, col2);
-	const int db = gdImageBlue(im, col1) - gdImageBlue(im, col2);
-	const int da = gdImageAlpha(im, col1) - gdImageAlpha(im, col2);
-	const int dist = dr * dr + dg * dg + db * db + da * da;
+    const float dr = (gdImageRed(im, col1) - gdImageRed(im, col2)) / (float) gdRedMax;
+    const float dg = (gdImageGreen(im, col1) - gdImageGreen(im, col2)) / (float) gdGreenMax;
+    const float db = (gdImageBlue(im, col1) - gdImageBlue(im, col2)) / (float) gdBlueMax;
+    const float da = (gdImageAlpha(im, col1) - gdImageAlpha(im, col2)) / (float) gdAlphaMax;
+    const float dist = sqrtf(dr * dr + dg * dg + db * db + da * da);
 
-	return 100.0 * dist < threshold * 195075.0;
+    return 100 * dist < 2 * threshold;
 }
 
 /*
--- a/tests/gdimagecolorreplace/gdimagecolorreplace.c
+++ b/tests/gdimagecolorreplace/gdimagecolorreplace.c
@@ -119,7 +119,7 @@ static void run_tests(gdImagePtr im, int *error)
 	gdImageSetPixel(im, 3, 2, seashell)
 
 	INITIALIZE_IMAGE();
-	n = gdImageColorReplaceThreshold(im, white, yellow, 2.0);
+	n = gdImageColorReplaceThreshold(im, white, yellow, 10.0);
 	CHECK_VALUE(n, 9);
 	CHECK_PIXEL(0, 0, black);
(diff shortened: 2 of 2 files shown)

### CODE REVIEW

**Comment by pierrejoye** [comment]
Looks good to me, Thanks @cmb69 !

**Comment by pierrejoye** [comment]
I think we need a 2.3.3 release, then most likely a 2.3.4. Once 2.3.3 is out (I will try today or later this week), we can branch GD_2_2 and do only sec fixes there. Master will be 2.4 then and merging this PR as well.

Thoughts?

**Comment by cmb69** [comment]
It seems that GD-2.3 has already been branched. Any objections to merge this into master?

File: src/gd_color.c [hunk 1]
```diff
@@ -4,6 +4,7 @@
 
 #include "gd.h"
 #include "gd_color.h"
+#include <math.h>
 
 /**
  * The threshold method works relatively well but it can be improved.
```
vapier, line 10: please update the comment docs to match what the code is doing

vapier (changes requested):
needs a rebase

src/gd_color.c
cmb69: Done. Note though that the mentioned L\*A\*B and Delta-E variants are orthogonal to my changes. These are about calculating the difference in different color space than (s)RGB (what might indeed be good idea).

My change is only about improving the somewhat arbitrary threshold calculation; `195075` is `3 * 255**2` what would make some sense only if there are no alpha channel differences. Adjusting the old value to `211204` would basically work (identical results for threshold 0 and 100), but has a different distribution (due to sqrt()). For instance:

col1|col2|old|adjusted|new
----|----|--:|------:|--:
0x00000000|0x00000000|0.0|0.0|0.0
0x00000000|0x20404040|6.8|6.3|25.1
0x00000000|0x40808080|27.3|25.2|50.2
0x00000000|0x60c0c0c0|61.4|56.7|75.4
0x00000000|0x7fffffff|108.3|100.0|100.0

The `new` column shows what one would expect (0x40808080 is about half-way off, etc.) The last row shows how bad the old (non-adjusted) calculation was, since it yields a threshold > 100 (although that's not supposed to happen).

src/gd_color.c
```diff
@@ -4,21 +4,26 @@
...
 {
-	const int dr = gdImageRed(im, col1) - gdImageRed(im, col2);
-	const int dg = gdImageGreen(im, col1) - gdImageGreen(im, col2);
-	const int db = gdImageBlue(im, col1) - gdImageBlue(im, col2);
-	const int da = gdImageAlpha(im, col1) - gdImageAlpha(im, col2);
-	const int dist = dr * dr + dg * dg + db * db + da * da;
+    const float dr = (gdImageRed(im, col1) - gdImageRed(im, col2)) / (float) gdRedMax;
+    const float dg = (gdImageGreen(im, col1) - gdImageGreen(im, col2)) / (float) gdGreenMax;
+    const float db = (gdImageBlue(im, col1) - gdImageBlue(im, col2)) / (float) gdBlueMax;
+    const float da = (gdImageAlpha(im, col1) - gdImageAlpha(im, col2)) / (float) gdAlphaMax;
```
cmb69, line 23: I've just learned that [float division might be particularly slow](http://ithare.com/infographics-operation-costs-in-cpu-clock-cycles/). Should check whether this is an issue (`gdColorMatch()` is called in a tight loop).

pierrejoye:
Finally (!!!) some time to look at some PRs and issues here.

The main issue with the current (private) gdColorMatch used for crop Ops is primarily caused by:

1. I treat alpha like rgb, with a max of 255, that's not the case, it is 127
2. I compared the squared distance against the linear threshold, very hard to control, or creepy

thoughts on:

```c
#define GD_MAX_COLOR_DIST_SQ 260100
int gdColorMatch(gdImagePtr im, int col1, int col2, float threshold)
{
    const int dr = gdImageRed(im, col1)   - gdImageRed(im, col2);
    const int dg = gdImageGreen(im, col1) - gdImageGreen(im, col2);
    const int db = gdImageBlue(im, col1)  - gdImageBlue(im, col2);
    const int da = (gdImageAlpha(im, col1) - gdImageAlpha(im, col2)) * 2;
    const int dist = dr*dr + dg*dg + db*db + da*da;

    const float t = threshold / 100.0f;
    return dist <= (int)(t * t * GD_MAX_COLOR_DIST_SQ);
}
```

added a constant too to make it more clear.

pierrejoye:
fixed in master, thanks :)

PR #776, opened 2021-09-20
fix help string (as off by default)

remicollet:
As off by default since c6af75565ad00732d71064edddfb7a8ea3e0b26b

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,10 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 
 ## [2.3.3] - 2021-09-12
 
+### Changed
+
+- support for the legacy/testing gd image formats is disabled by default
+
 ### Fixed
 
 - [#759](https://github.com/libgd/libgd/issues/759) update cmake to generate config.h in the build dir
--- a/configure.ac
+++ b/configure.ac
@@ -104,7 +104,7 @@ AX_OPENMP()
 dnl We should default this to off in future releases.
 AC_MSG_CHECKING([whether to support gd image formats])
 AC_ARG_ENABLE([gd-formats],
-  [AS_HELP_STRING([--disable-gd-formats], [Disable support for the legacy/testing gd image formats])],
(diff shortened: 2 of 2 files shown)


pierrejoye:
Looks good to me

pierrejoye (approved):
Looks good to me

vapier:
i'd leave these enabled for GD-2.3 and turn them off in GD-2.4/GD-3 (whichever comes next).  seems inappropriate to make these kind of backwards incompatible changes in a "patch" release.  i think semver advocates would argue it shouldn't be in GD-2.4 either, but we haven't exactly been following semver.

pierrejoye:
does semver applies to configure options?


Also we should strictly follow semver in 3.4 onwards. Including not adding
new features in x.y.z+1 :)

vapier:
> does semver applies to configure options?

i don't think semver pedantically gets that detailed, but imo the spirit is there

> Also we should strictly follow semver in 3.4 onwards. Including not adding

you mean GD-3 ?  i feel like GD-2 has been historically the same for us.

pierrejoye:
sorry, I mean GD-2.3 should not introduce abi/api breaks neither new
features.

GD-2.4, no API breaks but can have ABI changes like the need to rebuild.

GD-3 let change what we need to change

makes sense?

vapier:
sounds good as long as we write down our agreement in the README :)

pierrejoye:
absolutely!

also we do mention semver in our changelog file already ;-)

pierrejoye:
@remicollet possible to rebase it to master for 2.4 (master)? 

Unless you can say that it is ok to do it for 2.3.x :)

thesamesam:
Not read too much into it but isn't it _already_ default off in 2.3.3? Someone noticed downstream in [Gentoo](https://bugs.gentoo.org/815061) and on the original [bug](https://github.com/libgd/libgd/issues/428#issuecomment-922918620).

vapier:
> Not read too much into it but isn't it already default off in 2.3.3? 

yes, that's the discussion we've been having here

remicollet:
See
pr #779 for 2.3.x
pr #780 for master

PR #787, opened 2021-10-22
Improve image crop performance

Lerchensporn:
I would like to suggest some small performance tweaks to the image cropping. Most significantly, one can start the loop at y = crop.y instead of 0 when cropping horizontally, since these corners were already handled in the vertical crop. Using goto only makes a small performance difference, but I think it also makes the routine easier to read. The PHP code below I used to test the performance difference on some test image.

```
<?php

$imageblob = file_get_contents('uncropped.png');
$bitmap = imagecreatefromstring($imageblob);
$time = microtime(true);
for ($i = 0; $i < 100; ++$i) {
	$cropped = imagecropauto($bitmap, IMG_CROP_THRESHOLD, 0.5, 0xFFFFFF);
	imagedestroy($cropped);
}
print(microtime(true) - $time);

#$cropped = imagecropauto($bitmap, IMG_CROP_THRESHOLD, 0.5, 0xFFFFFF);
#imagepng($cropped, 'cropped.png');

# Unmodified cropping algorithm: 2.50 s
# Removed duplicate traversal of crop region: 2.33 s
# + Replaced division by multiplication: 2.26 s
# + Replaced match by goto: 2.22 s
# This is an 11% improvement.
```

### DIFF
--- a/src/gd_color.c
+++ b/src/gd_color.c
@@ -18,7 +18,7 @@ int gdColorMatch(gdImagePtr im, int col1, int col2, float threshold)
 	const int da = gdImageAlpha(im, col1) - gdImageAlpha(im, col2);
 	const int dist = dr * dr + dg * dg + db * db + da * da;
 
-	return (100.0 * dist / 195075) < threshold;
+	return 100.0 * dist < threshold * 195075.0;
 }
 
 /*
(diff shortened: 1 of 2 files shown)


pierrejoye:
thanks for this nice PR :)

Any objection to merge?

vapier:
sure, let's go for it.  thx @Lerchensporn.

Issue #788, opened 2021-10-28
HEIF output is distorted for certain resolutions


aw32:
For certain resolutions the HEIF output is distorted.

```
#include "gd.h"

int main (int argc, char *argv[]) {
	FILE *fp;
	gdImagePtr in;

	fp = fopen(argv[1], "rb");
	in = gdImageCreateFromPng(fp);
	fclose(fp);

	fp = fopen(argv[2], "wb");
        gdImageHeif(in, fp);
	fclose(fp);

	gdImageDestroy(in);
	return 0;
}
```
Test:
```
$ gcc -o test test.c -lgd
$ ./test in.png out.heif
$ file in.png
in.png: PNG image data, 333 x 250, 8-bit/color RGB, non-interlaced
$ file out.heif
out.heif: ISO Media, HEIF Image HEVC Main or Main Still Picture Profile
```

Input file with resolution 333x250:
![in](https://user-images.githubusercontent.com/18468487/139276265-8b733576-aaea-4485-9d5d-0f1c48c8a488.png)
Output file (converted to PNG):
![out](https://user-images.githubusercontent.com/18468487/139276643-58df8ec5-af1e-4a61-8bd0-b84295f23ce1.png)
Original output file:
[out.heif.zip](https://github.com/libgd/libgd/files/7435063/out.heif.zip)

The image should look similar to the input image.
However, it looks distorted.
Doing the same with `convert in.png out_im.heif` (Image Magick) looks ok.
For 332x249 it looks ok.
For 334x251 it also looks distorted.

- OS: Arch Linux
- GD: Tested with Package version 2.3.3-3 and Git b614af3
- Libheif: 1.12.0-1

pierrejoye:
Do you see the same using the libheif conversion tools?

aw32:
Using `heif-enc` the result is normal.
Can you reproduce the issue?

cmb69:
Does heif require padding:
https://github.com/libgd/libgd/blob/9af6d1254f2be57bf5179f2b02a0e90d3806570d/src/gd_heif.c#L360-L377

pierrejoye:
if the heif Image data has it, yes :)

cmb69:
[...]

PR #801, opened 2022-01-21
#788 fix bug in HEIF usage, stride is require

pierrejoye:
Fix HEIF when stride is different from actual used row length. As in, actually use stride.


tests/heif/bug788.c
```diff
@@ -0,0 +1,43 @@
...
+
+#include "gd.h"
+#include "gdtest.h"
+
+#include <libheif/heif.h>
+
+int main () {
+	FILE *fp;
+	gdImagePtr in;
+    gdImagePtr dst;
```
cmb69, line 13: Nit: mixed indentation (spaces vs. tabs) (also further down in the test file)

cmb69 (commented):
Thank you! LGTM except for the minor indentation issue in the test case.

tests/heif/bug788.c
pierrejoye: fixed WS, thanks :)

PR #823, opened 2022-02-04
Prevent libpng warning when loading interlaced PNG

developerbmw:
Currently a warning is being generated when reading interlaced PNGs due to libpng being used incorrectly.

Refer to https://github.com/glennrp/libpng/blob/libpng16/pngread.c#L727 for the source of this warning.

This PR enables the interlacing transform prior to calling `png_read_update_info()` to prevent this warning. There is no change in functionality since libpng automatically does this after emitting the warning anyway.

### DIFF
--- a/src/gd_png.c
+++ b/src/gd_png.c
@@ -415,6 +415,11 @@ BGD_DECLARE(gdImagePtr) gdImageCreateFromPngCtx (gdIOCtx * infile)
 		goto error;
 	}
 
+	/* enable the interlace transform if supported */
+#ifdef PNG_READ_INTERLACING_SUPPORTED
+	(void)png_set_interlace_handling(png_ptr);
+#endif
+
 	png_read_update_info (png_ptr, info_ptr);
 
 	/* allocate space for the PNG image data */


pierrejoye:
Hello @developerbmw 

Thanks for this found and PR :)

src/gd_png.c
```diff
@@ -414,6 +414,9 @@ BGD_DECLARE(gdImagePtr) gdImageCreateFromPngCtx (gdIOCtx * infile)
 		gd_error("gd-png color_type is unknown: %d\n", color_type);
 		goto error;
 	}
+	
+	/* interlacing handled by libpng */
+	png_set_interlace_handling(png_ptr);
```
pierrejoye, line 419: Given the function:

```C
png_set_interlace_handling(png_structrp png_ptr)
{
   png_debug(1, "in png_set_interlace handling");

   if (png_ptr != 0 && png_ptr->interlaced != 0)
   {
      png_ptr->transformations |= PNG_INTERLACE;
      return (7);
   }

   return (1);
}
```

Either we do not care about the image being interlaced or not at this stage, or we do and we should handle the return value.

If we don't, the could can be changed to:

```C
(void)png_set_interlace_handling(png_ptr);
```

Otherwise that function returns the number of passes afair.

If I am not mistaken, libpng can also be built without interlace support, meaning #ifdef PNG_READ_INTERLACING_SUPPORTED needs to be used here.

developerbmw:
Thanks @pierrejoye, good points.

I believe we don't care about the return value since we just want the side effect of the interlacing transform being enabled, so I've added the void cast. The return value of this function is handled later on within libpng at https://github.com/glennrp/libpng/blob/libpng16/pngread.c#L712-L737

I've added the PNG_READ_INTERLACING_SUPPORTED check as suggested.

pierrejoye:
thanks for updating the PR. I will merge it later today :)

PR #832, opened 2022-06-10
Fix #831: gdImageAvif memory leak

### LINKED ISSUE #831: gdImageAvif memory leak
When using `gdImageAvifPtrEx` or `gdImageAvif` memory is leaked.
Example:
```
#include "gd.h"
int main (int argc, char *argv[]) {
    FILE *fp;
    gdImagePtr in; 
    fp = fopen(argv[1], "rb");
    in = gdImageCreateFromPng(fp);
    fclose(fp);
    fp = fopen(argv[2], "wb");
        gdImageAvif(in, fp);
    fclose(fp);
    gdImageDestroy(in);
    return 0;
}
```

Valgrind shows lost memory:
```
$ valgrind --leak-check=full ./testavif test.png test.avif
[..]
208,450 (200 direct, 208,250 indirect) bytes in 1 blocks are definitely lost in loss record 16 of 16
 at 0x4845899: malloc (in /usr/lib/valgrind/vgpreload_memcheck-amd64-linux.so)
 by 0x50EE3FD: avifAlloc (in /usr/lib/libavif.so.14.0.1)
 by 0x50EE432: avifImageCreate (in /usr/lib/libavif.so.14.0.1)
 by 0x4870584: _gdImageAvifCtx (gd_avif.c:529)
 by 0x4870CF3: gdImageAvifEx (gd_avif.c:612)
 by 0x1091F8: main (in testavif)
```

 - OS: Arch
 - Version libgd git 7efcf4f9
 - libavif 0.10.1-1

It seems the following avifImageCreate is never freed:
https://github.com/libgd/libgd/blob/7efcf4f935d0466b6ada44bed62d729cd2cddbc4/src/gd_avif.c#L529

cmb69:
First, we must not forget to call `avifImageDestroy()` when we're
finished with the image.

Then we also need to cater to the allocated `dataBuf`.  To keep track
of that, we "extend" `avifIO` as `avifIOCtxReader`.  To simplify, and
to avoid unnecessary allocations, we use `realloc()`.  To better fit
with GD, we also use the GD memory allocation functions instead of the
ones provided by libavif.

### DIFF
--- a/src/gd_avif.c
+++ b/src/gd_avif.c
@@ -159,6 +159,11 @@ static avifBool isAvifError(avifResult result, const char *msg) {
 }
 
 
+typedef struct avifIOCtxReader {
+	avifIO io; // this must be the first member for easy casting to avifIO*
+	avifROData rodata;
+} avifIOCtxReader;
+
 /*
 	<readfromCtx> implements the avifIOReadFunc interface by calling the relevant functions
 	in the gdIOCtx. Our logic is inspired by avifIOMemoryReaderRead() and avifIOFileReaderRead().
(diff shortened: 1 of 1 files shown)


cmb69:
Any objections to merge this?

PR #833, opened 2022-08-01
Fix #790, don't rotate glyph positions with Raqm 0.9.0+

### LINKED ISSUE #790: Rotated TTF text not rendered correctly using raqm
**Describe the bug**
Using PHP to produce an image, when the text is rotated 90 degrees (i.e. vertical) the text is rendered with the characters on top of one another.

This only appears to be a problem with True Type fonts.

**To Reproduce**
```
<?php
header('Content-Type: image/png');
$im = imagecreatetruecolor(30, 300);
$white = imagecolorallocate($im, 255, 255, 255);
$grey = imagecolorallocate($im, 128, 128, 128);
$black = imagecolorallocate($im, 0, 0, 0);
imagefilledrectangle($im, 0, 0, 29, 299, $white);
$text = 'Testing...';
$font = '/usr/share/fonts/msttcorefonts/arial.ttf';
imagettftext($im, 20, 90, 26, 221, $grey, $font, $text);
imagettftext($im, 20, 90, 25, 220, $black, $font, $text);
imagepng($im);
?>
```

Using this code produces the problem.

**Expected behavior**
A white box with "Testing....." displayed vertically read from bottom to top.

**Actual results**
A white box with the characters on top of one another. See https://www.while.org.uk/try.png

 - OS: Fedora Linux
 - Version 35
- GD V2.3.2 (also tried with GD2.3.3)
- Freetype 2.11.0
**Additional context**
This only appears to be a recent problem when upgrading from Fedora 34 to Fedora 35. I initially reported it to the PHP development team who said it was an upstream problem and to report it here.

khaledhosny:
Raqm will apply the transformation matrix set on the FT_Face itself.


cmb69:
Thank you! This generally makes sense to me, but I wonder about that segfault in the Windows MinGW 64 CI; probably unrelated to this PR, but should still be investigated.

khaledhosny:
> but I wonder about that segfault in the Windows MinGW 64 CI; probably unrelated to this PR, but should still be investigated.

Seems to be transient failure. I tried reverting this commit and all builds passed, and they also passed after restoring the commit.

cmb69:
Thanks for checking, @khaledhosny!

@remicollet, @pierrejoye, any objections to merging this?

PR #835, opened 2022-08-23
Fix deprecared function prototypes

kraj:
Fixes following errors:
error: a function definition without a prototype is deprecated in all versions of C and is not supported in C2x [-Werror,-Wdeprecated-non-prototype]

Upstream-Status: Pending
Signed-off-by: Khem Raj <raj.khem@gmail.com>

### DIFF
--- a/src/gd_nnquant.c
+++ b/src/gd_nnquant.c
@@ -112,12 +112,7 @@ typedef struct {
 
 /* Initialise network in range (0,0,0,0) to (255,255,255,255) and set parameters
    ----------------------------------------------------------------------- */
-static void initnet(nnq, thepic, len, sample, colours)
-nn_quant *nnq;
-unsigned char *thepic;
-int len;
-int sample;
-int colours;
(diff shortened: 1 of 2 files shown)


src/gd_nnquant.c
```diff
@@ -362,9 +351,7 @@ register int al,b,g,r;
 /* Move neuron i towards biased (a,b,g,r) by factor alpha
 	 ---------------------------------------------------- */
 
-static void altersingle(nnq, alpha,i,al,b,g,r)
-nn_quant *nnq;
-register int alpha,i,al,b,g,r;
+static void altersingle(nn_quant *nnq, int alpha, int i,int al, int b, int g, int r)
```
cmb69, line 354:
```suggestion
static void altersingle(nn_quant *nnq, int alpha, int i, int al, int b, int g, int r)
```

cmb69 (commented):
Oh, that should have been changed long ago. So thank you! Please fix the minor issues commented below; otherwise this looks good.

cmb69:
Thank you!

PR #836, opened 2022-08-23
WebPDecodeARGB()'s return should be WebPFree()ed.

Maeyanie:
According to Google's spec, "The code that calls any of these functions must delete the data buffer (uint8_t*) returned by these functions with WebPFree()."
Using free() instead was causing heap corruption crashes for me (Windows MSVC 2019 compile)

### DIFF
--- a/src/gd_webp.c
+++ b/src/gd_webp.c
@@ -229,7 +229,7 @@ static int _gdImageWebpCtx (gdImagePtr im, gdIOCtx * outfile, int quality)
 	}
 
 	int res = gdPutBuf(out, out_size, outfile);
-	free(out);
+	WebPFree(out);
 	if (res != out_size) {
 		gd_error("gd-webp write error\n");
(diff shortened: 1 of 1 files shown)


vapier:
@morsssss

cmb69:
Thank you! This looks correct, but shouldn't we change the following as well:
https://github.com/libgd/libgd/blob/dbe3001a125003adb553b9e1cd9443353b1df56a/src/gd_webp.c#L232

Maeyanie:
Yes, looks like that should be changed too, good catch. My particular application only used the decoder half, so I didn't run into that one.

cmb69 (approved):
Thank you! This looks good to me (ref: https://developers.google.com/speed/webp/docs/api#simple_decoding_api). The failing test is unrelated to this PR.

Maybe we should wait a while for others to comment.

PR #848, opened 2022-10-27
Fix #847: enable back GD_BICUBIC* interpolation methods

### LINKED ISSUE #847: gdImageSetInterpolationMethod ignores GD_BICUBIC*
**Describe the bug**
`gdImageSetInterpolationMethod` does not longer accept `GD_BICUBIC` or `GD_BICUBIC_FIXED` constants.

As such, the following code snippet do not scale image with bicubic algorithms:

```
if (gdImageSetInterpolationMethod(im, algo)) {
  scaled = gdImageScale(im, width2, height2);
}
```

The change was introduced in 672475e540da84100252e720291c63b30e028d06.

**To Reproduce**
```c
#include "gd.h"
//#include "gdtest.h

int main()
{
  gdImagePtr im;
  int status;
  im = gdImageCreate(1, 1);
  gdImageSetInterpolationMethod(im, GD_BICUBIC_FIXED);
  if (gdImageGetInterpolationMethod(im) == GD_BICUBIC_FIXED) {
    status = 0;
  } else {
    status = 1;
  }
  // if (status > 0) {
  //   gdTestErrorMsg("expecting success return value\n");
  // }
  gdImageDestroy(im);
  return status;
}
```

**Expected behavior**
`gdImageSetInterpolationMethod` should set the mode and return 1 **OR** the ChangeLog should mention that `GD_BICUBIC*` are removed and ignored.

**Actual results**
`gdImageSetInterpolationMethod` returns 0, and code guarding a call to `gdImageScale` with `if (gdImageSetInterpolationMethod(im, GD_BICUBIC)) { … }` is silently ignored, this is what PHP-gd does: https://github.com/php/php-src/blob/master/ext/gd/gd.c#L3643

 - OS: Debian testing (with debian's libgd) or any Debian using the [deb.sury.org](https://github.com/oerdnj/deb.sury.org) repository with php-gd.
 - Version 2.3.3 
- All libraries used by GD and their respective version : not relevant here

**Additional context**
`IMG_BICUBIC` and `IMG_BICUBIC_FIXED` are mentioned and promoted on php-gd documentation for `imagescale` : https://www.php.net/manual/en/function.imagescale.php

We have some legacy code that use `IMG_BICUBIC_FIXED` and suddenly stop working after upgrading from debian 9 to debian 10 with php-sury repository which ships libgd3 2.3.3 instead of 2.2.5.

fulax:
I made the PR for the GD-2.3 branch, as 2.3.3 is the version with the regression.

### DIFF
--- a/tests/gdimagesetinterpolationmethod/github_bug_00847.c
+++ b/tests/gdimagesetinterpolationmethod/github_bug_00847.c
@@ -1,0 +1,12 @@
+#include "gd.h"
+#include "gdtest.h"
+
+int main()
+{
+  gdImagePtr im;
+  im = gdImageCreate(1, 1);
+  gdTestAssert(gdImageSetInterpolationMethod(im, GD_BICUBIC_FIXED) != 0);
+  gdTestAssert(gdImageGetInterpolationMethod(im) == GD_BICUBIC_FIXED);
+  gdImageDestroy(im);
+  return gdNumFailures();
+}
--- a/src/gd_interpolation.c
+++ b/src/gd_interpolation.c
@@ -2257,6 +2257,11 @@ BGD_DECLARE(int) gdImageSetInterpolationMethod(gdImagePtr im, gdInterpolationMet
 		case GD_BESSEL:
 			im->interpolation = filter_bessel;
 			break;
+		case GD_BICUBIC_FIXED:
+		case GD_BICUBIC:
+			/* no interpolation as gdImageScale calls a dedicated function */
+			im->interpolation = NULL;
+			break;
 		case GD_BLACKMAN:
 			im->interpolation = filter_blackman;
 			break;
--- a/tests/gdimagesetinterpolationmethod/.gitignore
+++ b/tests/gdimagesetinterpolationmethod/.gitignore
@@ -1,1 +1,2 @@
 /github_bug_00584
+/github_bug_00847
--- a/tests/gdimagesetinterpolationmethod/CMakeLists.txt
+++ b/tests/gdimagesetinterpolationmethod/CMakeLists.txt
@@ -1,5 +1,6 @@
 LIST(APPEND TESTS_FILES
 	github_bug_00584
+	github_bug_00847
 )
 
 ADD_GD_TESTS()
--- a/tests/gdimagesetinterpolationmethod/Makemodule.am
+++ b/tests/gdimagesetinterpolationmethod/Makemodule.am
@@ -1,5 +1,6 @@
 libgd_test_programs += \
-	gdimagesetinterpolationmethod/github_bug_00584
+	gdimagesetinterpolationmethod/github_bug_00584 \
+	gdimagesetinterpolationmethod/github_bug_00847
 
 EXTRA_DIST += \
 	gdimagesetinterpolationmethod/CMakeLists.txt


tests/gdimagesetinterpolationmethod/github_bug_00847.c
```diff
@@ -0,0 +1,20 @@
...
+  im = gdImageCreate(1, 1);
+  gdImageSetInterpolationMethod(im, GD_BICUBIC_FIXED);
+  if (gdImageGetInterpolationMethod(im) == GD_BICUBIC_FIXED) {
+    status = 0;
+  } else {
+    status = 1;
+  }
+  if (status > 0) {
+    gdTestErrorMsg("expecting success return value\n");
+  }
```
cmb69, line 17: This looks unnecessarily complicated. Instead we could just do:
```suggestion
  gdTestAssert(gdImageGetInterpolationMethod(im) == GD_BICUBIC_FIXED);
```

tests/gdimagesetinterpolationmethod/github_bug_00847.c
```diff
@@ -0,0 +1,20 @@
...
+  gdImageDestroy(im);
+  return status;
```
cmb69, line 19: For the above modifications to have effect, we should do:
```suggestion
  return gdNumFailures();
```

tests/gdimagesetinterpolationmethod/github_bug_00847.c
```diff
@@ -0,0 +1,20 @@
+#include "gd.h"
+#include "gdtest.h"
+
+int main()
+{
+  gdImagePtr im;
+  int status;
```
cmb69, line 7: This line is no longer necessary, if the other changes are done:
```suggestion
```

cmb69 (commented):
Thank you for the PR! The actual fix looks fine to me; like I said, I have some suggestions for the accompanying test case.

cmb69:
Hmm, the macOS failure is unrelated to this PR, but should be catered to nonetheless (not necessarily by you, @fulax). Apparently, we have the libpng version hard-coded to 1.6.37, but now we need to use 1.6.38. Would anybody know whether we can get rid of the actual version number? Otherwise, I'd make sense to use a variable for the version (so it needs to be changed only in one place).

fulax:
> Thank you for the PR! The actual fix looks fine to me; like I said, I have some suggestions for the accompanying test case.

All applied, thanks. I was not familiar with the test API, I just looked at some random test file and build from that one.

fulax:
> Would anybody know whether we can get rid of the actual version number?

I'm also not familiar with MacOS build environments, but the actual config seems to rely on pkg-config, so I guess `pkg-config --modversion libpng16` could be use to get that version number.

cmb69:
> I was not familiar with the test API, I just looked at some random test file and build from that one.

Yeah, we should have some documentation about the test suite. Until then, the best available info is in [gdtest.h](https://github.com/libgd/libgd/blob/master/tests/gdtest/gdtest.h).

> I'm also not familiar with MacOS build environments, but the actual config seems to rely on pkg-config, so I guess `pkg-config --modversion libpng16` could be use to get that version number.

Ah, that seems to be a good idea.

Anyhow, you also need to add the new test case to the respective [CMakeLists.txt](https://github.com/libgd/libgd/blob/master/tests/gdimagesetinterpolationmethod/CMakeLists.txt); currently, the test is apparently not run on CI.

cmb69 (approved):
Thank you! This looks good to me.

@vapier, what is our current merge strategy – target GD-2.3 or master?

fulax:
Hello,

Any chance we can go on with this issue? I could rewrite the patch or change the target branch if needed.

Let me know,
Thanks.

vapier:
is the issue fixed on master already ?  if not, we always merge to master first, and then cherry pick back to older branches if we want to fix it there.

cmb69:
Thanks @vapier (wasn't sure anymore).

@fulax, I've changed the branch, but that shows unrelated commits, so please rebase the PR onto current "master", force-push, and change the branch in the GH UI.

cmb69:
The test failures are unrelated; I'm going to merge. Thank you for the PR!

vapier:
at this point, we usually just cherry pick the fix back ourselves, although we don't say no to PRs in general.  which i've done now.

PR #851, opened 2022-10-30
Use libwebp pc module instead of libs

dg0yt:
WebP now provides a pkg-config module.


vapier:
what version of libwebp though? we've been supporting older versions that distros are shipping.

dg0yt:
> what version of libwebp though? we've been supporting older versions that distros are shipping.

It works with 1.2.4 in vcpkg. IIRC webp in Ubuntu 18.04 is too old. webp added pkg-config support 4 years ago with 1.0.1:
https://github.com/webmproject/libwebp/commit/eee00b662781abe7d45610cc5847438224db8c6b

If you want to retain compatibility with old versions of webp, a `pkg_check_modules` might be added to `FindWEBP.cmake`. (And vcpkg uses pkg-config even with MSVC.)

vapier:
our Autotools build already checks pkg-config and supports webp 0.2.0+.  so cmake should be comparable.

dg0yt:
> IIRC webp in Ubuntu 18.04 is too old.

I was wrong, even Ubunto 18.04 has it, as well as older Debian packages (0.6.1).

vapier:
i'll note that Ubuntu 16 & 14 LTS are still supported & actively used

the autotools generated .pc DTRT when it comes to finding libs via pkg-config or direct probing, and then generating the corresponding gd pc file

the current cmake/modules/FindWEBP.cmake only tests direct linkage ... it needs to check via PkgConfig first, and if that works, then we can add the result to PKG_REQUIRES_PRIVATES.

dg0yt:
> i'll note that Ubuntu 16 & 14 LTS are still supported & actively used

Is this still the lower bound?

vapier:
https://ubuntu.com/about/release-cycle

we can drop them both now.  16 goes fully EOL in a few months.

vapier:
many webp changes have landed since.  commit ce5aadf1990997ff8d318cfed3a2f9b2b729427a covered your CMakeLists.txt change, and i've cherry-picked your src/CMakeLists.txt as commit 49b47b0c6e09c22a9936a2b9fac061822813c8c5 now.

dg0yt:
Thank you!

PR #856, opened 2022-12-17
Fix #855: crash in gdImageClone after running gdImageFilledPolygon

### LINKED ISSUE #855: gdImageClone crashes when cloning an instance that filled polygon
**Describe the bug**
gdImageClone crashes when cloning after gdImageFilledPolygon.

**To Reproduce**
An example of the source is presented below.
It occurs with both gdImageCreate and gdImageCreateTrueColor.

```
#include "gd.h"
int main() {
	gdImagePtr im, clone;
	gdPoint pts[] = { {5, 5}, {50, 8}, {80, 4}, {85, 50}, {10, 60} };

	im = gdImageCreateTrueColor(100, 100);
	gdImageFilledPolygon(im, pts, 5, gdTrueColor(255, 0, 0)); /* it occurs only if there is this call */
	clone = gdImageClone(im); /* crashes on this line. */

	gdImageDestroy(clone);
	gdImageDestroy(im);

	return 0;
}
```
Possible cause:
gdImageFilledPolygon allocates an internal buffer polyInts, gdImageClone tries to clone it, but since it doesn't allocate it, it accesses a null pointer.

part of gdImageClone in gd.c :
```
	...
	if (src->polyAllocated > 0) {
		dst->polyAllocated = src->polyAllocated;
		for (i = 0; i < src->polyAllocated; i++) {
			dst->polyInts[i] = src->polyInts[i]; /* crashes due to dst->polyInts is NULL. */
		}
	}

	return dst;
}
```

**Expected behavior**
Add processing to allocate dst->polyInt if src->polyAllocated exists.

correction example of gdImageClone in gd.c :
```
	...
	if (src->polyAllocated > 0) {
		dst->polyInts = (int *)gdMalloc(sizeof(int) * src->polyAllocated); /* add this. */
		dst->polyAllocated = src->polyAllocated;
		for (i = 0; i < src->polyAllocated; i++) {
			dst->polyInts[i] = src->polyInts[i];
		}
	}
	...
```

**Actual results**
The current version crashes the process.
After fixing it, it finishes normally without crashing.

 - OS: Windows 10
 - Version: 2.3.3 and master
 - All libraries used by GD and their respective version : not relevant here

**Additional context**
Although the above correction example allocates polyInt, there might be an idea to omit the copy operation of polyInts.

Best Regards,

mocchi-2022:
I made the PR for [#855](https://github.com/libgd/libgd/issues/855#issue-1501289624).  The overflow check is also included. 
Similar to the ```style``` allocation fix(https://github.com/libgd/libgd/commit/a93eac0e843148dc2d631c3ba80af17e9c8c860f), return ```dst``` instead of ```NULL``` even if the overflow check fails.

### DIFF
--- a/src/gd.c
+++ b/src/gd.c
@@ -2940,7 +2940,7 @@ BGD_DECLARE(gdImagePtr) gdImageClone (gdImagePtr src) {
 	}
 
 	if (src->polyAllocated > 0 && overflow2(sizeof(int), src->polyAllocated) == 0) {
-		dst->polyInts = (int *) gdMalloc (sizeof (int) * src->polyAllocated);
+		dst->polyInts = gdMalloc (sizeof (int) * src->polyAllocated);
 		dst->polyAllocated = src->polyAllocated;
 		for (i = 0; i < src->polyAllocated; i++) {
 			dst->polyInts[i] = src->polyInts[i];
--- a/src/gd.c
+++ b/src/gd.c
@@ -2939,7 +2939,8 @@ BGD_DECLARE(gdImagePtr) gdImageClone (gdImagePtr src) {
 		dst->tileColorMap[i] = src->tileColorMap[i];
 	}
 
-	if (src->polyAllocated > 0) {
+	if (src->polyAllocated > 0 && overflow2(sizeof(int), src->polyAllocated) == 0) {
+		dst->polyInts = (int *) gdMalloc (sizeof (int) * src->polyAllocated);
 		dst->polyAllocated = src->polyAllocated;
 		for (i = 0; i < src->polyAllocated; i++) {
 			dst->polyInts[i] = src->polyInts[i];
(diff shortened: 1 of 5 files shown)


src/gd.c
```diff
@@ -2939,7 +2939,8 @@ BGD_DECLARE(gdImagePtr) gdImageClone (gdImagePtr src) {
 		dst->tileColorMap[i] = src->tileColorMap[i];
 	}
 
-	if (src->polyAllocated > 0) {
+	if (src->polyAllocated > 0 && overflow2(sizeof(int), src->polyAllocated) == 0) {
+		dst->polyInts = (int *) gdMalloc (sizeof (int) * src->polyAllocated);
```
vapier, line 2943: the `(int *)` cast isn't necessary.  i know we use it when setting up polyInts, but we should remove those.  not that you have to do that cleanup here.

vapier (changes requested):
one minor nit, but otherwise looks good, thanks

src/gd.c
mocchi-2022: Thank you for confirming. As you pointed out, this cast isn't necessary, so I delete it.

PR #862, opened 2023-02-18
Fix php/php-src#10614: Consistently set saveAlphaFlag

nielsdos:
The non-square degree rotation cases, and the 0 degree rotation case all set saveAlphaFlag to 1. But the other special cases for 90, 180, or 270 degrees don't. This results in losing the alpha channel, and in particular if an image is black on a transparent background, in a full-black image. Fix it by setting the saveAlphaFlag.

Credits to KapitanOczywisty for finding the root cause


nielsdos:
CI broke because of "configure: error: heif requested but not found", but that seems unrelated to this PR.

vapier:
fixes should be sent to master, not random branches.  we can cherry pick back to branches on request, but they merge to master first.

nielsdos:
My apologies, rebased on master now. Thanks.

nielsdos:
Hey @vapier , it's been some time. Could this get a review please? Some tests were failing (unfortunately the logs are gone by now), but IIRC they seemed like flaky tests or tests that didn't pass on current master as well.

nielsdos:
Ping again

PR #871, opened 2023-12-01
Update CI for macOS

dg0yt:
Disable continue-on-error on macOS CI:
Continue-on-error breaks propagating the "failed" state from the job to the macOS workflow.


vapier:
thx bud

PR #873, opened 2023-12-01
Fix "Run tests" step for windows mingw CI

dg0yt:
~Use script syntax for `command.com` instead of `sh`.~
Fix construction of `TMP` path.

### DIFF
--- a/.github/workflows/ci_windows_mingw.yml
+++ b/.github/workflows/ci_windows_mingw.yml
@@ -74,6 +74,6 @@ jobs:
       working-directory: ${{github.workspace}}
       run: |
         mkdir tmp
-        export TMP=${{github.workspace}}\\tmp
+        export TMP="$PWD/tmp"
         export CTEST_OUTPUT_ON_FAILURE=1
         ctest


vapier:
> Given that this CI configuration is using msys2 running on windows, and that cmake doesn't need bash, one might think that this CI configuration is meant to fill the gap left by maintainers or contributors rarely running windows...

that isn't a gap. we don't care how people invoke e.g. cmake. we assume those tools work.

> AFAIU all (setup-msys2) run invocations in this file use cmd syntax implicitly.
I don't know if there is a simple switch to use bash instead.

I'm referring to this:
https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#jobsjob_idstepsshell

dg0yt:
This PR fixes the step which fails. Maybe it is okay to accept this fix until someone is willing to spend the time on changing the default change for all steps in this workflow?

dg0yt:
Back to `bash` syntax: This is what the msys2 shell runs, [via a wrapper](https://github.com/msys2/setup-msys2/blob/5712755334301153e225db75ded6513e237802b1/main.js#L293).
Now using `$PWD` to construct the absolute path of the temporary directory.

dg0yt:
Ping @vapier.

dg0yt:
Ping @vapier.

vapier:
sorry for the delay bud.  this does seem like a much simpler approach, thx.

is the commit message still accurate ?  the script syntax used here isn't command.com iiuc.

dg0yt:
> is the commit message still accurate ? the script syntax used here isn't command.com iiuc.

You are right. I can remove the second line.

vapier:
thx bud for chasing it down

PR #877, opened 2024-02-21
Skip failure tests on riscv64.

JasenChao:
These two tests don't fit riscv64 and therefore fail at build time, hopefully skipped.
[This](https://openkoji.iscas.ac.cn/koji/buildinfo?buildID=131383) can prove that the modification is valid.

### DIFF
--- a/tests/gdimagecopyresampled/bug00201.c
+++ b/tests/gdimagecopyresampled/bug00201.c
@@ -45,7 +45,7 @@ int main()
 
 // Rounding issue, won't fix as it only happens on mingw 32bit.
 // __aarch64__/graviton. It fails within the CI while outside is 100% success over 100s builds&runs
-#if defined(__MINGW32__) || defined(__aarch64__) || defined(_M_ARM64)
+#if defined(__MINGW32__) || defined(__aarch64__) || defined(_M_ARM64) || defined(__riscv)
     return 77;
 #endif 
     background = blank_image(DEST_WIDTH,DEST_HEIGHT);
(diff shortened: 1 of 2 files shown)


tests/gdimagecopyresampled/bug00201.c
```diff
@@ -45,7 +45,7 @@ int main()
 
 // Rounding issue, won't fix as it only happens on mingw 32bit.
 // __aarch64__/graviton. It fails within the CI while outside is 100% success over 100s builds&runs
-#if defined(__MINGW32__) || defined(__aarch64__) || defined(_M_ARM64)
+#if defined(__MINGW32__) || defined(__aarch64__) || defined(_M_ARM64) || defined(__riscv)
```
vapier, line 48: this covers riscv32 & riscv64, but the commit message only says riscv64

tests/gdimagegrayscale/basic.c
```diff
@@ -12,7 +12,7 @@ int main()
 	char *path;
 // Rounding issue, won't fix as it only happens on mingw 32bit.
 // __aarch64__/graviton. It fails within the CI while outside is 100% success over 100s builds&runs
-#if defined(__MINGW32__) || defined(__aarch64__) || defined(_M_ARM64)
+#if defined(__MINGW32__) || defined(__aarch64__) || defined(_M_ARM64) || defined(__riscv)
```
vapier, line 15: same issue here -- needs to update comment

vapier (changes requested):

pierrejoye:
Could you try again using master

I migrate those tests a new perceptual diff, instead of exact match or similar, which is fragile and not very resilient to fp rounding diff.

vapier:
likely no longer relevant after commit f010284fb75defb4e247d404bf488a20542cc407

PR #880, opened 2024-05-10
cmake: add missing <stdlib.h> for exit to TestForHighBitCharacters

thesamesam:
The test will wrongly fail otherwise with newer stricter compilers that default to rejecting implicit function declarations.


vapier:
thx bud

PR #881, opened 2024-06-06
gdImageClone: fix typo during members copy.

devnexen:


vapier:
thx bud for the catch

PR #884, opened 2024-06-25
imageclone fix leaks when tile and/or brush are set.

devnexen:
those are cloned themselves which make things complicated to track to release manually thus taking care of cloned parts internally.

### DIFF
--- a/src/gd.c
+++ b/src/gd.c
@@ -244,6 +244,7 @@ BGD_DECLARE(gdImagePtr) gdImageCreate (int sx, int sy)
 	im->res_y = GD_RESOLUTION;
 	im->interpolation = NULL;
 	im->interpolation_id = GD_BILINEAR_FIXED;
+	im->cloned = 0;
 	return im;
 }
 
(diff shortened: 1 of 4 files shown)


thg2k:
I believe this might cause some troubles if the cloned image brush/tile are later set to a new image.

The only really clean solution I can think of is to clone the tiles and brushes (non recursively) on set and always free them on image destroy.

cmb69:
> I believe this might cause some troubles if the cloned image brush/tile are later set to a new image.

Right. And what happens if a tile/brush is set on an image, but the tile/brush is destroyed? Do we consider this a programmer error? And what happens if the same image is set as tile *and* brush?

> The only really clean solution I can think of is to clone the tiles and brushes (non recursively) on set and always free them on image destroy.

Alternatively it should be possible to implement ref-counting for images. That might break some client's expectations, though (e.g. `gdImageDestroy()` might not actually destroy the image). However, cloning the tiles/brushes when cloning an image might also break some expectations (e.g. clone image → change tile/brush → draw with tile/brush).

Given that libgd is relatively low-level, it *might* be better to not change its behavior, but to adapt consumers (e.g. PHP).

PR #893, opened 2024-10-14
Check libavif API return values, if available

cmb69:
Prior to libavif 1.1.0, `avifAlloc()` was infallible (it called `abort()` on OOM conditions); thus, several API functions which used `avifAlloc()` did not report failure.  That changed as of libavif 1.0.0[1], so checking and handling failure conditions can now be done. However, due to `avifAlloc()` being fallible as of libavif 1.1.0, this error checking and handling is mandatory to avoid more serious issues.

[1] <https://github.com/AOMediaCodec/libavif/blob/eb02b2ec52df5c0f50b71fbc51321c5ce435aaca/CHANGELOG.md?plain=1#L273-L281>

### DIFF
--- a/src/gd_avif.c
+++ b/src/gd_avif.c
@@ -546,12 +546,11 @@ static avifBool _gdImageAvifCtx(gdImagePtr im, gdIOCtx *outfile, int quality, in
 	// Note that MATRIX_COEFFICIENTS_IDENTITY enables lossless conversion from RGB to YUV.
 
 	avifImage *avifIm = avifImageCreate(gdImageSX(im), gdImageSY(im), 8, subsampling);
-#if AVIF_VERSION >= 1000000
 	if (avifIm == NULL) {
 		gd_error("avif error - Creating image failed\n");
 		goto cleanup;
 	}
-#endif
+
 	avifIm->colorPrimaries = AVIF_COLOR_PRIMARIES_BT709;
 	avifIm->transferCharacteristics = AVIF_TRANSFER_CHARACTERISTICS_SRGB;
 	avifIm->matrixCoefficients = lossless ? AVIF_MATRIX_COEFFICIENTS_IDENTITY : AVIF_MATRIX_COEFFICIENTS_BT709;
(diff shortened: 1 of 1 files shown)


src/gd_avif.c
```diff
@@ -402,7 +402,13 @@ BGD_DECLARE(gdImagePtr) gdImageCreateFromAvifCtx (gdIOCtx *ctx)
 	// (While AVIF image pixel depth can be 8, 10, or 12 bits, GD truecolor images are 8-bit.)
 	avifRGBImageSetDefaults(&rgb, decoder->image);
 	rgb.depth = 8;
+#if AVIF_VERSION >= 1000000
```
vapier, line 405: assuming the APIs used exist in older versions, better to write this in C.  it provides better compile-time coverage of code (like typos).

```
if (AVIF_VERSION >= 1000000) {
  ...
} else {
  ...
}
```

vapier (changes requested):

src/gd_avif.c
cmb69: I don't think this can work, because then the compiler would see an assignment of a return value for some `void` functions.

src/gd_avif.c
```diff
@@ -540,14 +546,25 @@ static avifBool _gdImageAvifCtx(gdImagePtr im, gdIOCtx *outfile, int quality, in
 	// Note that MATRIX_COEFFICIENTS_IDENTITY enables lossless conversion from RGB to YUV.
 
 	avifImage *avifIm = avifImageCreate(gdImageSX(im), gdImageSY(im), 8, subsampling);
-
+#if AVIF_VERSION >= 1000000
```
cmb69, line 549: To clarify my comment above: we could drop this version check, since `avifImageCreate()` always returned a pointer. Same for `avifEncoderCreate()` below.

vapier: anywhere we can drop CPP `#if` logic is an improvement

cmb69: Done.

src/gd_avif.c
vapier: makes sense since `avifRGBImageAllocatePixels` was void before 1.0.0

vapier:
i'm guessing the avif test failures in the CI are not new with this change

cmb69:
No, Ubuntu fails for a while, see e.g. https://github.com/libgd/libgd/actions/runs/11061813974/job/30735128327. Possibly because an [old libavif is used](https://github.com/libgd/libgd/blob/6e42ee80eb3bcf840c9fc03c6adc0550c1e02c63/.github/workflows/ci_ubuntu.yml#L113), but maybe because some incompatibilities with AOM and/or dav1d (not sure why both are installed for CI; dav1d should not be required). The test_avif_avif_im2im failure is strange; apparently not even a simple image can be saved as AVIF.

vapier:
thx bud

PR #895, opened 2024-10-26
Fix HEIF image handling with uninitialized library

### LINKED ISSUE #889: Test heif/heif_read fails with libheif 1.18.0+
**Describe the bug**
The test `heif/heif_read` fails with libheif 1.18.0 or newer which contain https://github.com/strukturag/libheif/commit/0b236060d0585d744e26e2f745cb18419a38ccb8.  The failing test in gd appears to be triggerd by a new check in libheif that returns an error `Image has no 'ispe' property` from `heif_context_read_from_memory_without_copy` for `tests/heif/label.heic`.

**To Reproduce**
```
 ./bootstrap.sh
 ./configure \
    --prefix=/usr \
    --disable-rpath
make
make check
```

**Expected behavior**
All checks should pass.

**Actual results**
The test `heif/heif_read` fails with:
```
FAIL: heif/heif_read
====================

dlopen: libavcodec.so.61: cannot open shared object file: No such file or directory
GD Warning: gd-heif context creation failed
heif/heif_read.c:23: Assert failed in <heif/heif_read.c:23>
FAIL heif/heif_read (exit status: 2)
```

 - OS: Arch Linux
 - Version rolling
- fontconfig 2.15.0
- libxpm 3.5.17
- libwebp 1.4.0
- libavif 1.1.1
- libheif 1.18.2

**Additional context**
[test-suite.log](https://github.com/user-attachments/files/17007528/test-suite.log)
https://gitlab.archlinux.org/archlinux/packaging/packages/gd/-/issues/1

YakoYakoYokuYoku:
From version `1.13.0` of libheif the library needs to be initialized to load its plugins and it has to be done by any consumer of the library, including libgd, I've forgot to add such thing. This PR adds two function for it, `gdHeifInit` and `gdHeifDeinit`, the former gets called by any method that handles HEIF images and the latter is used for cleanup. Fixes #889.

### DIFF
--- a/src/gd_heif.c
+++ b/src/gd_heif.c
@@ -23,12 +23,51 @@
 #define GD_HEIF_ALLOC_STEP (4*1024)
 #define GD_HEIF_HEADER 12
 
+static BGD_THREAD_LOCAL int heifInit = 0;
+
 typedef enum gd_heif_brand {
 	GD_HEIF_BRAND_AVIF = 1,
 	GD_HEIF_BRAND_MIF1 = 2,
 	GD_HEIF_BRAND_HEIC = 4,
 	GD_HEIF_BRAND_HEIX = 8,
 } gd_heif_brand;
+
+/*
+  Function: gdHeifInit
+
+    Initialize libheif, no need to call manually because other HEIF
+    functions do so.
+
+  Returns:
+
+    On success 0 is returned, otherwise 1 on error.
+*/
+BGD_DECLARE(int) gdHeifInit()
+{
+	if (heifInit)
+		return 0;
+	struct heif_error err = heif_init(NULL);
+	if (err.code != heif_error_Ok)
+		return 1;
+	heifInit = 1;
+	return 0;
+}
+
+/*
+  Function: gdHeifInit
+
+    Deinitialize libheif, call this whenever is not longer needed to
+    handle HEIF images.
+*/
+BGD_DECLARE(void) gdHeifDeinit()
+{
+	if (!heifInit) {
+		gd_error("gd-heif invalid library deinitialization");
+		return;
+	}
+	heif_deinit();
+	heifInit = 0;
+}
 
 /*
   Function: gdImageCreateFromHeif
@@ -134,6 +173,11 @@ static gdImagePtr _gdImageCreateFromHeifCtx(gdIOCtx *infile, gd_heif_brand expec
 		return NULL;
 	}
 	gdSeek(infile, 0);
+
+	if (gdHeifInit()) {
+		gd_error("gd-heif failure to initialize library");
+		return GD_FALSE;
+	}
 
 	while (n == GD_HEIF_ALLOC_STEP) {
 		temp = gdRealloc(filedata, size + GD_HEIF_ALLOC_STEP);
@@ -298,6 +342,11 @@ static int _gdImageHeifCtx(gdImagePtr im, gdIOCtx *outfile, int quality, gdHeifC
 	}
 
 	if (overflow2(gdImageSX(im) * 4, gdImageSY(im))) {
+		return GD_FALSE;
+	}
+
+	if (gdHeifInit()) {
+		gd_error("Failure to initialize heif library");
 		return GD_FALSE;
 	}
 
--- a/src/gd.h
+++ b/src/gd.h
@@ -382,6 +382,10 @@ typedef const char *gdHeifChroma;
 #define GD_HEIF_CHROMA_422 "422"
 #define GD_HEIF_CHROMA_444 "444"
 
+BGD_DECLARE(int) gdHeifInit();
+
+BGD_DECLARE(void) gdHeifDeinit();
+
 /* define struct with name and func ptr and add it to gdImageStruct gdInterpolationMethod interpolation; */
 
 /* Interpolation function ptr */
--- a/src/gdhelpers.h
+++ b/src/gdhelpers.h
@@ -12,6 +12,24 @@ extern "C" {
 #include <stdlib.h>
 #endif /* _WIN32_WCE */
 
+#if defined(_WIN32) || defined(CYGWIN) || defined(_WIN32_WCE)
+# ifdef __GNUC__
+#  if (__STDC_VERSION__ >= 201112L)
+#   define BGD_THREAD_LOCAL _Thread_local
+#  else
+#   define BGD_THREAD_LOCAL __thread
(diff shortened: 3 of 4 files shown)


src/gd_heif.c
```diff
@@ -23,13 +23,52 @@
 #define GD_HEIF_ALLOC_STEP (4*1024)
 #define GD_HEIF_HEADER 12
 
+static int heifInit = 0;
```
vapier, line 26: bools should be bool, not int

vapier (changes requested):
we can't expose format-specific init functions like this. no one will call them, and we don't want multiple inits. I'm not sure there's no point in calling deinit at all? in which case, can't we have the heif backend call the init function itself on demand and that's it?

cmb69:
First, thank you for the PR @YakoYakoYokuYoku!

> I'm not sure there's no point in calling deinit at all?

Well, they may release resources there, so it would be *nice* to call the function.

src/gd_heif.c
cmb69: Maybe more importantly, that won't work for multi-threading, I think.

src/gd_heif.c
```diff
@@ -135,6 +174,11 @@ static gdImagePtr _gdImageCreateFromHeifCtx(gdIOCtx *infile, gd_heif_brand expec
 	}
 	gdSeek(infile, 0);
 
+  if (gdHeifInit()) {
+    gd_error("gd-heif failure to initialize library");
+    return GD_FALSE;
+  }
```
cmb69, line 180: Indentation seems to be off here (and below).

src/gd_heif.c
vapier: neither type works for multithreading.  they'd need to be atomic and/or grab pthread locks.

that said, I don't think we've ever thought about MT, so I wouldn't be surprised if there were already issues in the codebase.

vapier:
> Well, they may release resources there, so it would be _nice_ to call the function.

i grok the hypothetical, but i question the realistic/practical aspects.  i still don't think people will (or reasonably can) call format-specific init/deinit functions.  the user of gd (e.g. php code) can be disconnected from the builder of gd (e.g. the server/distro), and the code can be written in a way that is format agnostic so users could upload arbitrary content.

calling format-specific init automatically is fairly easy for us in gd since it'd be a one-time event, but calling deinit & re-init could easily be difficult to get right.  if it's a loop processing multiple images, do we deinit when the image is written out or destroyed ?  do we do some refcounting in case multiple images are loaded at once ?  do we do some basic GC to delay ?

unless the deinit releases a significant amount of resources, I argue that we should never do it, nor expose an API for it.

src/gd_heif.c
cmb69: I'm not aware of any multi-threading issues in libgd when used with PHP. :)

vapier: gdkanji.c is not MT safe.  none of the error handler APIs are MT safe (which is to say, the handlers are per-process, not per-thread).  I guess most MT issues after that would be library-specific and not really on gd's plate.

maybe along those lines, should we punt these problems completely onto the user ?  if the only thing `gdHeifInit` does is call `heif_init`, and the only reason people would call `gdHeifInit` is because they want to explicitly use the HEIF format, then why not force them to call `heif_init` themselves ?  that would also let them call `heif_deinit` if they really wanted to.

if we can't handle it automatically in a format/library agnostic way, i don't think we're bringing any value-add here.

cmb69:
> unless the deinit releases a significant amount of resources, I argue that we should never do it, nor expose an API for it.

From https://github.com/strukturag/libheif/blob/33ec3d63470944906dbccffa085970083969de3a/libheif/api/libheif/heif.h#L544-L545:

> For backwards compatibility, it is not really necessary to call heif_init(), but some library memory objects will never be freed if you do not call heif_init()/heif_deinit().

I haven't checked the amount of resources which would not be freed, but given that `heif_init()` and `heif_deinit()` are only available as of libheif 1.13.0, we cannot call the functions unconditionally anyway, unless we raise the libheif requirements (so far we require libheif >= 1.7.0).

 Given that there are more urgent issues regarding libgd than some memory leak at the end of the process (well, usually only there), I think we should *postpone* calling these functions. And the cleaner solution would likely be to introduce some `gdInit()`/`gdFinalize()` first, but these may have their own issues (and wouldn't be available with older libgd).

vapier:
> I haven't checked the amount of resources which would not be freed

Looking at the [implementation](https://github.com/strukturag/libheif/blob/33ec3d63470944906dbccffa085970083969de3a/libheif/init.cc#L142), it unloads any loaded external plugins, and clears a colorspace LUT.  It doesn't seem like those alone are a big enough deal for us to care about.

Really the only concerning thing is what @YakoYakoYokuYoku noted at the top: if you don't call init, then no external plugins will be loaded.

> but given that heif_init() and heif_deinit() are only available as of libheif 1.13.0, we cannot call the functions unconditionally anyway, unless we raise the libheif requirements (so far we require libheif >= 1.7.0).

If the libheif version is available at compile time, we can check that.  Or we can add configure-time checks for the functions and use them if available.  I won't seriously suggest probing via `dlopen` and related APIs.

> And the cleaner solution would likely be to introduce some `gdInit()`/`gdFinalize()` first.

I'm still on the side of doing nothing here.  This only affects people using libheif with external plugins, and those people are free to call init themselves.  Otherwise it's extra boilerplate that is a nop, and decades of existing GD code prob won't update to use it.

The libheif API says its init/deinit calls are refcounted, so we could call init safely everytime HEIF APIs are used, but we wouldn't call deinit, so we'd be leaking that logic.  Which, practically speaking, is what we already do with older versions, so arguably that's maintaining status quo.

pierrejoye:
Should we not rather add a gd main init, and add whatever is needed in there?

InitOnceExecuteOnce on windows
pthread_once on posix 
not sure about other.

For shared uses, document it, explain if one uses dl/LoadLibrary, unload it and load it again, it is required to call that function again "manually" as the underlying kernel won't do it.

for static link usage, caller must call that function. Nothing we can do there (like in php f.e.) but that's easy as php has a per process/thread init where needed, easy to add in the ext/gd

Alternatively, it could be done on demand, add a check if it has been initialized in _gdImageHeifCtx, if not init, return on error, continue if it was successful? same code as now but no public function exposed yet, as it is, if I remember correctly, the only requiring an init, maybe Freetype too, I have to check.

PR #905, opened 2024-12-21
Improve build time of Windows CI jobs

cmb69:
These take particularly long (about 90 to 120 seconds), presumably because NMake builds can't use parallel processing.  Thus we generate Visual Studio solutions which can be build in parallel.

### DIFF
--- a/.github/workflows/ci_windows.yml
+++ b/.github/workflows/ci_windows.yml
@@ -68,4 +68,4 @@ jobs:
           echo %TEMP%
           set PATH=${{github.workspace}}\vcpkgusr\${{ matrix.arch }}-windows\bin;%PATH%
           set CTEST_OUTPUT_ON_FAILURE=1
-          nmake test
+          ctest -C ${{env.BUILD_TYPE}}
--- a/.github/workflows/ci_windows.yml
+++ b/.github/workflows/ci_windows.yml
@@ -39,7 +39,7 @@ jobs:
       - name: Configure CMake
         # vcpkg libde265 fails on arm64 and match prefix
         if: matrix.arch == 'amd64_arm64'
-        run:  cmake -D CMAKE_GENERATOR="Visual Studio 17 2022" -DENABLE_PNG=1 -DENABLE_FREETYPE=1 -DENABLE_JPEG=1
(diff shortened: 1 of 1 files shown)


cmb69:
Hmm, build is faster, but configuration takes longer. Might not be worth to pursue this.

.github/workflows/ci_windows.yml
```diff
@@ -56,7 +56,7 @@ jobs:
       - name: Build
         run: |
          cd ${{github.workspace}}/build
-         nmake
+         cmake --build . --config ${{env.BUILD_TYPE}} --parallel 2
```
vapier, line 59: is there an `nproc`-like tool we can run so we don't have to hardcode `2` ?  or is there a cmake flag to use all available cores ?  or we just use equiv of make's `-j` option where we queue all the jobs and let the OS sort them out ?

vapier (commented):
I think there's value in reducing the number of tools we run across platforms, especially tools that only exist on a single platform (i.e. `nmake`).

`ninja` is pretty popular nowadays.  Can we generate ninja outputs and run that tool instead of nmake or cmake ?

cmb69:
> `ninja` is pretty popular nowadays. Can we generate ninja outputs and run that tool instead of nmake or cmake ?

I'm not a big fan of bringing in additional tooling, but Ninja is preinstalled on GH runners, and works very well.

.github/workflows/ci_windows.yml
cmb69: When using Ninja, this is automatic. From https://ninja-build.org/manual.html:

> Builds are always run in parallel, based by default on the number of CPUs your system has.

vapier:
> > `ninja` is pretty popular nowadays. Can we generate ninja outputs and run that tool instead of nmake or cmake ?
> 
> I'm not a big fan of bringing in additional tooling, but Ninja is preinstalled on GH runners, and works very well.

In general I agree, but I think ninja has been widely adopted in the OSS ecosystem now, especially with it being integrated into cmake, and it being strongly supported on "all" platforms (i.e. Linux, macOS, Windows).  And it seems like a good fit for what you're trying to do in this PR :).

vapier (approved):
Anything else you want to do at this point ?  Otherwise I'd squash this PR into a single commit and merge it now.

cmb69:
Nope, fine for me as is. Please go ahead. :)

PR #909, opened 2024-12-22
Fix warning when writing TIFF

cmb69:
As of libtiff 4.4.0, the library warns that `COMPRESSION_DEFLATE` is a legacy codec identifier, and that `COMPRESSION_ADOBE_DEFLATE` is more widely supported.  Since there appears to be no further difference in libtiff, we switch to the latter for possibly better compatibility.


vapier:
when was the define first added ?  do we have a min ver requirement for it ?

cmb69:
> when was the define first added ? do we have a min ver requirement for it ?

See https://gitlab.com/libtiff/libtiff/-/commit/57e10aa76ea894a2796830b450fe1e8c2d1af8bc. That was 20 years ago; I don't think a min version requirement is relevant for *this*.

However, we do not have a min version requirement for libtiff; autotools prefer pkg-config which looks for libtiff-4, but falls back to checking for existence of `TIFFClientOpen()` which is available since last century. CMake just uses `find_package()`. I don't know what to do about this; I wouldn't even know which version we require (I only did builds against 4.6.0 and 4.7.0, both somewhat recent). Maybe leave as is until someone complains, and adjust then.

vapier:
Usually better to see what release contained the change as sometimes projects can integrate a change but not release it for a while (we've def done that).

vapier:
Thanks for double checking.  libtiff has an inconsistent release history.

PR #910, opened 2024-12-22
Join build and tests workflows

cmb69:
As is, we have four individual workflows running for each push or pull request; while their results are gathered in the overview, they still have individual run IDs, and so navigating the results can be tedious. Maybe even worse, if jobs fail in multiple of these workflows, there are multiple email notifications about that, what is annoying.

Thus, we join the workflows, but still retain the individual workflow files.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1,3 +1,7 @@
+# The main CI workflow which runs the individual workflows
+# and gathers their results in the workflow overview.
+# See <https://docs.github.com/en/actions/sharing-automations/reusing-workflows>
+
 name: CI build and test
 
 on:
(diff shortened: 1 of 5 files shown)


cmb69:
https://github.com/libgd/libgd/actions/runs/12457378954?pr=910 looks like a small improvement to me, and indeed there was only one email notification.

Should probably shorten the names for the overview above; and maybe integrate the Shellcheck CI as well.

vapier:
looks reasonable to me.  can you put a header in .github/workflows/ci.yml leaving some breadcrumbs for people reading the file ?

cmb69:
Is [this](https://github.com/libgd/libgd/pull/910/commits/5760bd1b52521faae3dda7ed9a4fa3cb8efb664c) okay?

vapier (approved):

PR #912, opened 2024-12-24
Fix BMP header size

cmb69:
The `bfSize` member of `BITMAPFILEHEADER` is supposed to store the size of the BMP[1], and as such needs to be equal to the file size (for BMPs stored as files).  Otherwise that may trip up readers; at least they may assume that the BMP is corrupted; e.g. ImageMagick warns about this.

While for compressed BMP images we actually calculate the size as we go, for uncompressed BMP images we calculate the size in advance, but so far did not account for padding of the rows.

We also fix the `biSizeImage` of the `BITMAPINFOHEADER`[2] accordingly.

[1] <https://learn.microsoft.com/en-us/windows/win32/api/wingdi/ns-wingdi-bitmapfileheader>
[2] <https://learn.microsoft.com/en-us/windows/win32/api/wingdi/ns-wingdi-bitmapinfoheader>

### DIFF
--- a/src/gd_bmp.c
+++ b/src/gd_bmp.c
@@ -199,11 +199,11 @@ static int _gdImageBmpCtx(gdImagePtr im, gdIOCtxPtr out, int compression)
 	}
 
 	/* bitmap header + info header + data */
-	total_size = 14 + info_size + bitmap_size + padding * im->sy;
+	total_size = 14 + info_size + bitmap_size;
 
 	/* write bmp header info */
 	gdPutBuf("BM", 2, out);
-	gdBMPPutInt(out, total_size);
+	gdBMPPutInt(out, total_size + padding * im->sy);
 	gdBMPPutWord(out, 0);
 	gdBMPPutWord(out, 0);
 	gdBMPPutInt(out, 14 + info_size);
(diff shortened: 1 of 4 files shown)


cmb69:
Ubuntu and macOS CI failures are unrelated.

src/gd_bmp.c
```diff
@@ -192,8 +192,14 @@ static int _gdImageBmpCtx(gdImagePtr im, gdIOCtxPtr out, int compression)
...
 
+	/* The line must be divisible by 4, else its padded with NULLs */
+	padding = ((int)(im->trueColor ? 3 : 1) * im->sx) % 4;
+	if (padding) {
+		padding = 4 - padding;
+	}
+
 	/* bitmap header + info header + data */
-	total_size = 14 + info_size + bitmap_size;
+	total_size = 14 + info_size + bitmap_size + padding * im->sy;
```
cmb69, line 202: That would actually break the size of compressed images. We must only add the padding when we write the `bfSize` field.

cmb69:
The CI failures are unrelated again (note that there are occassional segfaults on MinGW for different tests; that might justify some further investigation).

cmb69:
Rebased and force-pushed to get CI green (at least Ubuntu should be fixed in the meantime).

pierrejoye:
thanks, and much cleaner implementation too!

PR #913, opened 2024-12-25
Transparent support for writing 1bpp and 4bpp BMPs

### LINKED ISSUE #911: BMP output support for specific color depths
Forwarded from https://github.com/php/php-src/issues/17195.

@mmellon suggested to add an argument which specifies the color depth of the BMP image. I'm not sure whether that is the best idea (besides that would constitute an API break for us, unless we would introduce a new function). Anyhow, currently we do

https://github.com/libgd/libgd/blob/4150a3174615397c38fc55310a1af35a7861054b/src/gd_bmp.c#L210

For palette images, it seems to me we could basically use `ceil(log2(colorsTotal))` instead of the hard-coded `8`, and then write 1-,  4, or 8-bit images as needed (we already support these depths when reading). This should be backward compatible, and it would be up to clients to prepare the image so it's fits their purpose; that is calling `gdImageTrueColorToPalette()` if necessary.

cmb69:
As is, we write palette images as 8bpp BMP images.  Besides that this is wasteful, some readers may not even be able to read 8bpp BMP images. Thus, we write palette images with at most two colors as 1bpp BMPs, unless compression is requested, because BMPs do not support compressed encoding of 1bpp images; we write palette images with at most 16 colors as 4bpp BMPs, and apply `BMP_BI_RLE4` compression if requested.

We piggy-back on the existing implementation for `BMP_BI_RLE8`; that keeps it simple, and although it might not yield the best compression, it ought to be good enough.

References:
<https://en.wikipedia.org/wiki/BMP_file_format>
<https://learn.microsoft.com/en-us/windows/win32/gdi/bitmap-compression>

Closes #911.

---

While implementing this, I've noticed that the RLE compression is not quite as good as it could be. While there is room for improvement, it might get pretty tricky, and it may not be worth spending further time on that, given that RLE encoding only yields somewhat good compression for images with few details, anyway.

### DIFF
--- a/src/gd_bmp.c
+++ b/src/gd_bmp.c
@@ -91,7 +91,7 @@ BGD_DECLARE(void *) gdImageBmpPtr(gdImagePtr im, int *size, int compression)
 	gdIOCtx *out = gdNewDynamicCtx(2048, NULL);
 	if (out == NULL) return NULL;
 	if (!_gdImageBmpCtx(im, out, compression))
-	rv = gdDPExtractData(out, size);
+		rv = gdDPExtractData(out, size);
 	else
 		rv = NULL;
 	out->gd_free(out);
--- a/src/gd_bmp.c
+++ b/src/gd_bmp.c
@@ -33,8 +33,8 @@
 #include "gd_errors.h"
 #include "bmp.h"
 
-static int compress_row(unsigned char *uncompressed_row, int length);
-static int build_rle_packet(unsigned char *row, int packet_type, int length, unsigned char *data);
+static int compress_row(unsigned char *uncompressed_row, int length, int ppbyte, int nibble);
+static int build_rle_packet(unsigned char *row, int packet_type, int length, unsigned char *data, int ppbyte, int nibble);
 
 static int bmp_read_header(gdIOCtxPtr infile, bmp_hdr_t *hdr);
 static int bmp_read_info(gdIOCtxPtr infile, bmp_info_t *info);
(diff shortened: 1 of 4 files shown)


cmb69:
CI failures are unrelated.

src/gd_bmp.c
```diff
@@ -91,7 +91,7 @@ BGD_DECLARE(void *) gdImageBmpPtr(gdImagePtr im, int *size, int compression)
 	gdIOCtx *out = gdNewDynamicCtx(2048, NULL);
 	if (out == NULL) return NULL;
 	if (!_gdImageBmpCtx(im, out, compression))
-		rv = gdDPExtractData(out, size);
+	rv = gdDPExtractData(out, size);
```
vapier, line 94: whitespace change looks unrelated & incorrect ?

tests/bmp/bmp_low_depth.c
```diff
@@ -0,0 +1,67 @@
+#include "gd.h"
```
vapier, line 1: please add a file header explaining the test

tests/bmp/bmp_low_depth.c
```diff
@@ -0,0 +1,67 @@
...
+#include "gdtest.h"
+
+int get_bpp(unsigned char *data)
```
vapier, line 4: not a big deal, but please mark local funcs as static

vapier (changes requested):
minor nits

pierrejoye:
Full BMP support is implemented, but OS/2 bmp.

Closing this PR as it is now supported.

Thanks :)

PR #914, opened 2024-12-26
Fix argument order of `_aligned_malloc()`

cmb69:
<https://learn.microsoft.com/en-us/cpp/c-runtime-library/reference/aligned-malloc>


thg2k:
@cmb69 Hey, I'm not familiar with Windows but from the doc page you linked I understand that _aligned_malloc should be paired with _aligned_free, and from a quick look at the code this seems not to be the case. WDYT?

PR #915, opened 2024-12-27
Avoid undefined behavior

cmb69:
Left-shifting the (32bit) signed int 1 by 31 exceeds the supported range, and is undefined behavior according to the C standards.  Instead of casting to int, which is implicit anyway, we could cast to unsigned int instead (or use `1U`), but directly using `INT_MAX` is more self- explaining.

PS: cf. <https://github.com/php/php-src/pull/17270/commits/86987c8c71a51ebbc09d97366c1c006c812092e6>

### DIFF
--- a/src/gd_nnquant.c
+++ b/src/gd_nnquant.c
@@ -309,7 +309,7 @@ static int contest(nn_quant *nnq, int al, int b, int g, int r)
 	double bestd,bestbiasd;
 	register int *p,*f, *n;
 
-	bestd = ~(((int) 1)<<31);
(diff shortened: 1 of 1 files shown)


vapier:
code seems a bit janky in assigning an integer to a double, but eh, you're def not making the code worse

cmb69:
> code seems a bit janky in assigning an integer to a double

`bestd` was originally an `unsigned int`, but that was changed via 489e90f7aa7476a31a6f09045ea42be878bc73d1; possibly to avoid overflow conditions?

PR #916, opened 2024-12-27
Fix heap corruption on Windows

cmb69:
The recently merged #914 introduced a potential head corruption on Windows, since memory allocated with `_aligned_malloc()` needs to be freed with `_aligned_free()`.  Thus we introduce `free16()`.

---

Running the test suite against libgd built with libimagequant confirms the issue (three test cases fail with 0xc0000374).

Thanks to @thg2k for pointing out the problem!

### DIFF
--- a/src/gd_topal.c
+++ b/src/gd_topal.c
@@ -1534,6 +1534,14 @@ static void *malloc16(size_t size)
 	return _aligned_malloc(size, 16);
 #endif
 }
+static void free16(void *ptr)
+{
+#ifndef _WIN32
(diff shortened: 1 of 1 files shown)


cmb69:
Interestingly, these tests don't seem to fail on MinGW (https://github.com/libgd/libgd/actions/runs/12518941566).

vapier:
> Interestingly, these tests don't seem to fail on MinGW

Fresh processes will have clean heaps and get a lot of well aligned memory to start with.

PR #918, opened 2024-12-28
Drop useless remapping of palette in gdImageNeuQuant()

cmb69:
This code has been taken from pngnq.c[1] where it makes much sense, because the image is always output in PNG format.  For our case, it does not make sense, since the output format may be different, and we already have respective remapping code in gd_png.c.  Furthermore, this remapping may destroy the advantageous topological properties of the NeuQuant color quantization[2].

And actually, that code wouldn't work as it's supposed to, since the alpha channel cannot be 255 for libgd, and as such the remapping just inverses the order of the palette entries.

[1] <https://sourceforge.net/projects/pngnq/>
[2] <https://web.archive.org/web/20070416070905/http://members.ozemail.com.au/~dekker/NeuQuant.pdf>, section 6

### DIFF
--- a/src/gd_nnquant.c
+++ b/src/gd_nnquant.c
@@ -497,8 +497,6 @@ BGD_DECLARE(gdImagePtr) gdImageNeuQuant(gdImagePtr im, const int max_color, int 
 	const int newcolors = max_color;
 	const int verbose = 1;
 
-	int bot_idx, top_idx; /* for remapping of indices */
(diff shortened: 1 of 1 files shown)


pierrejoye:
Thanks!

Added via https://github.com/libgd/libgd/commit/03359522e0d65f09af549bfda3264f249964d57c and a test case :)

PR #920, opened 2025-01-01
Introduce gdLayerDiff() and gdEffectDiff

cmb69:
`gdLayerDiff()` calculates the difference of two given colors; the difference of each RGBA channel is assigned to the result.  The operation is commutative.

The difference of any color `c` and `0x00000000` is always `c` (no-op); the difference of `c` and `0x7fffffff` is the inverted color (negation). Neither of these are particularly relevant (note that the latter is similar to `gdImageNegate()` but also inverts the alpha channel). However, some interesting effects can be created by only setting some channels, and possibly not to the maximum values.

As usual, `gdEffectDiff` can be set with `gdImageAlphaBlending()` so it is applied for each `gdImageSetPixel()` call.  A particular interesting use case is to use `imagecopy()` to create a diff of two images.  Black pixels indicate no difference; the "brighter" the pixels, the larger the difference.  The result is similar to `gdTestImageDiff()` in our test suite (although less sophisticated).

---

Note that I first wanted to implement this as calculating the euclidean distance of the pixels and just setting the R, G and B channel to the clamped result (similar to to `magick -compose difference`). However, it seems to me that the effect is more versatile as it is implemented now.

Also note that the test case is basically a copy of gdeffectmultiply.c/gdeffectoverlay.c, and I consider neither of these test cases particularly valuable; besides that they are placed in the "wrong" directory, and that you get no useful feedback in case of failure (should use meaningful test assertions), they barely verify the proper working of the effects. However, improving these test cases is orthogonal to this PR.

Furthermore note that storing an integer value as `alphaBlendingFlag` and then dispatching to the desired function in `gdImageSetPixel()` is neither elegant nor robust; we should consider to store the function pointer directly, but obviously, this would be a BC break.

### DIFF
--- a/src/gd.c
+++ b/src/gd.c
@@ -4217,7 +4217,7 @@ BGD_DECLARE(int) gdLayerMultiply (int dst, int src)
 BGD_DECLARE(int) gdLayerDiff (int dst, int src)
 {
 	int r1,b1,g1,a1,r2,b2,g2,a2;
-	int diff_a,diff_r,diff_g,diff_b,diff;
+	int diff_a,diff_r,diff_g,diff_b;
 
 	a1 = gdTrueColorGetAlpha(dst);
 	a2 = gdTrueColorGetAlpha(src);
--- a/src/gd.c
+++ b/src/gd.c
@@ -1299,6 +1299,9 @@ BGD_DECLARE(void) gdImageSetPixel (gdImagePtr im, int x, int y, int color)
 					case gdEffectMultiply :
 						im->tpixels[y][x] = gdLayerMultiply(im->tpixels[y][x], color);
 						break;
+					case gdEffectDiff :
+						im->tpixels[y][x] = gdLayerDiff(im->tpixels[y][x], color);
+						break;
 				}
 			} else {
 				im->pixels[y][x] = color;
@@ -4070,6 +4073,7 @@ BGD_DECLARE(int) gdImageCompare (gdImagePtr im1, gdImagePtr im2)
  *   - <gdImageAlphaBlending>
  *   - <gdLayerOverlay>
  *   - <gdLayerMultiply>
+ *   - <gdLayerDiff>
  */
 BGD_DECLARE(int) gdAlphaBlend (int dst, int src)
 {
@@ -4131,6 +4135,7 @@ static int gdAlphaOverlayColor (int src, int dst, int max );
  *   - <gdImageAlphaBlending>
  *   - <gdAlphaBlend>
  *   - <gdLayerMultiply>
+ *   - <gdLayerDiff>
  */
 BGD_DECLARE(int) gdLayerOverlay (int dst, int src)
 {
@@ -4170,6 +4175,7 @@ static int gdAlphaOverlayColor (int src, int dst, int max )
  *   - <gdImageAlphaBlending>
  *   - <gdAlphaBlend>
  *   - <gdLayerOverlay>
+ *   - <gdLayerDiff>
  */
 BGD_DECLARE(int) gdLayerMultiply (int dst, int src)
 {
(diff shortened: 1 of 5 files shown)


src/gd.c
```diff
@@ -4193,6 +4199,45 @@ BGD_DECLARE(int) gdLayerMultiply (int dst, int src)
...
+ *
+ * See also:
+ *   - <gdImageAlphaBlending>
+ *   - <gdAlphaBlend>
+ *   - <gdLayerMultiply>
+ *   - <gdLayerOverlay>
+ */
+BGD_DECLARE(int) gdLayerDiff (int dst, int src)
+{
+	int r1,b1,g1,a1,r2,b2,g2,a2;
```
vapier, line 4219: spaces after commas please

tests/gdimagesetpixel/gdeffectdiff.c
```diff
@@ -0,0 +1,33 @@
+#include "gd.h"
```
vapier, line 1: please add a proper file header: license, copyright, and text explaining the test goals.

tests/gdimagesetpixel/gdeffectdiff.c
```diff
@@ -0,0 +1,33 @@
...
+#include "gdtest.h"
+
+int main()
+{
+	gdImagePtr im;
+	int x, y, c;
+	int r=0;
```
vapier, line 8: spaces around binary operators like `=`

tests/gdimagesetpixel/gdeffectdiff.c
```diff
@@ -0,0 +1,33 @@
...
+
+
```
vapier, line 10: only one blank line

tests/gdimagesetpixel/gdeffectdiff.c
```diff
@@ -0,0 +1,33 @@
...
+	im = gdImageCreateTrueColor(256, 256);
+	gdImageAlphaBlending( im, gdEffectReplace );
```
vapier, line 12: no space after `(` or before `)`

src/gd.c
```diff
@@ -4193,6 +4199,45 @@ BGD_DECLARE(int) gdLayerMultiply (int dst, int src)
...
+	int diff_a,diff_r,diff_g,diff_b;
+
+	a1 = gdTrueColorGetAlpha(dst);
```
vapier, line 4222: i'm on the fence whether all these variables are needed vs a single statement.

```
    return gdTrueColorAlpha(
        abs(gdTrueColorGetAlpha(dst) - gdTrueColorGetAlpha(src)),
        abs(gdTrueColorGetRed(dst) - gdTrueColorGetRed(src)),
        abs(gdTrueColorGetGreen(dst) - gdTrueColorGetGreen(src)),
        abs(gdTrueColorGetBlue(dst) - gdTrueColorGetBlue(src)));
```

src/gd.c
```diff
@@ -4193,6 +4199,45 @@ BGD_DECLARE(int) gdLayerMultiply (int dst, int src)
 		);
 }
 
+/**
+ * Function: gdLayerDiff
+ *
+ * Difference of two colors; commutative operation.
```
vapier, line 4205: should it note that it's absolute ?

vapier (changes requested):
at a high level, can you talk about your end goals here ?  imagemagick's compare tooling is already extremely powerful, and I have a hard time seeing us building out anything competitive.
https://usage.imagemagick.org/compare/

cmb69:
> at a high level, can you talk about your end goals here ?

Well, this was triggered by the discussion on https://github.com/php/php-src/pull/14877, where [`gdTestImageDiff()`](https://github.com/libgd/libgd/blob/77adfd27f776352e8095d98ef4902c28975d19af/tests/gdtest/gdtest.c#L423-L504) has been mentioned, including that it might be exposed as general function. As a POC I've wrapped its logic in `gdLayerDiff()`. But frankly, I never liked the diffs this function created (especially the upscaling of the color channel differences seems somewhat arbitrary and generally doubtful). So I've looked into what ImageMagick provides, and found https://usage.imagemagick.org/compare/#difference quite interesting so after some attempts I came up with the current solution, which is straight forward (no arbitrary scaling; no channel combination), and as such may serve as building block for more elaborate solutions.

I'll come up with a better test case and some examples soonish.

pierrejoye:
Still testing new gdTestImageDiff, a perceptual gdTestImagePerceptualDiff.

Within the tests, I did something like attached pics, to help debugging. We could also add the list of pixels coordinates, expected and actual values to help debugging.

For now I did an overlay with a 80%, quite common in other tests and comparison tools, to clearly see where are the differences. As some deltas may not be easily visible and drawing the actual values does not make sense per se (not visible by human, etc). Thoughts?

diff
testpdiff
testpdiff2

PR #925, opened 2025-01-02
Close #921: Deprecate "native" MSVC and MSYS build support

### LINKED ISSUE #921: Drop "native" MSVC and MSYS build support?
I'm referring to what is described in https://github.com/libgd/libgd/blob/master/windows/readme.md and https://github.com/libgd/libgd/blob/master/windows/msys/README.MSYS.md.

While I've used the former a couple of years ago (but switched to CMake a while ago), I don't think it makes sense to maintain that any longer. The Makefile is pretty outdated (most notably no support for libheif), and the test generation even more so (only fontconfig is optional; all other tests would be configured, even if the supporting libraries would not be available, causing lots of build failures).

From a quick look, the MSYS build has pretty much the same issues.

Given that nobody cared to keep that up to date for almost four years, it might be reasonable to completely drop that. If we want to be conservative, we should at least clearly document that these build chains are obsolete/unmaintained (or even deprecate them, with the plan to remove them in the future).

cmb69:
PS: I'm also fine with GH flavored markdown, i.e.
````md
> [!WARNING]
> … deprecated …
````

### DIFF
--- a/windows/msys/README.MSYS.md
+++ b/windows/msys/README.MSYS.md
@@ -1,5 +1,9 @@
 # MSYS Fallback Makefile
 
+⚠️
+Due to lack of maintenance, building libgd this way is deprecated.
+Use CMake or autotools instead.
+
 This is a simple, straightforward Makefile for building LibGD with
 MinGW on MSYS (or possibly Cygwin).  It is here for anyone who doesn't
 want to deal with autotools or CMake on Windows or who can't get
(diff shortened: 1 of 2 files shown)


vapier:
> PS: I'm also fine with GH flavored markdown, i.e.
> 
> ```
> > [!WARNING]
> > … deprecated …
> ```

a note block like that would prob be better as it'll stand out more when rendered.

cmb69:
> a note block like that would prob be better as it'll stand out more when rendered.

Done. :)

vapier:
thx as always bud

Issue #320, opened 2016-10-03
gdImageRotateInterpolated() converts the source image to truecolor


cmb69:
If a palette image is passed to gdImageRotateInterpolated() it is converted to truecolor. That is at the very least unexpected, because gdImageRotateInterpolated() is supposed to return a modified result, but not to change the source image.
### Test Program

```
#include <stdio.h>
#include <assert.h>
#include <gd.h>


int main()
{
    gdImagePtr src, dst;
    int black;

    src = gdImageCreate(10, 10);
    black = gdImageColorAllocate(src, 0, 0, 0);

    printf("src is truecolor: %d\n", gdImageTrueColor(src));

    dst = gdImageRotateInterpolated(src, 30.0, black);
    assert(dst != NULL);

    printf("src is truecolor: %d\n", gdImageTrueColor(src));
}
```
### Expected Output

```
src is truecolor: 0
src is truecolor: 0
```
### Actual Output

```
src is truecolor: 0
src is truecolor: 1
```

pierrejoye:
All interpolation converts to TC yes. It makes thing too complicated with palette while massively reducing the quality.

cmb69:
This is not about the output (or the internal calculation), but rather the fact that the *source* image is converted to true-color. This might have better been solved by only supporting true-color images in the first place (and failing for palette images), but we cannot change that anymore. However, we should document that this conversion is happening.

pierrejoye:
oh, the original image is converted? sorry I misread. That should not happen. I will look into it.

cmb69:
https://github.com/libgd/libgd/blob/9af6d1254f2be57bf5179f2b02a0e90d3806570d/src/gd_interpolation.c#L1882-L1887
Solution would be simple: just create a copy of the src image before converting to palette. I'm not sure whether it's worth it. Users can do that themselves if they need to retain the source image (often there is no need to).

pierrejoye:
It is needed to be converted anyway. 

The fix is:
gdImageClone ion line 1885
gdImagePaletteToTrueColor(clone)
back_src = src;
src = clone;

At this stage I would like to go with the "fix" rather than making it non obvious. If the src changes, the parameter should be ** at the very least (to make it clear and not const *).

cmb69:
That's fine for me as well. :)

pierrejoye:
PR done, please check :)

PR #929, opened 2025-01-04
Fix fix for #320

cmb69:
At least `gdImageRotateNearestNeighbour()` requires a truecolor image, so we should actually pass the truecolor clone.  To be safe regarding future changes, we also do this for `gdImageRotateGeneric()`.  And if rotation of zero degrees is requested, we return the already cloned image, instead of creating another clone.

We also extend the existing test case to cover all relevant code paths.

---

I've noticed a segfault while trying to port the [original fix](4a0d7d86f063b399a7549d20f24cc7f4ae059b0e) to PHP's bundled libgd.

### DIFF
--- a/src/gd_interpolation.c
+++ b/src/gd_interpolation.c
@@ -1840,15 +1840,8 @@ BGD_DECLARE(gdImagePtr) gdImageRotateInterpolated(const gdImagePtr src, const fl
 	/* 0 && 90 degrees multiple rotation, 0 rotation simply clones the return image and convert it
 	   to truecolor, as we must return truecolor image. */
 	switch (angle_rounded) {
-		case    0: {
-			gdImagePtr dst = gdImageClone(src);
-
-			if (dst == NULL) {
-				return NULL;
-			}
-			if (src_cloned) gdImageDestroy(src_tc);
-			return dst;
-		}
+		case    0:
+			return src_cloned ? src_tc : gdImageClone(src);
 
 		case -27000:
 		case   9000:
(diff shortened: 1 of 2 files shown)


cmb69:
Ugh, need to check what's going on with Ubuntu clang CI.

cmb69:
Rebased and force-pushed to get CI passing.

vapier (approved):
your reasoning sounds correct, but i'm not super familiar with the interpolation APIs

vapier:
coverity has flagged the unit test after this PR merged.  if you don't have access to https://scan.coverity.com/projects/libgd, create an acct and I can add you.

```
*** CID 529990:  Memory - illegal accesses  (USE_AFTER_FREE)
/tests/gdimagecopyrotated/bug00320.c: 24 in rotate()
18         dst = gdImageRotateInterpolated(src, angle, black);
19         gdTestAssert(dst != NULL);
20         gdTestAssert(gdImageTrueColor(src) == 0);
21         gdTestAssert(dst != src);
22     
23      gdImageDestroy(src);
>>>     CID 529990:  Memory - illegal accesses  (USE_AFTER_FREE)
>>>     Calling "gdImageDestroy" dereferences freed pointer "dst".
24      gdImageDestroy(dst);
25     }
26     
27     int main()
28     {
29         rotate(GD_DEFAULT, 30.0);
```

cmb69:
https://github.com/libgd/libgd/blob/c2a5cbc74f407da0e38e9cbfdefb454e5a3300c2/tests/gdimagecopyrotated/bug00320.c#L18-L21

One of these assertion would likely fail, if the following `gdImageDestroy()` triggered a UAF. I've double checked with ASan locally (we already run that in a CI jobs). Looks like a false positive to me.

vapier:
there might be some attributes or annotations we can put on our function signatures so coverity understands them. it has no idea an assert function aborts unless we tell it explicitly.

otherwise, we can dismiss the report as a false positive via the web interface.

cmb69:
Actually, `gdTestAssert()` does not abort, so I've filed #944.

PR #932, opened 2025-01-06
Remove dead code

cmb69:
We've already returned for `GD_NEAREST_NEIGHBOUR` a couple of lines above.


vapier (approved):
we should delete one of the checks.  assuming the 2nd one is the one we want to delete instead of the 1st one ...

cmb69:
> assuming the 2nd one is the one we want to delete instead of the 1st one ...

Yeah, that's the problem. I don't really know. I just went with what we have now, and removed the second check.

PR #933, opened 2025-01-06
Fix CMake build regarding C++ wrapper

cmb69:
We should only build the C++ wrapper, if it is actually enabled.  This makes the `ENABLE_CPP_API` macro superfluous, so we can remove it.

While we're at it, we also drop the meaningless `__cplusplus` condition and the duplicate inclusion of config.h.

### DIFF
--- a/src/gd_io_stream.cxx
+++ b/src/gd_io_stream.cxx
@@ -1,7 +1,6 @@
 #ifdef HAVE_CONFIG_H
 #	include "config.h"
 #endif
-#ifdef ENABLE_CPP_API
 /* *****************************************************************************
 ** Initial file written and documented by:
 ** Kevin Shepherd <kshepherd@php.net> December 2007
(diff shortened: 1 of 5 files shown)


src/gd_io_stream.cxx
```diff
@@ -144,6 +138,3 @@ void ostreamIOCtx::FreeCtx(gdIOCtxPtr ctx)
 	{
 	delete (ostreamIOCtx * )ctx;
 	}
-
-#endif /* __cplusplus */
-#endif //ENABLE_CPP_API
```
vapier, line 149: does this mean the file is missing a final newline ?  if so, please add it.  i'm not familiar with this GH output.

src/gdpp.cxx
```diff
@@ -263,6 +260,3 @@ std::istream & operator>> (std::istream & in, GD::Image & img)
 	img.CreateFrom(in);
 	return in;
 	}
-
-#endif /* __cplusplus */
-#endif //ENABLE_CPP_API
```
vapier, line 268: here too

vapier (approved):

src/gd_io_stream.cxx
cmb69: No; the newline *was* missing previously; that's why it that sign has red background (missing final newline has been removed).

PR #934, opened 2025-01-06
Properly detect whether we need to link against libm

cmb69:
We should not make assumptions where we need libm, and where not. Instead we basically do the same as for autotools, namely checking whether we need to link against libm to have `sin()` available.

The actual detection is basically according to
<https://stackoverflow.com/questions/32816646#32820822>.


CMakeLists.txt
```diff
@@ -65,6 +65,25 @@ SET (CMAKE_ARCHIVE_OUTPUT_DIRECTORY
   "Single Directory for all static libraries."
   )
 
+include(CheckFunctionExists)
+
+# detect whether we need to link against libm
```
vapier, line 70: "Detect" and period at the end

cmake/modules/gd.cmake
```diff
@@ -7,10 +7,10 @@ MACRO(ADD_GD_TESTS)
 	FOREACH(test_name ${TESTS_FILES})
 		SET(test_prog_name "test_${TEST_PREFIX}_${test_name}")
 		add_executable(${test_prog_name} "${test_name}.c")
-		IF(WIN32)
-			target_link_libraries (${test_prog_name} gdTest ${ARGV0})
-		ELSE(WIN32)
+		IF(NEED_LINKING_AGAINST_LIBM)
```
vapier, line 10: we shouldn't need this at all.  libgd.so itself should link against libm.  pretty sure the m linkage here was to workaround us missing it in the main library.

examples/CMakeLists.txt
```diff
@@ -38,9 +38,9 @@ ENDFOREACH(test_name)
 # Add libraries *after* we've called add_executable on all the tests.
 
 if (PNG_FOUND AND JPEG_FOUND)
```
vapier, line 40: delete the block entirely

vapier (changes requested):

cmake/modules/gd.cmake
cmb69: Hmm, but what if the user builds a static libgd?

cmb69: Yeah, you're right. I'll work on that soon.

cmb69: Indeed, looks like this is already properly handled by `target_link_libraries()` for libgd and gdTest, respectively.

vapier (approved):

PR #936, opened 2025-01-07
Add missing .gitignores for tests

cmb69:


vapier:
we should add a CI test for this ...

cmb69:
> we should add a CI test for this ...

See #940.

PR #940, opened 2025-01-07
Let CI fail on untracked files

cmb69:


cmb69:
This requires more work, since we're placing at least some build artifacts in-tree, so `git status` will list these.

PR #941, opened 2025-01-08
Don't check for dependencies of dependencies

cmb69:
It's not our business to know which dependencies a package has; that's the job of `find_package()`.

---

It seems to me that our whole CMake `find_package()` dependency lookup is a bit wonky; for instance, we completely ignore [`PNG_DEFINITIONS`](https://github.com/Kitware/CMake/blob/b5ba2731192e6175159ed4a915acf0efaf44a219/Modules/FindPNG.cmake#L29-L31). And apparently, we ignore [IMPORTED TARGETS](https://cmake.org/cmake/help/latest/guide/using-dependencies/index.html#imported-targets), although:

> Where these are available, the project should prefer to use them instead of any CMake variables that may also be provided.


vapier:
cmake has a lot of problems independent of gd.  there's so many lessons of the past they failed to learn.  the fact that you find cmake files hardcoding paths or architectures is a testament to this failure.

Issue #900, opened 2024-12-18
TIFF handling is broken with libtiff after CVE-2023-6277 fixes


eds-collabora:
**Describe the bug**
TIFF files cannot be constructed from gd, and some tiff tests fail, e.g. `tiff/tiff_dpi`

**To Reproduce**
Build and run the tests against any version of libtiff with commit `5320c9d89c054fa805d037d84c57da874470b01a` applied.

**Expected behavior**
Tests should pass

**Actual results**
```
+ exec libgd-gd-2.3.3/tests/tiff/.libs/tiff_dpi
TIFFWriteDirectorySec: Warning, Creating TIFF with legacy Deflate codec identifier, COMPRESSION_ADOBE_DEFLATE is more widely supported.
TIFFWriteDirectorySec: Warning, Creating TIFF with legacy Deflate codec identifier, COMPRESSION_ADOBE_DEFLATE is more widely supported.
TIFFFetchDirectory: Warning, Requested memory size for TIFF directory of 156 is greather than filesize 0. Memory not allocated, TIFF directory notread.
TIFFReadDirectory: Failed to read directory at offset 1208.
GD Warning: Cannot open TIFF imagetiff/tiff_dpi.c:56: could not create dst
```

**Environment (please complete the following information):**
 - OS: ArchLinux
 - Version 2.3.3-7
- All libraries used by GD and their respective version (eg. Freetype 2.2.5)
libtiff-4.6.0-4

**Additional context**
This is my understanding of the behaviour of the `tiff_dpi` test: 
- In `gdImageCreateFromTiffCtx` we receive an `infile` of type `gdIOCtx *`
- This constructs a `tiff_handle` with  `new_tiff_handle`, leaving the size stored in the handle at zero.
- This is then handed as-is to `TIFFClientOpen` which refuses to read the image, because the reported file size is now 0, and the necessary file size for the contents it wants to map is nonzero.

cmb69:
I'm not sure what to do with TIFF support generally. Years ago [adding TIFF support to PHP](https://github.com/php/php-src/pull/2029#issuecomment-236952960) was rejected, because:

> Please no tiff yet. This is still experimental, no multi page support
> (defeats the whole tiff usage).

I don't think this has changed in the meantime.

cmb69:
[...]

Issue #814, opened 2022-01-29
imageftbbox regression between 2.2.5-5.2 and 2.3.0-2


mirabilos:
**Describe the bug**
imageftbbox returns a too small bounding box in libgd3 2.3.0-2 (Debian bullseye, testing/bookworm, sid) for all PHP versions

**To Reproduce**
See https://github.com/php/php-src/issues/8001

If I `LD_PRELOAD` libgd3 2.2.5-5.2 (Debian buster) `libgd.so.3.0.5`, the correct bounding box is returned.

**Expected behavior**
    [ascender] => 13
    [descender] => 1

**Actual results**
    [ascender] => 11
    [descender] => 0

**Environment (please complete the following information):**
 - OS: Debian bullseye
 - Versio: 2.2.5 (works), 2.3.0 (broken)
- All libraries used by GD and their respective version (eg. Freetype 2.2.5) Freetype 2.10.4+dfsg-1

**Additional context**
Downstream bugreport: https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=980759

pierrejoye:
Thanks for the report. :)

Do you use RAQM? 

Could you attach the font and script here too please?

mirabilos:
Unfortunately, http://www.mirbsd.org/wtf.cgi?q=RAQM

Sure.
[x.zip](https://github.com/libgd/libgd/files/7965225/x.zip)

```php
<?php
$text = "2020-12-23T13:25:44Z";
$font = "./Inconsolatazi4varl_qu-Regular.otf";	// attached
$fontsize = 14;
$bbox = imageftbbox($fontsize, 0, $font, $text);
$ascender = abs($bbox[7]) /* +1 */;
$descender = abs($bbox[1]);
$size_w = abs($bbox[0]) + abs($bbox[2]);
$size_h = $ascender + $descender;
$x = -$bbox[0];
$y = $ascender;
$im = imagecreatetruecolor($size_w, $size_h);
$bgcol = imagecolorallocate($im, 0xDD, 0xDD, 0xDD);
$fgcol = imagecolorallocate($im, 0x00, 0x00, 0x23);
imagefilledrectangle($im, 0, 0, $size_w - 1, $size_h - 1, $bgcol);
imagefttext($im, $fontsize, 0, $x, $y, $fgcol, $font, $text);
imagetruecolortopalette($im, FALSE, 256);
if (0) {
	imagepng($im, NULL, 9);
} else {
	print_r(array(
[...]

PR #946, opened 2025-01-18
Fix #867: Support libavif 1.0 in CMake build

### LINKED ISSUE #867: Support libavif 1.0 in CMake build
Because of the [`SameMajorVersion` libavif CMake config file policy](https://cmake.org/cmake/help/latest/module/CMakePackageConfigHelpers.html#generating-a-package-version-file) the latest 1.0.x will currently get rejected when [`find_package(libavif 0.8.2)`](https://github.com/libgd/libgd/blob/0d75136bd3e8651ded7c64a140791ed10de1c63c/CMakeLists.txt#L145) is requested.
```
CMake Warning at CMakeLists.txt:5 (find_package):
  Could not find a configuration file for package "libavif" that is
  compatible with requested version "0.8.2".

  The following configuration files were considered but not accepted:

    C:/msys64/ucrt64/lib/cmake/libavif/libavif-config.cmake, version: 1.0.1
```
No problem w/ Autotools build as it grabs and compares version from pkg-config using simple >=...

cmb69:
libavif applies the `SameMajorVersion` policy, so libavif 1.x.y is not considered to be compatible with 0.x.y.  To still be able to require at least libavif 0.8.2, we first look for any libavif 1.x.y, and only if that could not be found, we apply the old `find_package()` command.


cmb69:
This passes configure in the [Ubuntu CI, which was broken](https://github.com/libgd/libgd/actions/runs/12843920853/job/35816393287) due to updating to libavif 1.0.4, what happened because we're now running on Ubuntu 24.04 instead of Ubuntu 22.04. The failing tests are likely caused by updateing of other packages; I'll look into that as soon as possible.

@vapier, I think it makes some sense to have rolling CI (i.e. to use ubuntu-latest instead of ubuntu-22.04 or whatever), but it likely would make sense to have a scheduled nightly run, so such issues can be noticed even if there are no commits for a while. I was quite surprised by the sudden CI failure.

cmb69:
Note that I've checked this locally on Debian Bookworm, where I have libavif 0.11.1. It might make sense to *additionally* run on older Ubuntu (22.04 or even 20.04) to check that everything works with older libavif and other libraries.

Note also that neither our Windows nor MinGW builds are actually configured with `-DENABLE_AVIF`.

The test failures are likely due to #900, and and freetype/libraqm update (apparently not so minor [rendering differences](https://github.com/libgd/libgd/actions/runs/12844337449/job/35817225633?pr=946#step:10:301)) (that might be #814).

pierrejoye:
Cmake updated while fixing avif CI. Thanks!

PR #947, opened 2025-01-18
Also run CI on Ubuntu 22.04

cmb69:
This way we can test against some older library versions without needing to set this up ourselves.

---

Ubuntu 20.04 has EOL in May, and I guess it will be supported as GH hosted runner for a while longer. We could switch to Ubuntu 22.04 then.

### DIFF
--- a/.github/workflows/ci_ubuntu.yml
+++ b/.github/workflows/ci_ubuntu.yml
@@ -30,15 +30,15 @@ jobs:
               cxx: "clang"
             }
           - {
-              name: "Ubuntu 20.04 Intel GCC",
-              os: ubuntu-20.04,
+              name: "Ubuntu 22.04 Intel GCC",
+              os: ubuntu-22.04,
               cc: "gcc",
               cxx: "g++",
               cflags: "-msse2"
             }
           - {
-              name: "Ubuntu 20.04 Intel Clang",
-              os: ubuntu-20.04,
+              name: "Ubuntu 22.04 Intel Clang",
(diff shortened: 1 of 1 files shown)


cmb69:
Oh, libavif-dev is not available on Ubuntu 20.04. Might be better to use Ubuntu 22.04 right away.

.github/workflows/ci_ubuntu.yml
```diff
@@ -17,18 +17,31 @@ jobs:
...
             }
+          - {
+              name: "Ubuntu 22.04 Intel GCC",
+              os: ubuntu-22.04,
+              cc: "gcc",
+              cxx: "g++",
+              cflags: "-msse2"
+            }
+          - {
+              name: "Ubuntu 22.04 Intel Clang",
```
vapier, line 40: if the point is to get older lib coverage, i don't think we need to compile it twice

cmb69: Hmm, even if we assume that the libs are ABI compatible, we would need to transfer the build from one runner to another. So we would have to build on one runner, then upload to somewhere, only then start a dependent runner, download and run the test there. Looks like a serious complication, and might not even run faster.

But maybe I misunderstand what you have in mind.

vapier: I mean we only need 1 Ubuntu 22 builder, so let's delete this one

vapier (approved):

PR #948, opened 2025-01-18
Fix destruction of images in bug00320.c

cmb69:
We must only free the image, if it had been allocated.

Fixes CID 530896.


cmb69:
I'll have a look at CID 530895 soon.

vapier (approved):

PR #949, opened 2025-01-18
Attempt to fix CID 530895

cmb69:
If `gaussian_coeffs()` was called with a non-positve `radius`, bad things could happen, but we don't that; we add an `assert()` as safety net, and to enlighten static analyzers.

---

Frankly, I'm not sure what exactly is CoverityScan complaining about when they compain that `sum` might be zero, but this pre-condition `assert()` seems to be a good idea anyway.

We should probably also make sure that we don't call `gaussian_coeffs()` with a `radius > INT_MAX / 2 - 1` (or something like that; possibly using `overflow2()`).

### DIFF
--- a/src/gd_filter.c
+++ b/src/gd_filter.c
@@ -842,6 +842,8 @@ gaussian_coeffs(int radius, double sigmaArg) {
     double sum = 0;
     int x, n, count;
 
+    assert(radius >= 1);
+
     count = 2*radius + 1;
 
     result = gdMalloc(sizeof(double) * count);


src/gd_filter.c
```diff
@@ -842,6 +842,8 @@ gaussian_coeffs(int radius, double sigmaArg) {
     double sum = 0;
     int x, n, count;
 
+    assert(radius >= 1);
```
vapier, line 845: we should never use `assert()` in library code that users can get to.  would be nice if we could add a CI check to ban it in all library code.  tests & examples are fine to use assert/etc... of course.

we should return errors back up, and if we think it's helpful (e.g. for bad API calls), log an error via gd_error.

vapier (changes requested):

src/gd_filter.c
cmb69: This is in a static function which is only called by the `gdImageCopyGaussianBlurred()` API which alread has:

https://github.com/libgd/libgd/blob/0b8d70805d6f0b53a88902826ddc02208f8e32a1/src/gd_filter.c#L1009-L1011

This `assert()` is only a sanity check that we use the static function properly, and also a hint for static analyzers (and manual code review).

vapier: the nuance when it's "ok" to use assert and when people are just being lazy is very hard to communicate.  people see it being used in some places and think it's ok in general.

since it's a static function, the compiler should inline it.  and the compiler should be able to see the higher level has the same check and elide one of them.  a quick check over here suggests it does, so the generated code is the same.

vapier: enabling extra debug in a library should not cause it to abort and kill a process.  assert usage in programs is different from libraries.  we should have no asserts in the gd library regardless of its compilation mode.  the existing asserts are "ok" in the sense that they always force NDEBUG which means the calls are empty.  interesting that coverity accepts that behavior ... I can see both sides there.

PR #961, opened 2025-07-13
Remove unused HAVE_LIBTTF

petk:
The ttf library isn't used anymore and this macro isn't defined.


pierrejoye:
Thanks :) Good catch! left over from dropping freetype 1.x

PR #966, opened 2025-09-06
gdTransformAffineBoundingBox() fix UB overflow.

devnexen:
ref: https://github.com/php/php-src/issues/19730


pierrejoye:
thanks! Slightly different patch and test applied, see https://github.com/libgd/libgd/commit/f6202ec0de1c8340cf9b69c4a86c84fc63703207

PR #967, opened 2025-09-07
gdImageEllipse/gdImageFilledEllipse overflow fix.

devnexen:
ref: https://github.com/php/php-src/issues/19739


vapier:
we try to avoid ad-hoc overflow checks.  this is kind of what the overflow2() helper is for, although that operates on `int` rather than `int64_t`.  can you define a new helper for this ?

pierrejoye:
Thanks! 

I pushed a more portable test case and overflow function for these overflows (I think a few other needs it, like arcs f.e., will check that later)

See https://github.com/libgd/libgd/commit/0057de6fb1bd6d036215b0bb1c14817427e4220b

PR #969, opened 2025-09-10
gdImageRectangle() integer overflow fix.

devnexen:


vapier:
this doesn't feel like an integer overflow, and the checks don't seem to make sense.  the coordinates should not be negative (we might be missing a check for that), which means subtracting one positive int from another would never yield a value larger than INT_MAX or smaller than INT_MIN.  I guess you're also making sure we don't divide by 0 ?

please include a test case showing the problem you're trying to fix.

pierrejoye:
A note here, coordinates can be negative, gdImageSetPixel checks the boundaries. 

Unlike a similar case in the eclipse functions, for rectangle, we may check if the requested rectangle fits in the image or active clip area (there is a macro for that), no op if not at all, and separate drawing loop if it does (one to three of the edges, corners, may not be visible). Eclipse is more tricky to clip, not to check boundaries tho'

PR #972, opened 2025-10-17
Fix memory leak in filter failure

ndossche:
Detected with an experimental static analyzer.


nielsdos:
cc @devnexen How do we deal with porting fixes to php-src ? Looks like upstream is not very responsive.

devnexen:
I do not know any better than you do.

vapier:
thx bud

PR #974, opened 2025-10-21
Preserve transparent color information when converting a true color image to palette image

ArnaudBienner:
I found this when using `GD_QUANT_LIQ` (`GD_QUANT_DEFAULT` with `-DENABLE_LIQ=ON`).
As indicated in the correspond unit test, `GD_QUANT_NEUQUANT` has the same issue.

When LIQ library is not used (`-DENABLE_LIQ=OFF`) the default quantization algorithm (`GD_QUANT_JQUANT`) doesn't have the issue.

The problem was that the code to preserve the transparent color wasn't executed when using liq, so I moved it into a common `success` label used in both case.

### DIFF
--- a/src/gd_topal.c
+++ b/src/gd_topal.c
@@ -1552,6 +1552,7 @@ static int gdImageTrueColorToPaletteBody (gdImagePtr oim, int dither, int colors
 {
 	my_cquantize_ptr cquantize = NULL;
 	int i, conversionSucceeded=0;
+	boolean oimTransparent = FALSE;
 
 	/* Allocate the JPEG palette-storage */
 	size_t arraysize;
@@ -1581,6 +1582,7 @@ static int gdImageTrueColorToPaletteBody (gdImagePtr oim, int dither, int colors
 	 * must reserve a palette entry for it at the end of the palette. */
 	if (oim->transparent >= 0) {
 		maxColors--;
+		oimTransparent = TRUE;
 	}
 	if (colorsWanted > maxColors) {
 		colorsWanted = maxColors;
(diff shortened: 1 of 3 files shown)


pierrejoye:
Thanks for the report and patch :) I applied a slightly different or more complete patch:

https://github.com/libgd/libgd/commit/db2bd2ea7154ab4421ee62fee3e1d40afe124bee

The original patch moved the final transparent-entry handling into a shared success path, but that is not quite sufficient for LIQ and NeuQuant:

- For LIQ, the source `transparent` marker may be an opaque truecolor value, so libimagequant never sees it as transparent unless we explicitly convert matching pixels to alpha 0 in the LIQ input callback.
- For NeuQuant, the conversion returned a palette image but could leave `transparent` as the old truecolor pixel value, which is not a valid palette index.
- Simply appending a transparent palette entry after LIQ remapping would not guarantee that source transparent pixels are mapped to that entry, and could duplicate transparency if LIQ already produced one.

The committed fix instead:
- preserves the source transparent truecolor marker before conversion,
- makes LIQ treat matching pixels as fully transparent,
- ensures LIQ/NeuQuant results have a valid transparent palette entry,
- remaps pixels matching the original transparent marker to that entry,
- avoids using `gdImageCopy()` for NeuQuant in-place conversion because it skips pixels matching `src->transparent`.

I also added a regression test covering `GD_QUANT_JQUANT`, `GD_QUANT_NEUQUANT`, `GD_QUANT_DEFAULT`, and `GD_QUANT_LIQ` when libimagequant is available, for both in-place conversion and `gdImageCreatePaletteFromTrueColor()`.

PR #978, opened 2025-12-07
fix gdImageEllipse()/gdFilledImageEllipse() overflow (UBSAN)

devnexen:


devnexen:
ping @vapier :)

pierrejoye:
As the fix will work to fix the possible overflow, it breaks a key feature by return; in <= 0 (should be <0 too, top left is 0,0).

A better approach would require to use int128 for dx, dy, r, rx, ry. and eventually int64 for the other. But needs to verify if some platform may fail to build or related compiler limitations, we can then use the overflow test in these cases.

PR #980, opened 2026-01-23
add compile option ENABLE_FLAT_MEMORY to optimize memory allocation

SvOlli:
ENABLE_FLAT_MEMORY will call a single alloc* for a full image of pixel data, instead of one alloc per line of pixel data. Pointers to lines will still be set accordingly. All test cases succeeded.

autotools: --enable-flat-memory
cmake: -DENABLE_FLAT_MEMORY

Default is off to ensure maximum compatibility.

* malloc or calloc where applicable


vapier:
your commit message documents "what", but there's no explanation as to "why".
https://cbea.ms/git-commit/

that leaves me wondering why we should merge this.

the code also has incorrect style & indentation, but let's discuss the concept first.

SvOlli:
malloc and free are rather expensive operations. So when creating a FullHD image, why have 1082 malloc calls instead of just 3?
In my quickly hacked test, FullHD image creation and destruction is more than 30 times faster and though less resource burning.
Think of it as my humble contribution to save electrical energy in datacenters around the world trying to save a bit of the planet...
The speedup for my project at home I've already got. ;-)

pierrejoye:
We never really document pixels member as private, but I have seen projects using it, or doing their own little recipes in a more or less carefully.

I think it is a good optimization, and I think it could be the default as well. In a major version update I started some time ago, I drop that for a new image format, continuous buffer only.

In any case, I am not confident to add this to 2.x as it may causes invalid free for anyone doing gdFree(im->pixels[1]); f.e. (or arbitrary row), leading to corrupted heaps or similar.

SvOlli:
Pierre, thanks for the info.

Still, a "gdFree(im->pixels[1]);" will result in a double free on "gdImageDestroy(im);" which is also bad behavior. Also, this was the main reason, why I didn't make this the default with my change.

Nevertheless, I can totally understand the intention on not touching 2.x with change that touches architecture, when the work on the next major has already started. I can still use my fork until then.

However, is there a way to take a peek and make a few small contributions for the next major? I didn't find any branch here or in your personal fork. I found a few other optimizations in 2.x which I'd like to suggest there, before that becomes also stable.

I'm closing this, as it when it does not get merged there is no reason to keep it open.

pierrejoye:
@SvOlli see https://github.com/libgd/libgd/tree/features/surface

I will come back to it, load of refactoring, codecs to support etc. on the todos :)

PR #982, opened 2026-02-08
Prevent signed integer overflow in gdImageCopy* functions

### LINKED ISSUE #981: gdImageCopy functions may overflow signed int during bounds calculations
**Describe the bug**
Several GD copy routines can hit signed integer overflow when destination coordinates are close to `INT_MAX`.  
This happens during internal bounds or clipping calculations that add `dstX + w` or `dstY + h` using signed `int`.  

**To Reproduce**
```c
#include <gd.h>
#include <limits.h>

int main(void)
{
    gdImagePtr dst = gdImageCreateTrueColor(2, 2);
    gdImagePtr src = gdImageCreateTrueColor(2, 2);
    if (!dst || !src) return 2;

    int black = gdImageColorAllocate(dst, 0, 0, 0);
    int white = gdImageColorAllocate(src, 255, 255, 255);

    gdImageFilledRectangle(dst, 0, 0, 1, 1, black);
    gdImageFilledRectangle(src, 0, 0, 1, 1, white);

    /* dstX is chosen so that dstX + w overflows signed int */
    int dstX = INT_MAX - 5;
    int dstY = 0;

    gdImageCopyRotated(dst, src, dstX, dstY, 0, 0, 10, 10, 45.0);

    gdImageDestroy(dst);
    gdImageDestroy(src);
    return 0;
}
```

**Expected behavior**
No signed integer overflow should occur.

**Actual results**
With UndefinedBehaviorSanitizer enabled, a runtime error is reported for signed integer overflow in the copy or clipping calculations.

 - OS: MacOs
 - Version master
 - Compiler: clang or gcc
 - Build flags: -fsanitize=undefined (UBSan)

**Additional context**
Related report: [https://github.com/php/php-src/issues/21163](https://github.com/php/php-src/issues/21163?utm_source=chatgpt.com)

jordikroon:
Fixes #981


src/gd.c
```diff
@@ -2937,6 +2937,41 @@ BGD_DECLARE(gdImagePtr) gdImageClone (gdImagePtr src) {
...
+	int srcX, int srcY,
+	int w, int h
+) {
+	/* Check for null pointers */
+	if (!dst || !src) {
+		return 0;
+	}
+
+	/* Check for overflow in dstX + w */
+	if (w > 0 && dstX > INT_MAX - w) {
```
vapier: overflow helpers should be helpers in gd_security.c so we don't duplicate their logic in diff places

jordikroon: Thanks! I moved the logic to gd_security and made it more in line with overflow2.

jordikroon:
@vapier everything has been addressed. If you could take another look that would be nice.

pierrejoye:
Similar to https://github.com/libgd/libgd/pull/969

the bound check added helper is too restrictive and will break BC, specifically this check "if (w <= 0 || h <= 0) {"

both draw rectable, and gdImageCopy can share the same boundaries checks and clip to define the relevant areas or initial coordinates.  It will avoid  a duplicated logic as well.

PR #984, opened 2026-03-12
Decode the first frame of animated webp files instead of failing

aigilea:
This is a minimalistic PR to align libgd behavior when decoding animated webp files (currently fails with "gd-webp cannot allocate temporary buffer") with the one decoding animated gifs or animated pngs (decodes the first frame).

This PR makes use of libwebp demuxer api to find the initial frame first instead of passing the whole file to the decoder. This change doesn't affect decoding of non-animated webp files as demuxer api handles them correctly as files containing just one frame.

This PR solves numerous issues in downstream software products (nginx, nextcloud, xenforo, drupal, etc) relying on libgd to resize images.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -294,6 +294,12 @@ GD_LIB_PKG_CHECK([LIBWEBP], [WEBP], [libwebp >= 0.2.0], [
     gd_found_lib=yes
   ])
 ])
+GD_LIB_PKG_CHECK([LIBWEBPDEMUX], [WEBP], [libwebpdemux >= 0.2.0], [
+  AC_CHECK_LIB([webpdemux], [WebPDemux], [dnl
+    AS_VAR_APPEND([LIBWEBP_LIBS], [" -lwebpdemux"])
+    gd_found_lib=yes
+  ])
+])
 
 dnl Check for heif support.
 GD_LIB_PKG_CHECK([LIBHEIF], [HEIF], [libheif >= 1.7.0], [
--- a/src/gd_webp.c
+++ b/src/gd_webp.c
@@ -22,6 +22,10 @@
 #include "webp/decode.h"
 #include "webp/encode.h"
 
+#ifdef HAVE_LIBWEBPDEMUX
+#include "webp/demux.h"
+#endif
+
 #define GD_WEBP_ALLOC_STEP (4*1024)
 
 /*
@@ -110,6 +114,7 @@ BGD_DECLARE(gdImagePtr) gdImageCreateFromWebpCtx (gdIOCtx * infile)
 	gdImagePtr im;
 	int x, y;
 	uint8_t *p;
+	WebPBitstreamFeatures info;
 
 	do {
 		temp = gdRealloc(filedata, size+GD_WEBP_ALLOC_STEP);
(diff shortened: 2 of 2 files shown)


src/gd_webp.c
```diff
@@ -136,14 +140,32 @@ BGD_DECLARE(gdImagePtr) gdImageCreateFromWebpCtx (gdIOCtx * infile)
 		return NULL;
 	}
 
+	webpdata.bytes = filedata;
+	webpdata.size = size;
+	demux = WebPDemux(&webpdata);
+	if (!demux)
+	{
```
vapier: cuddle the braces and match existing file style

src/gd_webp.c
```diff
@@ -136,14 +140,32 @@ BGD_DECLARE(gdImagePtr) gdImageCreateFromWebpCtx (gdIOCtx * infile)
...
+	if (!WebPDemuxGetFrame(demux, 1, &iter)) {
+		gd_error("gd-webp cannot find the image");
+		WebPDemuxDelete(demux);
+		gdFree(temp);
+		return NULL;
+	}
+
 	im = gdImageCreateTrueColor(width, height);
 	if (!im) {
 		gdFree(temp);
```
vapier: doesn't this section also need to release the webpdemux stuff ?

configure.ac
```diff
@@ -294,6 +294,12 @@ GD_LIB_PKG_CHECK([LIBWEBP], [WEBP], [libwebp >= 0.2.0], [
     gd_found_lib=yes
   ])
 ])
+GD_LIB_PKG_CHECK([LIBWEBPDEMUX], [WEBP], [libwebpdemux >= 0.2.0], [
+  AC_CHECK_LIB([webpdemux], [WebPAnimDecoderNew], [dnl
```
vapier: this logic implies someone could have libwebp available but not libwebpdemux, but the code changes place webpdemux usage under HAVE_LIBWEBP instead of under HAVE_LIBWEBPDEMUX

vapier (changes requested):
doesn't this add overhead when processing webp files that aren't animated ?  is there a way to tell whether demux is needed and only use it when it is ?

aigilea:
Thanks for a quick feedback.

Everything's addressed.

aigilea:
The only thing I had to leave it the configure.ac macro call. If I change LIBWEBPDEMUX to LIBWEBP in the macro call, then while the macro reports checking for libwebpdemux it only adds -lwebp to the final linker list.
There was a typo in the first today's push so it compiled correctly, but after fixing this typo I've got a broken build so I've gone back to LIBWEBPDEMUX.
If there's a more correct way to add two pkg-config libraries under the same flag please tell.

src/gd_webp.c
```diff
@@ -130,18 +135,45 @@ BGD_DECLARE(gdImagePtr) gdImageCreateFromWebpCtx (gdIOCtx * infile)
...
 
 	im = gdImageCreateTrueColor(width, height);
 	if (!im) {
 		gdFree(temp);
 		return NULL;
 	}
-	argb = WebPDecodeARGB(filedata, size, &width, &height);
+
+	if (info.has_animation)
+	{
```
vapier: cuddle up to the previous line

src/gd_webp.c
```diff
@@ -110,6 +111,10 @@ BGD_DECLARE(gdImagePtr) gdImageCreateFromWebpCtx (gdIOCtx * infile)
 	gdImagePtr im;
 	int x, y;
 	uint8_t *p;
+	WebPBitstreamFeatures info;
+	WebPData webpdata;
+	WebPDemuxer *demux;
```
vapier: if these variables are only needed inside the `info.has_animation` block below, move them there

vapier (changes requested):
> The only thing I had to leave it the configure.ac macro call. If I change LIBWEBPDEMUX to LIBWEBP in the macro call, then while the macro reports checking for libwebpdemux it only adds -lwebp to the final linker list.

this is fine.  but the `gd_webp.c` file should check `HAVE_LIBWEBPDEMUX` and only include the dmux header & call its APIs when available.

aigilea:
All done

pierrejoye:
Hello :)

Just for the info here, I am adding libwebp animation support, similar to what we have in GIF (reading added earlier as PR)

pierrejoye:
See https://github.com/libgd/libgd/pull/1001

pierrejoye:
Closing, full webp animation PR is yet published. Will be in 2.4.0.

PR #987, opened 2026-03-23
Initialize platform variables in gdft.c

### LINKED ISSUE #986: compile error C4703: potentially uninitialized local pointer variable 'platform_specific'
**Describe the bug**
When I compile libgd 2.3.3 on MSVC 2022, with /sdl for added security I get an error:
`error C4703: potentially uninitialized local pointer variable 'platform_specific' used [C:\libgd-2.3.3\_build\src\libgd.vcxproj]`.

**To Reproduce**
1. `SET CFLAGS=/sdl`
2. Compile libgd 2.3.3 on MSVC 2022 (2019 probably also would show this, but I haven't tested.)

**Expected behavior**
libgd compiles without errors.

**Actual results**
`Generating Code...
C:\build\libgd-2.3.3\src\gdft.c(1499,1): error C4703: potentially uninitialized local pointer variable 'platform_specific' used [C:\libgd-2.3.3\_build\src\libgd.vcxproj]`

 - OS: Windows
 - Version [2.3.3]
 - Libraries
 -- JPEG: jpeg.lib (version "62")
 -- Freetype: freetype.lib (version "2.13.3")
 -- BZip2: libbz2.lib (version "1.0.8")

icanhasmath:
Initializing these values explicitly to NULL ensures that they cannot be exploited at runtime.

Closes: #986
compile error C4703: potentially uninitialized local pointer variable 'platform_specific'

### DIFF
--- a/src/gdft.c
+++ b/src/gdft.c
@@ -1137,7 +1137,7 @@ BGD_DECLARE(char *) gdImageStringFTEx (gdImagePtr im, int *brect, int fg, const 
 	gdCache_head_t *tc_cache;
 	/* Tuneable horizontal and vertical resolution in dots per inch */
 	int hdpi, vdpi, horiAdvance, vertAdvance, xshow_alloc = 0, xshow_pos = 0;
-	FT_Size platform_specific = NULL, platform_independent = NULL;
+	FT_Size platform_specific = NULL, platform_independent;
 
 	if (strex) {
 		if ((strex->flags & gdFTEX_LINESPACE) == gdFTEX_LINESPACE) {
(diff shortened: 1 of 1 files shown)


src/gdft.c
```diff
@@ -1137,7 +1137,7 @@ BGD_DECLARE(char *) gdImageStringFTEx (gdImagePtr im, int *brect, int fg, const
 	gdCache_head_t *tc_cache;
 	/* Tuneable horizontal and vertical resolution in dots per inch */
 	int hdpi, vdpi, horiAdvance, vertAdvance, xshow_alloc = 0, xshow_pos = 0;
-	FT_Size platform_specific, platform_independent;
+	FT_Size platform_specific = NULL, platform_independent = NULL;
```
vapier: I can see why compilers are bad at tracking `platform_specific`, but that doesn't apply to `platform_independent`.  so only initialize the former.

vapier (changes requested):

src/gdft.c
icanhasmath: Agreed - removed.

vapier:
thx bud

PR #992, opened 2026-05-04
Fix: -Wmaybe-uninitialized for image_type in gdImageCreateFromTiffCtx

pticon:
```
src/gd_tiff.c: In function 'gdImageCreateFromTiffCtx':

src/gd_tiff.c:955:43: error: 'image_type' may be used uninitialized [-Werror=maybe-uninitialized]
  955 |                 (image_type == GD_PALETTE || image_type == GD_INDEXED || image_type == GD_GRAY)) {
      |                                           ^

src/gd_tiff.c:819:25: note: 'image_type' was declared here
  819 |         int             image_type;
      |                         ^~~~~~~~~~

'''

### DIFF
--- a/src/gd_tiff.c
+++ b/src/gd_tiff.c
@@ -816,7 +816,7 @@ BGD_DECLARE(gdImagePtr) gdImageCreateFromTiffCtx(gdIOCtx *infile)
 	char	has_alpha, is_bw, is_gray;
 	char	force_rgba = FALSE;
 	char	save_transparent;
-	int		image_type;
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

**Comment by pierrejoye** [comment]
I think it is a false positive.

it does:
```c
	if (!force_rgba &&
	        (image_type == GD_PALETTE || image_type == GD_INDEXED || image_type == GD_GRAY)) {
```

force_rgba is the default, sets to true if no photometric matched. Each photometric's path sets image_type

PR #993, opened 2026-05-18
Enable builds on Fedora infrastructure using Packit

bkmgit:
May help in enabling builds on Aarch64 and S390x. Spec file is slightly modified from what is used at:
https://src.fedoraproject.org/rpms/gd/blob/rawhide/f/gd.spec 

To get it to work, someone with commit access will need to agree to make a Fedora account at:
https://accounts.fedoraproject.org/

and agree to Fedora Code of Conduct, then follow the steps at:

http://packit.dev/docs/guide#1-set-up-packit-integration

### DIFF
--- a/.packit.yaml
+++ b/.packit.yaml
@@ -1,0 +1,13 @@
+# https://packit.dev/docs/configuration/
+
+specfile_path: packaging/fedora/libgd.spec
+
+notifications:
+  pull_request:
+    successful_build: true
+
+jobs:
+- job: copr_build
+  trigger: pull_request
+  targets:
+  - fedora-rawhide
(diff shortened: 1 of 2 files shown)


vapier:
doesn't GH already provide Linux aarch64?

only value I can see with s390 is big endian since we don't have powerpc BE. although I don't know how many endian bugs we've actually had to justify the overhead.

bkmgit:
There is also ppc64le. README indicates Aarch64 is not tested, though it is tested in CI.  Pull request to update README https://github.com/libgd/libgd/pull/994

vapier:
I think we had ppc64le in the past via Travis or something.  it also wasn't terribly helpful in catching additional bugs.  and it's LE so doesn't handle the BE edge case.

I think the only bugs we have that are known across arches are tied to floating point, and those tend to be our testcases tuned to the rounding implementations in default x86 optimizations.  ideally those would get smoothed out with some tolerance so it'd work regardless of IEEE implementation/optimization behavior.

bkmgit:
Ok. Closing this then.

PR #1004, opened 2026-05-28
fix(png): remove libpng default user limits for large images

### LINKED ISSUE #931: Write png does not work when width or height exceeds 1000001
**Describe the bug**
As in the title

**To Reproduce**

```c
#include <stdio.h>
#include <gd.h>

int main() {
    const int width = 1, height = 1000001;
    gdImagePtr im = gdImageCreate(width, height);
    int color;

    for (int y = 0; y < height; y++) {
        color = gdImageColorAllocate(im, (255 * y) / height, (255 * y) / height, (255 * y) / height);
        gdImageSetPixel(im, 0, y, color);
    }

    FILE *pngFile = fopen("output.png", "wb");
    gdImagePng(im, pngFile);
    fclose(pngFile);
    gdImageDestroy(im);

    return 0;
}
```

then `gcc a.c -lgd && ./a.out`

**Expected behavior**
works

**Actual results**
```
libpng warning: Image height exceeds user limit in IHDR
GD Warning: gd-png: fatal libpng error: Invalid IHDR data
GD Warning: gd-png error: setjmp returns error condition
```

**Additional context**
Reported in SageMath.

https://github.com/sagemath/sage/issues/39289

Issue comes from libpng default guard, this can be removed with something like

```
  png_set_user_limits(png_ptr, 0x7fffffffL, 0x7fffffffL);
```

call to `libpng`.

re2zero:
## Summary
- Remove libpng's default 1,000,000 pixel user limits in both read and write paths by calling `png_set_user_limits()` with INT_MAX values
- Add test `bug00931.c` that creates a 100001×1 image (exceeding default limit), writes it to PNG, reads it back, and verifies dimensions and pixel data

## Changes
| File | Change |
|------|--------|
| `src/gd_png.c` | Added `png_set_user_limits(png_ptr, 0x7fffffffL, 0x7fffffffL)` in `gdImageCreateFromPngCtx` (read path) and `_gdImagePngCtxEx` (write path) |
| `tests/png/bug00931.c` | New test verifying large image PNG round-trip |
| `tests/png/CMakeLists.txt` | Register new test |

## Test Plan
- [x] New test `test_png_bug00931` passes — creates 100001×1 image, writes PNG, reads back, verifies
- [x] All 12 existing PNG tests pass (no regression)
- [x] Verified against libpng 1.6.53

Fixes #931


pierrejoye:
Thansks!
As mentioned in the report, agreed for that change. Users can manage their memory limit in a more effective/smart way instead.

PR #1006, opened 2026-06-05
fix(cmake): use ENABLE_* options to gate feature dependencies

### LINKED ISSUE #850: CMake: Inconsistent feature control
TLDNR: Replace `<Pkg>_FOUND` with `ENABLE_<Feature>`.

**Describe the bug**
In `CMakeLists.txt`, there are ON/OFF-options for feature control, e.g. `ENABLE_GD_FORMATS`:
https://github.com/libgd/libgd/blob/7b0d4198f2cba4cbc3ec68658e59474056622873/CMakeLists.txt#L12

These options are used to control lookup of *required* dependencies, e.g.
https://github.com/libgd/libgd/blob/7b0d4198f2cba4cbc3ec68658e59474056622873/CMakeLists.txt#L123-L125

This is fine.
However, in further processing (for macro definitions, selection of programs, export of gdlib.pc), the options are not used directly, but the `<Pkg>_FOUND` values are used instead:
https://github.com/libgd/libgd/blob/7b0d4198f2cba4cbc3ec68658e59474056622873/CMakeLists.txt#L192-L196

IMO this is wrong and should use the original `ENABLE_<Feature>` variables. A CMake package might have been found even if the user disabled the corresponding libgd feature. In particular, this happens in vcpkg for static triplets when CMake configs or Find module wrappers chainload additional packages to resolve transitive usage requirements.
- FreeType may use ZLIB.
- TIFF may use ZLIB, JPEG and WebP.

This problem is unlikely to be present with libgd's Find modules in `cmake/modules` which are unaware of actual static usage requirements. Modifying/replacing these modules is mandatory for static linkage. I still think it is an inconsistency which should be resolved.

As a variant, feature PNG also looks for ZLIB:
https://github.com/libgd/libgd/blob/7b0d4198f2cba4cbc3ec68658e59474056622873/CMakeLists.txt#L127-L130
I can imagine that this helps with static linkage. But I don't think it should override the effect of `ENABLE_GD_FORMATS`.

I could create a PR (based on packaging in vcpkg). I would also merge the second spots (`if(<Pkg>_FOUND)`) with the first spots (`if(<Feature>_ENABLED`). This should be possible without source changes which would need adjustments to the autotools build system. But there maybe a need for upfront discussion.

**To Reproduce**

Build for static triplet in vcpkg with different feature configurations and check build, or do the equivalent directly in CMake.
~~~
for I in core core,jpeg core,tools,jpeg ; do 
  for J in "" ,png ,tiff ,freetype ,fontconfig ,webp ; do
    ./vcpkg remove libgd; ./vcpkg install libgd[$I$J] --no-binarycaching
    <Check manually>
  done
done
~~~

Or maybe (untested):
~~~
cmake ... -DENABLE_GD_FORMATS=0 -DZLIB_FOUND=1
~~~

**Expected behavior**
If a feature is not enabed, it shall never be compiled into libraries, and related executables must not be built.

**Actual results**
A feature and related executables may be built even if explicit disabled by the user.

 - OS: any
 - Version: 2.3.3
 - Libraries from [vcpkg master today](https://github.com/microsoft/vcpkg/tree/253704407ae68efa37bf8f5b59b3e06dd40d3d3f).

**Additional context**
https://github.com/microsoft/vcpkg/pull/27551
which now uses a minimal patch to forward `ENABLE_<Feature>` to  `<Pkg>_FOUND` to achieve the desired build control.

add-uos:
## Summary

- Replace all `IF(<PKG>_FOUND)` checks with `IF(ENABLE_<FEATURE> AND <PKG>_FOUND)` in `CMakeLists.txt`
- This ensures features are only enabled when the user explicitly opts in via `-DENABLE_<FEATURE>=ON` AND the library is present
- Previously, a feature would be enabled if the library was found, even when the user had explicitly disabled it with `-DENABLE_<FEATURE>=OFF`
- Also fixes a typo: `PKG_REQUIRES_PRIVAES` → `PKG_REQUIRES_PRIVATES` for RAQM

## Changes

| File | Change |
|------|--------|
| `CMakeLists.txt` | Added `ENABLE_<FEATURE>` guard to all `*_FOUND` conditionals (lines 224–306) |

## Test Plan

- [x] `cmake -DENABLE_PNG=OFF -DENABLE_JPEG=OFF` — no `HAVE_LIBPNG`/`HAVE_LIBJPEG` in config.h
- [x] `cmake -DENABLE_PNG=ON -DENABLE_JPEG=ON` — `HAVE_LIBPNG`/`HAVE_LIBJPEG` correctly defined
- [x] `cmake` with default options (all OFF) — clean configure, no `HAVE_LIB*` defines
- [x] Multi-feature configure (`-DENABLE_PNG -DENABLE_JPEG -DENABLE_FREETYPE -DENABLE_TIFF -DENABLE_WEBP`) — all found and enabled correctly

Fixes #850


pierrejoye:
One idea to keep them and not checking if a dependency was found is to have an error, so user knows what was asked to be enabled was not found.

A more helpful, and catching missing deps for requested features,  change would be to have if ENABLE_X and NOT X_FOUND error instead, what do you think?

add-uos:
Thanks for the suggestion! I've updated the PR to use `FATAL_ERROR` when `ENABLE_X` is set but the dependency was not found, as you proposed. This gives users clear feedback about missing dependencies instead of silently skipping them.

Each feature block now follows this pattern:

```cmake
IF(ENABLE_X AND NOT X_FOUND)
    message(FATAL_ERROR "ENABLE_X is set but X was not found")
endif ()
IF(X_FOUND)
    # ... feature setup ...
ENDIF(X_FOUND)
```

The original `IF(X_FOUND)` conditions are preserved, and the `PKG_REQUIRES_PRIVATES` typo fix for RAQM is kept.

pierrejoye:
thanks :)

Merged!

PR #1008, opened 2026-06-08
Preserve Ultra HDR gain maps through GD transforms

gregbenz:
## Summary

- Preserve compressed Ultra HDR base/gain-map components through GD transform writes instead of regenerating from merged HDR output.
- Apply queued resize/crop/rotate/mirror operations to both the SDR base JPEG and gain-map JPEG, preserving gain-map metadata and ICC where appropriate.
- Avoid stale EXIF on transformed base JPEGs and force no chroma subsampling for re-encoded gain maps.
- Extend the UHDR example to accept percentage resize values.

## Tests

- `cmake --build .codex-build --target test_uhdr_uhdr_roundtrip test_uhdr_uhdr_metadata_resize test_uhdr_uhdr_ptr_roundtrip --parallel 4`
- `ctest -R uhdr --output-on-failure`
- `ctest --output-on-failure`
- `git diff --check`


pierrejoye:
thanks!! let merge it :)

gregbenz:
@pierrejoye are versioned releases no longer offered for GD? When I reference these updates for others, I can just note to use the latest main branch, but not sure how others might compare what they have after the fact.

pierrejoye:
It is master,  so it will be 2.4.0

I wait for some external details to do the next, hopefully last 2.3.x release and 2.4..0 test release

gregbenz:
Thanks, I'll reference 2.4.0 as initial support for gain maps. I appreciate all your help!

PR #1014, opened 2026-06-26
fix(test): skip fontconfig/basic when no fonts are installed

### LINKED ISSUE #367: 2.2.4 fails fontconfig/basic on linux
I get following error in a linux container with fontconfig-2.12.1

```+ cat tests/test-suite.log
====================================
   GD 2.2.4: tests/test-suite.log
====================================

# TOTAL: 151
# PASS:  150
# SKIP:  0
# XFAIL: 0
# FAIL:  1
# XPASS: 0
# ERROR: 0

.. contents:: :depth: 2

FAIL: fontconfig/basic
======================

fontconfig/basic.c:27: fontconfig: Couldn't find font.FAIL fontconfig/basic (exit status: 1)
```

xref: https://github.com/conda-forge/libgd-feedstock/pull/13

dengzhongyuan365-dev:
## Summary
- Move the skip logic from C code to CMake build system
- When fontconfig is compiled but no fonts are installed (e.g. minimal containers), the test is not registered at all via `execute_process(fc-list)` check
- Revert `basic.c` to its original strict form (no early return)

## Changes
| File | Change |
|------|--------|
| `tests/fontconfig/CMakeLists.txt` | Add `fc-list` detection; only register the `basic` test when fonts are available |
| `tests/fontconfig/basic.c` | Reverted to master (removed early return that masked all errors) |

## Why this approach
The previous implementation returned 0 for *any* `gdImageStringFT` error, which would hide actual code defects. Per reviewer feedback, the detection is now done at the CMake level — if no fonts are installed, the test is simply not registered, keeping the test strict for environments that do have fonts.

## Test Plan
- [ ] All tests pass with fontconfig + freetype enabled (fonts available)
- [ ] `fontconfig/basic` test does not appear in ctest list when `fc-list` returns empty
- [ ] Build system prints `"Fontconfig has no fonts available - skipping fontconfig/basic test"` status message in minimal containers

Fixes #367

### DIFF
--- a/tests/fontconfig/CMakeLists.txt
+++ b/tests/fontconfig/CMakeLists.txt
@@ -1,8 +1,20 @@
 if(FONTCONFIG_FOUND)
 IF(FREETYPE_FOUND)
-LIST(APPEND TESTS_FILES
-    basic
-)
+	# Skip fontconfig tests when no fonts are installed (e.g. minimal containers)
+	execute_process(
+		COMMAND fc-list
+		OUTPUT_VARIABLE FC_LIST_OUTPUT
+		ERROR_QUIET
+		OUTPUT_STRIP_TRAILING_WHITESPACE
+	)
+
+	if(NOT FC_LIST_OUTPUT STREQUAL "")
+		LIST(APPEND TESTS_FILES
+			basic
+		)
+	else()
+		message(STATUS "Fontconfig has no fonts available - skipping fontconfig/basic test")
+	endif()
 ENDIF(FREETYPE_FOUND)
 ENDIF(FONTCONFIG_FOUND)
 
--- a/tests/fontconfig/basic.c
+++ b/tests/fontconfig/basic.c
@@ -4,10 +4,6 @@
  * Without actually using fontconfig, passing an empty fontlist to
  * gdImageStringFT() would return an error ("Could not find/open font").
  * We're checking that it returns NULL.
- *
- * If fontconfig is compiled in but no fonts are installed (e.g., in a
- * minimal container), the test is skipped since this is an environment
- * issue rather than a code defect.
  */
 
 #include "gd.h"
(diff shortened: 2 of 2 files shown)


pierrejoye:
Hello :)

Thanks for your PR!

I see two different changes. One to add grey levels format export. The later is not related to the test fix.

Generally speaking, for the grey level export to png (or others), I prefer new options that will expose each codec full options, with some defaults f.e. instead of doing something like the automatic detection, which may not be what the user wants. 

Could you remove that change and only do the PR for the test fix? The grey level PNG can be a separate one or a discussion as an issue to find the best way to allow this :)

dengzhongyuan365-dev:
Hi @pierrejoye,

Thanks for the feedback! I've removed the greyscale PNG export changes from this PR. It now only contains the fontconfig test fix 。

Regarding CI: the `test_png_bug00086` SEGFAULT in the MINGW64 job is a pre-existing flaky test that parses malformed PNG data — it's unrelated to this change. Our branch is based on the latest master with only this one additional commit, and all other 8 CI jobs pass (Ubuntu GCC/Clang, macOS, Windows x64/x86/arm64, MINGW CLANG64).

pierrejoye:
Thanks for the update, now I see the actual changes.

There is one issue with this, any other errors will return as "success", making the tests rather useless.

I would prefer to do it in cmake, and skip the test if FC has no font installed.

f.e.
```cmake
execute_process(
    COMMAND fc-list
    OUTPUT_VARIABLE FC_LIST_OUTPUT
    ERROR_QUIET
    OUTPUT_STRIP_TRAILING_WHITESPACE
)

# Check if the output variable is empty
if(FC_LIST_OUTPUT STREQUAL "")
    message(WARNING "Fontconfig has no fonts available!")
    set(FONTCONFIG_HAS_FONTS FALSE)
else()
    message(STATUS "Fontconfig fonts detected.")
    set(FONTCONFIG_HAS_FONTS TRUE)
endif()
```

then in that test folder CMakeLists.txt, add the test only if FONTCONFIG_HAS_FONTS is true. That way the test remains relevant and will catch any other errors.

pierrejoye:
thank you!

PR #1018, opened 2026-06-29
add CMake options : BUILD_TEST, BUILD_EXEMPLES, BUILD_PROGRAMS , BUILD_DOCS, BUILD_CPACK

nono303:
Glad to see libgd is back is the business 😉
However, my long-standing CMake patch for building only the shared lib _(used with PHP on a [patched ext-gd](https://github.com/nono303/win-build-scripts/blob/master/patches/php-src/8.5/ext-gd.patch))_ keeps often breaking due to the numerous pushes lately.

So I'm taking this opportunity to open this PR if it seems relevant _(in any case, it doesn't introduce any regressions)_

@pierrejoye, I've set the default values to 0; feel free to decide what you think is best!

```cmake
OPTION(BUILD_TEST "Build tests" 0)
OPTION(BUILD_EXEMPLES "Build exemples" 0)
OPTION(BUILD_PROGRAMS "Build programs" 0)
OPTION(BUILD_DOCS "Build docs" 0)
OPTION(BUILD_CPACK "Build CPACK" 0)
```

++NoNo


pierrejoye:
hello!

Thanks for the PR!

Generally speaking it looks good.I did not check yet the CI but the default should match, so some runners still do the same work with the same setups. Have you looked at them?

Also, looking at your repo, you may be interested in https://github.com/pierrejoye/php-src-syncgd/tree/SYNC_GD_2_4

and the libgd's side here https://github.com/libgd/libgd/pull/1019

I am adding meaningful new API in libgd when some custom features exist in php (like the png and basefilter, a bit hackish, so I added a png export with options and use it in php instead of the hack).

The goal is outside the alloc and error functions/macros, the source must remain identical.

nono303:
Hi @pierrejoye 
✅ CI seems OK with all successful with default values.

vapier (changes requested):
"examples", not "exemples"

pierrejoye:
hello,

one last thing, pls make the options as it is now, on by default. Maybe the doc can be disabled by default, but the others are needed for many CIs, herre too :)

pierrejoye:
thank you :)

PR #1019, opened 2026-06-30
Features/add for php

pierrejoye:
This branch syncs:
- some fixes present in php's libgd and not here
- add some new APIs that could be useful here, they have been added to php's libgd or in the gd ext using private data access. adding cleaner APIs make it more maintainable
- ensure codes is almost identical on both sides
- ensure we don't add more customs on php's side

The php's side of this sync work is in https://github.com/pierrejoye/php-src-syncgd/tree/SYNC_GD_2_4


nono303:
Hi @pierrejoye 
I'll test this but concerning the [PHP side](https://github.com/pierrejoye/php-src-syncgd/tree/SYNC_GD_2_4), your fork is based on master with PHP 8.6 Internal API changes including `ZEND_CONTAINER_OF()` which will not compile on PHP < 8.6
I've patched it for my php 8.5 test like:
```patch
diff --git a/ext/gd/gd.c b/ext/gd/gd.c
index a4623d68..a8036220 100644
--- a/ext/gd/gd.c
+++ b/ext/gd/gd.c
@@ -144,7 +144,9 @@ static zend_function *php_gd_image_object_get_constructor(zend_object *object)
 	return NULL;
 }
 
-#define php_gd_exgdimage_from_zobj_p(obj) ZEND_CONTAINER_OF(obj, php_gd_image_object, std)
+static inline php_gd_image_object *php_gd_exgdimage_from_zobj_p(zend_object *obj) {
+	return (php_gd_image_object *)((char*)(obj) - XtOffsetOf(php_gd_image_object, std));
+}
 
 /**
  * Converts an extension GdImage instance contained within a zval into the gdImagePtr
```
Can you confirm this is correct?

I also had to add quickly
```patch
diff --git a/ext/gd/php_gd.h b/ext/gd/php_gd.h
index de83a8e0..0e70e0ad 100644
--- a/ext/gd/php_gd.h
+++ b/ext/gd/php_gd.h
@@ -121,4 +121,6 @@ PHP_GD_API struct gdImageStruct *php_gd_libgdimageptr_from_zval_p(zval* zp);
 
 #endif
 
+#define zend_ini_bool_literal(name) zend_ini_parse_bool(zend_ini_str((name), sizeof("" name) - 1, false))
+
 #endif /* PHP_GD_H */

```

pierrejoye:
hello :)

correct, also I didn't touch yet the extension code itself

my flow currently is:
1. fully sync libgd and ensure it builds, tests pass, some tests in php test buggy behaviors, updating them too (a handful)
2. drop gd 2.x external gd supports
3. add then the new apis in ext/gd

does that make sense?

nono303:
Your flow seems good to me!
My approach for `ext-gd` was to ignore php-src bundled `ext\gd\libgd` _(long time outdated)_ and link it with my own external shared libgd.
The main advantage of this is to decorrelate gd dependencies from the build of the PHP gd extension wrapper.
I've done it like this for long year without any issue but maybe I was wrong if `ext\gd\libgd` bundled src for PHP differs from original gd src libs... Only you will be able to tell me ;)
As a draft to open discussion :fast: https://github.com/pierrejoye/php-src-syncgd/pull/1

pierrejoye:
Sync done. We may add one thing here and there but everything is in sync now (in php's main, targetting php 8.6 only).

Issue #952, opened 2025-01-27
Relative font paths are not necessarily detected as such on Windows


cmb69:
Somewhere between GD 2.0.9 and GD 2.0.12, the handling of relative font paths has been changed. Previously, only absolute path got special treatment, but now this is also supposed to be done for relative font paths:

https://github.com/libgd/libgd/commit/871d0d852c2314e0f363ae4fd36fcdcbad00f541#diff-bf5179a04ba41e3f91087455a646396685b3eec7d0f2fcbfb463bae0b63b03a2L382-R384

The change obviously missed Windows file paths which may use backslashes instead of forward slashes. This is inconsistent, and very unintuitive for Windows users, since e.g. `./my_font.ttf` and `.\my_font.ttf` would be looked up differently (the former directly; the latter by using `GDFONTPATH`).

An obvious fix would be:
````diff
 src/gdft.c | 5 +----
 1 file changed, 1 insertion(+), 4 deletions(-)

diff --git a/src/gdft.c b/src/gdft.c
index 2e05be96..80b70356 100644
--- a/src/gdft.c
+++ b/src/gdft.c
@@ -1825,10 +1825,7 @@ static char * font_path(char **fontpath, char *name_list)
 			return "could not alloc full path of font";
 		}
 		/* if name is an absolute or relative pathname then test directly */
-		if (strchr (name, '/')
-		        || (name[0] != 0 && name[1] == ':'
-		            && (name[2] == '/' || name[2] == '\\')))
-		{
+		if (strpbrk(name, "/\\")) {
 			sprintf (fullname, "%s", name);
 			if (access (fullname, R_OK) == 0) {
 				font_found++;
````

**Comment by pierrejoye** [comment]
I will look at it, but if, I mean really "if", we want to support UNC path, that may require more than the proposed patch. I tend to think to let the caller calls the windows API to convert the path (which can check security access as well) and pass it to libgd. I need to investigate the implications of allowing UNC paths.

but the fix looks wrong.
```diff
-		if (strchr (name, '/')
-		        || (name[0] != 0 && name[1] == ':'
-		            && (name[2] == '/' || name[2] == '\\')))
-		{
+		if (strpbrk(name, "/\\")) {
```

name[1] == ':' tests for paths using drive's letter, c:\ or c:/
after that patch, it only looks for / or \.

Why I think it should take an absolute path, it is not the role of libgd to try to fix pathes, etc. calling layer can do it better, or already does, like php realpath f.e. Thoughts?

[...]

PR #1022, opened 2026-07-23
fix(gdft): detect backslash path separators in font_path()

### LINKED ISSUE #413: font_path() does not support UNC file names
I'm forwarding this from https://bugs.php.net/bug.php?id=61988.

[Obviously](https://github.com/libgd/libgd/blob/gd-2.2.5/src/gdft.c#L1723-L1738), font_path() does not support UNC file names, which always start with a double back-slash (`\\`).

dengzhongyuan365-dev:
## Summary
- `font_path()` checked for forward slashes and Windows drive letters to decide whether a font name is an absolute/relative pathname, but did not detect backslash (`\`) separators.
- This caused UNC paths (`\\server\share\font`) and relative Windows paths (e.g. `.\font.ttf`) to be searched in `GDFONTPATH` instead of being tested directly with `access()`, unlike their forward-slash equivalents.
- Add `strchr(name, '\\')` alongside the existing `strchr(name, '/')` check. If the direct `access()` fails, behavior still falls through to the `GDFONTPATH` search loop, so this is safe.

## Verification
- Builds cleanly with `-DENABLE_FREETYPE=1`.
- Test suite: 128/129 pass. The single failure (`test_gdimageaffine_interpolation`) is pre-existing and caused by PNG support being disabled in the build environment, unrelated to this change.

Fixes #413, #952

### DIFF
--- a/src/gdft.c
+++ b/src/gdft.c
@@ -1608,7 +1608,7 @@ static char *font_path(char **fontpath, char *name_list)
             return "could not alloc full path of font";
         }
         /* if name is an absolute or relative pathname then test directly */
-        if (strchr(name, '/') ||
+        if (strchr(name, '/') || strchr(name, '\\') ||
             (name[0] != 0 && name[1] == ':' && (name[2] == '/' || name[2] == '\\'))) {
             sprintf(fullname, "%s", name);
             if (access(fullname, R_OK) == 0) {


dengzhongyuan365-dev:
Closing this PR; will re-submit from the contributor's own account. Sorry for the noise.

PR #1033, opened 2026-09-03
Fix malformed ICC profiles in transformed UltraHDR JPEGs

gregbenz:
### What this fixes

Transforming an UltraHDR JPEG with an ICC profile can produce an invalid APP2 marker. The ICC signature includes an explicit null byte even though C already adds a terminating null byte to the string, making the signature one byte too long. This can also leave a second `ICC_PROFILE` wrapper inside the profile data.

Some tolerant JPEG and UltraHDR decoders still open the resulting file, but stricter consumers reject it. On macOS, affected images do not receive a Finder thumbnail and Preview reports that the file may be damaged or unsupported.

### Changes

- Use the string literal's existing terminator as the single null byte required by the JPEG ICC APP2 format.
- Extend the existing UltraHDR metadata test to validate the actual ICC marker structure and profile header in the compressed base JPEG.
- Continue verifying that the original ICC profile is preserved exactly.

### Compatibility

There are no public API or ABI changes. This only corrects malformed ICC framing when an UltraHDR image is transformed and re-encoded. UltraHDR writes without queued transformations remain byte-for-byte passthroughs.

### Testing

- Added a regression that fails before the fix and passes afterward.
- Passed all 241 configured libgd tests.
- Passed all 7 UltraHDR tests with AddressSanitizer and UndefinedBehaviorSanitizer enabled.
- Validated transformed output from real-world sRGB and Display P3 UltraHDR images with libultrahdr, ExifTool, Apple Quick Look, ImageMagick, and libjpeg.
- Tested resize, crop, rotation, mirroring, quality changes, chained transformations, SDR extraction, and single-channel gain maps across additional UltraHDR samples.


gregbenz:
CI note: the eight `Fuzz codecs` jobs fail during CMake configuration, before building or running any fuzzer. I checked a couple and they  say "Could NOT find JXL (missing: JXL_CMS_LIBRARY)". This appears to be a shared CI dependency issue rather than a failure caused by this change.

pierrejoye:
Thanks @gregbenz  :)

PR #1037, opened 2026-09-15
Fix memory leaks in tiffWriter() on error

ndossche:
`scan` and `tiff` are leaked on various error paths. In general, it's better to consolidate all the duplication to a single error block, which this commit does.

Note: perhaps the general error path should also call gd_error and not only the error path under TIFFWriteEncodedStrip ?

Note: these issues were found by my experimental static analysis tool I'm developing.


vapier:
looks reasonable, thx

